Issue #609, opened 2014-05-29
Admin 'Up' and 'Down' arrows/buttons not working


Moc:
Occurs on frontpage rules and extended user fields. 
- On frontpage rules, the up and down button are not functional (clicking them does not change anything) 
- The up/down image is missing on the extended user fields page

CaMer0n:
This needs to be upgraded to use the Admin-UI which provides sorting and inline-editing.

PR #639, opened 2014-06-13
Issue #609 - Admin 'Up' and 'Down' arrows/buttons not working

OTroccaz:
Functional alternative proposal as I think it is not possible to send
some parameter/value in a form with an image input type for
users_extended. I adapted it also in the same way for frontpage. This is
my first contribution and hope it will help ...


CaMer0n:
User Extended will need to be reworked using the admin-ui before we can commit changes.

PR #759, opened 2014-09-10
caption for carousel menu Fixed #672

### LINKED ISSUE #672: News carousel menu - caption (question)
I needed caption for news carousel menu. 
Default is always empty.
image

Did I do it right way? Or there is other way how to do it in theme folder? (I have correct tablestyle)
image

To be less confusing for others, maybe it should be coded this way:
image

Jimmi08:
Fixed #672

### DIFF
--- a/e107_plugins/news/news_carousel_menu.php
+++ b/e107_plugins/news/news_carousel_menu.php
@@ -54,7 +54,7 @@ e107::js('footer-inline',"
           {NEWSIMAGE}
            <div class="carousel-caption">
             <small>{NEWSDATE}</small>
-            <h1>{NEWSTITLE}</h1>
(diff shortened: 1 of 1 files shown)


e107_plugins/news/news_carousel_menu.php
```diff
@@ -127,6 +127,7 @@
 	
 	$footer = str_replace("{NAV}", implode("\n",$nav), $NEWS_MENU_TEMPLATE['carousel']['end']); 
 
-	e107::getRender()->tablerender('',$header.$text.$footer,'news-carousel'); //TODO Tablerender(). 
+	$caption = '';
```
SecretR: `$caption` should become a menu config value so that webmaster will be able to set (or leave blank)  caption for every menu instance.

Issue #197, opened 2013-05-02
social media Registration / signin


thunderlight411:
i'm testing the social media registration and log in functionality
and i must say that the registration for Facebook and Twitter are working very well.
but after a log off i can't log in with any of the 2 accounts into the website

When i check out the DB i see that the username for a Facebook user is named 
something like FacebookXXXX

is it possible to add the Facebook log in functionality into the logon area ?

Thanks for the reply

CaMer0n:
I have just upgraded hybridAuth and made some changes. Please test and let me know if it helps with the above. Thank you.

PR #831, opened 2014-12-30
Added the social media buttons into the login page and sign-in menu item See issue #197

saarloosw:
I can't test all social media platforms, but tested it on fb and it's working great.


CaMer0n:
I have added something where you could switch off the e107 user-registration and use social signup/login exclusively. This is exactly what I have done on e107.org for developers. In this case, we could add these icons back into the login menu, we just need to check the social-logins pref first.

Jimmi08:
Did somebody test new changes with facebook? I know that you use it with github on e107.org, but did you try it with facebook?  I tested it on localhost (yes, I have FB app with localhost url), but this can be reason why I wasn't able to test it.

CaMer0n:
I haven't tested with facebook in a long time. Testing does indeed need to be done on a live-site which FB and others can access during their authentication etc.

Issue #901, opened 2015-02-20
chatbox_menu: PHP Fatal error:  Call to a member function toHTML() on a non-object.


lonalore:
If anonymous posting is enabled, I get this error, and the website does not load completely.

lonalore:
I created a pull-request for this issue.

PR #902, opened 2015-02-20
Issue #901: chatbox_menu: PHP Fatal error: Call to a member function toH...

lonalore:
...TML() on a non-object.


CaMer0n:
Excellent! Thank you.

PR #903, opened 2015-02-20
Feature: Ability to hide the amount of chatbox posts from user profiles.

lonalore:


lonalore:
I reverted my last commit, because excluding is not necessary if plugin is not installed, bacause the e_user.php is only included by e107 if the plugin is installed.

PR #904, opened 2015-02-23
Fix classis style icons not lining up right

LaocheXe:
This: a.core-mainpanel-link-icon	{ height:80px; display:block; width: 120px; padding-top:20px;  }

To This: a.core-mainpanel-link-icon	{ height:80px; width: 120px; padding-top:20px;  }

Removed: display:block; to fix admin area icons in the classis style. They should line up with the font after removing it.


CaMer0n:
Please check it with all the different admin layouts (infopanel etc) and if it's okay, I'll commit your change.

LaocheXe:
Checked and it looks like all of them are displaying right.

PR #921, opened 2015-03-04
Update css for bbcode-panel

LaocheXe:
Added:
width: 95%; 
To:
div.bbcode-panel
So that the:
div.bbarea.large, .tbox.large, .helpbox.large {
Matches


CaMer0n:
@LaocheXe is this still necessary?  (I made recent css changes)

LaocheXe:
I have not tested it since enabling TinyMC

PR #978, opened 2015-04-21
Use language label instead of fixed "Sign In"

GHswitt:
Use language define LOGIN_MENU_L51 ("Sign In") instead of fixed "Sign In".


CaMer0n:
Thank you!!

PR #979, opened 2015-04-21
German language file for login_menu

GHswitt:
I've started translating some parts to use e107 with a german language interface (priority set for user language, not admin). This is the first part.


CaMer0n:
We don't store non-english files in this repo. But you may want to get in touch with other parts of the German community as they may already have a repo for language files.

PR #993, opened 2015-05-06
PHP Parse error:  syntax error, unexpected ';'

lonalore:
I made a mistake when I removed LAN string from the code. Sorry. :/

### DIFF
--- a/e107_admin/plugin.php
+++ b/e107_admin/plugin.php
@@ -3612,7 +3612,7 @@ exit;
 				}	
 				else 
 				{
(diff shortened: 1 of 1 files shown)


CaMer0n:
[file] is a known bbcode. please just use [x] and if multiple variables are required, you can use [y] and [z].

lonalore:
Ok. Yesterday I started to remove debug strings from LAN files, then I replace these kind of replacement patterns I made.

CaMer0n:
Got it and replied. Thanks

PR #997, opened 2015-05-08
Issue #6: Remove and optimize (hardcoded) LANS

lonalore:


CaMer0n:
Thank you! Btw. please take a look at the $tp->lanVars();  (for [x], [y], [z] substitutions).

Issue #1007, opened 2015-05-12
images not displaeyd


Batatafrita:
After the latest updates performed in 12/05 images placed on the site are not displayed.
They do not appear in the news or custom pages

CaMer0n:
Could you please provide a public URL to where this is occurring?

Batatafrita:
http://www.vaabrasil.com.br/sulamericano2015/page.php?12

tgtje:
if i may : see #803 last answer from Jimako

changing (line 148) using the idea as
![image](https://cloud.githubusercontent.com/assets/1811857/7619350/5658408c-f9b8-11e4-93b3-b1d6dc624456.png)

( so e_HTTP  >   becomes ../thumb.... ) 
result = working 
Using the latest files from april 29   #  master

CaMer0n:
Thank you, would you mind submitting a pull-request with that changes you already made?

Batatafrita:
I made the changes and went back to work
See for :
http://www.vaabrasil.com.br/sulamericano2015/news.php

Thanks and hugs

CaMer0n:
Thank you @tgtje . I found that the true cause of the issue was actually in the toDB() method. I have committed a fix for it, as you can see above. Thank you for the pull-request, I will leave it there as a reference in case a similar issue arises with the TinyMce parser.

tgtje:
applied latest files from git ( files till 8 hours ago > + admin ui 
upgrade )
The new fix seems to work just fine (i did test solemly  for this image 
problem).

PR #1009, opened 2015-05-14
fix parser php 

tgtje:
As descibed in topic #1007  here is the change i made (idea originated from Jimako (Jimmi08)


CaMer0n:
Closing as it's now "47 changed files".

PR #1018, opened 2015-05-21
Added Display options for Portfolio Gallery

LaocheXe:
Added the options to display the Portfolio Gallery in the Modern Business layout/template.


CaMer0n:
In the comments, you'll see it says: //TODO Add {GALLERY_PORTFOLIO}  to portfolio_menu.php 
ie. it should all be moved to a menu, then it can be removed, added, and made visible to userclasses etc. etc.

PR #1048, opened 2015-06-10
Language creation links php

tgtje:
First go , please check if in order updated admin/links php to use defines AND English/admin/lan_links php with added new defines


CaMer0n:
Your commit and changes are good! (y) Unfortunately, I reverted that change in the core just before seeing this. This mechanism has been moved to an external plugin.

PR #1053, opened 2015-06-13
Language LANs change (latest files used)

tgtje:
Txts replaced by LANs, where needed created defines in files active in region and most likely (if none existing)
Worked on : 
E......admin/ admin php  hovering txt's
Admin > dropdown admin option (personalize AND  links txt lan's )
Admin> stats DEMO panel (up to installed )

Note : installed stats screendisplays  the  3 digit DAY ;  is likely coded ? > can not find it....

hope i've got everything, tested a lot embed screenshot to display what has changed (Dutch = new defines working)

pers
demo
installed
logg

I am aware it is not probably the best way (workflow) but Dutch is quite well developed so any non working will show quickly

### DIFF
--- a/e107_admin/auth.php
+++ b/e107_admin/auth.php
@@ -68,7 +68,7 @@ if (ADMIN)
 		if(e107::getUser()->getSessionDataAs())
 		{ // TODO - lan
 			$asuser = e107::getSystemUser(e107::getUser()->getSessionDataAs(), false);
-			e107::getMessage()->addInfo('Successfully logged in as '.($asuser->getId()  ? $asuser->getName().' ('.$asuser->getValue('email').')' : 'unknown'). ' <a href="'.e_ADMIN_ABS.'users.php?mode=main&amp;action=logoutas">[logout]</a>');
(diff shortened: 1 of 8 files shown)


CaMer0n:
I believe there is something wrong with your editor.  As you can see from the commit, it has modified 880 lines of infopanel.php. Please correct and resubmit. Also, it is better to use a LAN description that matches the content of the LAN. eg. LAN_LOGIN_PWD_REQ

tgtje:
Ok will rework on that (change editor)
(please do not look at it in just posted commits, i believe the editor did it again)
i saw message after commiting, and found it displaying strange also (changed 4 items on 4 lines) in 2 files

still trying to get pull requests on a per item base..

CaMer0n:
Thank you! You might want to try using Notepad++ and experimenting with the line-ending encoding (ie. Linux, Windows, etc). (it has an option to convert them) Just resave the script and then check how many lines were changed.

tgtje:
I will, always used pspadeditor for years, but it seems troublesome when creating for Github lately, however other files are fine..... well we never will know. Going to Notepad ++ as standard. And if that does not  work (for me) i'll try sublime. Thanks for letting know.

PR #1057, opened 2015-06-16
top nav admin dropdown menu txt to  system LANs

tgtje:
On frontpage bootstrap and bootstrap 3 top nav admin (dropdown) txts replaced by LANs used in system


tgtje:
@Cameron : please see private message

PR #1060, opened 2015-06-16
Bootstrap and Bootstrap3 theme shortcode php

tgtje:
txt replaced by system own existing LANs
what/where ? > admin dropdown menu txts AFTER login at Top Nav Bar/ frontpage


CaMer0n:
That's more like it! :+1:

PR #1063, opened 2015-06-17
last remaining txt to LAN

tgtje:
All texts  replaced by LANs (files are now complete)

### DIFF
--- a/e107_plugins/login_menu/languages/English.php
+++ b/e107_plugins/login_menu/languages/English.php
@@ -75,5 +75,8 @@ define('LAN_LOGINMENU_48', 'Menu Config');
 define('LAN_LOGINMENU_49', 'Email: ');
 define('LAN_LOGINMENU_50', 'Username or Email: ');
 define("LAN_LOGINMENU_51", "Sign In");
+define("LAN_LOGINMENU_52", "Logged in as ");
+define("LAN_LOGINMENU_53", "Sign In with Google "); //future ? bootstrap
+define("LAN_LOGINMENU_54", "Sign In with Twitter "); //future ? bootstrap
 
 ?>
--- a/e107_themes/bootstrap/theme_shortcodes.php
+++ b/e107_themes/bootstrap/theme_shortcodes.php
@@ -28,14 +28,14 @@ class theme_shortcodes extends e_shortcode
 			if(deftrue('USER_REGISTRATION'))
 			{
 				$text .= '
-				<li><a href="'.e_SIGNUP.'">Sign Up</a></li>
+				<li><a href="'.e_SIGNUP.'">'.LAN_LOGINMENU_3.'</a></li>
 				';
 			}
 			
 			$text .= '
 			<li class="divider-vertical"></li>
 			<li class="dropdown">
-				<a class="dropdown-toggle" href="#" data-toggle="dropdown">Sign In <strong class="caret"></strong></a>
+				<a class="dropdown-toggle" href="#" data-toggle="dropdown">'.LAN_LOGINMENU_51.'<strong class="caret"></strong></a>
 				<div class="dropdown-menu" style="padding: 15px; padding-bottom: 0px;">
 				<form method="post" onsubmit="hashLoginPassword(this)" action="'.e_REQUEST_HTTP.'" accept-charset="UTF-8">
 				{LM_USERNAME_INPUT}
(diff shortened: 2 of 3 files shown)


CaMer0n:
Much improved! Thank you! Let's use "Sign In with [x]" and then do a lanVars() replacement for each in the theme_config. Please re-submit with that change, and then I'll add it. Thank you!!

tgtje:
Tried 2 days ( as far as my php knowledge goes) with the * Generic variable translator for LAN definitions.)
but am stuck, tried all possible changes on positions and or arrays etc...

could it be i am stuck with this : $tp = e107::getParser(); //TODO - find & replace $tp, $e107->tp

Please remember i am NO coder, i just try and test (trial and error until result > test)

(would be of course benefit if i could see a real working model with its 'paths' #where in which file and positions( they do seem to trigger call member etc..).
Thanks for reading

Moc:
Does this example help @tgtje?

```php
<?php
require_once("class2.php");
require_once(HEADERF);

// Just for this example
$tp = e107::getParser();
$text = '';
// ====================


// This is in the LAN file
define("LAN_LOGINMENU_53", "Sign in with [x]");

// This is in the theme shortcodes file or it could be pulled dynamically (@cameron?)
$social_media_providers = array("Google", "Twitter");

// Loop through the array of providers
foreach($social_media_providers as $provider)
{
	$text .= $tp->lanVars(LAN_LOGINMENU_53, $provider);
}


// Just for this example, render the output
e107::getRender()-> tablerender("Test LAN", $text);
require_once(FOOTERF);
```

Moc:
What about something like this? @tgtje @CaMer0n 

```php
<?php
require_once("class2.php");
require_once(HEADERF);

// Just for this example
$tp = e107::getParser();
$text = '';
// ====================


// This is in the LAN file
define("LAN_LOGINMENU_53", "Sign in with [x]");

// This is in the theme shortcodes file or it could be pulled dynamically (@cameron?)
$social_media_providers = array("Google", "Twitter");

// Loop through the array of providers
foreach($social_media_providers as $provider)
{
	$id = strtolower($provider);
	$text .= '<input class="btn btn-primary btn-block" type="button" id="sign-in-'.$id.'" value="'.$tp->lanVars(LAN_LOGINMENU_53, $provider).'">';
}


// Just for this example, render the output
e107::getRender()->tablerender("Test LAN", $text);
require_once(FOOTERF);
```

tgtje:
Want to post a lot of ideas, but let's ask a question first (easy one i guess).
Is/can the MAIN language file ( English .php) hold definitions that work in ALL related/running files >being a lan call from a plugin or used in mail or on template etc.....
So using 1 definition for all occurencies of 1 'txt term # name, title.(not as command) > something i would see/read on a page' throughout everything in system

Example  the term Subject (as a standalone word visible) occurs 5 times in definitions (all 5 are different LAN's, but use > Subject
Problem is: there are still remaing txt uses that are not converted for LAN def. use but could be (to complete all) ... (get the thought behind ?)

I can explain more for what i would do if someone wants to know... ( i often think out of the box and like K.I.S.S. # still got room on my pc. so i give myself some time before i start the experiment ;-))
( it would deliver 'build like above posted (cancelled) pull request if that would be ok)

( thinking about it...would be nice to have a brainstorm spot instead of posting it here or under issues at github # any question put there, triggers the count number UP > imo scares people away)

CaMer0n:
Looks good @Moc Yes, e107_languages/English/English.php is included everywhere that class2.php is included. So 99% of e107 (including plugins and themes) can use LANs from this file.

tgtje:
Thanks for the answer, now some explaining (proposal or alike, give it a thought please).
The question was always (for translating issues) to use as less LANs as possible. Recurring terms (words,names etc..) would benefit from 1 define to be used on multiple settings.

As explained above (and knowing there are still items to be switched to LANs) would it not be a usuable idea to bundle those ( and solve usage for 1 LAN def. for multiple occurencies) to make a hold (# testcase) to take a snapshot for all files (freeze moment), run a 'file and replace' and add that 1 definition into English/English php . Will not be an easy one on 1 go, but think do-able). I am aware it would upset many translators because it means a total different concept, and any plugin and theme builder HAS to either follow the preset 'standard' or create specific theme lans...( a simple concept which will take some time to 'overhaul'  everything and having English php as a kind of dbase file from which all drain). Feasible ?

CaMer0n:
Yes, it needs to be done. We are still in the beta phase, so it should be done NOW, not later. All repeated/recurring words/phrases that you find which could be added to English.php, please add them to the list in #6 so they can be reviewed.

PR #1065, opened 2015-06-21
Update forum_post.php

skcihmit:
under function updateReply()
$url = e107::url('forum','topic',$this->data,true);
to
$url = e107::url('forum','topic',$this->data);

With true at the end, I get the redirect url creating a recursive e107_plugins/forum within the url and causing a 404.


CaMer0n:
Thank you!! :+1:

PR #1067, opened 2015-06-22
Multi-language for News/Admin

e107bg:


CaMer0n:
I'm sorry, but I can't merge this because it erases e107.htaccess. Also, there are some terms which should be candidates for inclusion in lan_admin.php . eg. 'width', 'height', 'theme' etc.

PR #1073, opened 2015-06-23
Multi-language for e107

e107bg:


Moc:
Wouldn't it be better to do something like this, in https://github.com/e107bg/e107/commit/7c443062193164047cc9fe23c90545168793305c :

```php
define("ADLAN_MM_1", "The folder [x] is not writable. Please correct before proceeding.");
```

And then:
```php 
$tp->lanVars(ADLAN_MM_1, e_MEDIA)
```


Currently the LAN is not optimized. It means more work for translators and also tricker for developers as some languages may not use the same language structure.

CaMer0n:
Unfortunately, too many issues and conflicts introduced in this commit.

PR #1076, opened 2015-06-25
LAN_DELETE and LAN_ERROR

tgtje:
Where possible changed old defines ( at file level) to new master lan's
decommented old defines and on plugin/page/e_search deleted the blank space after closing php tag


CaMer0n:
Excellent work! :+1:

PR #1086, opened 2015-06-26
Update forum_icons_template.php

mcpeace-maw:
This is the support for font icons in the forum front end. I try it and works.


CaMer0n:
If we put this here, it will need to be removed from bootstrap3/theme.php also.

mcpeace-maw:
I think is better to have it here for all themes that use Font Awesome 4 and remove it from bootstrap 3 theme.

PR #1089, opened 2015-06-27
Update forum_icons_template.php

mcpeace-maw:
Bootstrap theme use font awesome 3.2.1 and I added support for that version of font awesome too. The forum icons are OK now in this theme too but panel classes used now in the forum are added in bootstrap latest versions and the theme use bootstrap 3.2.1 so it need some new css to looks OK.


Moc:
1085

CaMer0n:
I believe it would be better to add Fontawesome4 to the bootstrap theme.

mcpeace-maw:
Yes I think the same.

PR #1090, opened 2015-06-27
Update theme.php

mcpeace-maw:
OK, added font awesome 4.3. Now the font icons are OK in forum. The theme need some css adjusts too if use old bootstrap. Maybe custom forum template files with old bootstrap classes in the theme folder?

### DIFF
--- a/e107_themes/bootstrap/theme.php
+++ b/e107_themes/bootstrap/theme.php
@@ -3,7 +3,8 @@ if ( ! defined('e107_INIT')) { exit(); }
 /*
  * This is a 100% Pure Bootstrap Theme for e107 v2 
  */
-define('BOOTSTRAP',true); 
+define("BOOTSTRAP",true); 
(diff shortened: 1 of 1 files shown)


CaMer0n:
Yes, I believe the old bootstrap2 will need a custom forum template.

Moc:
@cameron: isn't e107 phasing out BS2? Why not use BS3 only? It's been stable for months and to maintain two different versions seems like a waste of resources.

CaMer0n:
I would love to drop BS2. In fact, if I could get the admin area working with BS3, I would remove BS2 from the core. It's a huge job though.

mcpeace-maw:
I just finished to update this theme (bootstrap core theme) to bootstrap 3.3.5 and theme upload not working on e107.org site. If someone want to test it, send me an email.

CaMer0n:
Can you please email me your theme so I can test the upload on e107.org?

PR #1091, opened 2015-06-27
Update forum.css

mcpeace-maw:


CaMer0n:
Because the page is so short, the min-height is prevent the footer of the site appearing halfway up the screen. You should be able to override this with your theme's style.css file.

mcpeace-maw:
I see all from outside, And I think the forum plugin must work out of the box in any theme with no changes. The rgbs colors make that, will work in light and dark themes the same. The imposed min-height I think is not right to be in the plugin, if is needed in a core theme must be added in that core theme,

CaMer0n:
Can you please post a URL to the min-height issue you see?

mcpeace-maw:
http://mcpeace.x10.mx/e1/e107_plugins/forum/forum.php

mcpeace-maw:
With min-height = 500px added in forum.css now I must add in all my themes min-height = 0 to remove the empty space.

PR #1093, opened 2015-06-28
Update featurebox_category_template.php

mcpeace-maw:
It is a bootstrap 3 carousel and need col-md-12 not span12 class.


mcpeace-maw:
I think this plugin need a css file inside his directory to work OK out of the box, also Camera Slider folder is better to be inside this plugin directory, if is not used in other core plugin.

CaMer0n:
It needs to work with bootstrap2 and 3. So it should contain the span12 and the col-md-12.

PR #1094, opened 2015-06-28
Update featurebox_template.php

mcpeace-maw:
Bootstrap 3 carousel don't need span class. I think also the carousel need fluid containers.
It is better with images without defined height, if the imposed image ration is not respected by the user, the image will not looks OK. The user have no clue about what ratio must have the images.

### DIFF
--- a/e107_plugins/featurebox/templates/featurebox_template.php
+++ b/e107_plugins/featurebox/templates/featurebox_template.php
@@ -11,10 +11,10 @@ global $sc_style;
 
 // e107 v2.x Defaults. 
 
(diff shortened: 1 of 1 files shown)


CaMer0n:
If the height is not fixed, then it may vary between slides, no?

tgtje:
Just a comment, you both know more about it then me, so i just post an url (found this interesting one; maybe you are aware of it) as it looks like tackling it both ways. http://parkhurstdesign.com/sites/bootstrapsandbox/slider-fixed.php

PR #1097, opened 2015-06-28
hover messages for admin php page (new lan definition added)

tgtje:
add new LAN DEF to English php

suggestion to use LAN_PROMPT (+ description) for future lan changes involving description fields on hover ??

adm


CaMer0n:
I would suggest the following, as it is not specific to the admin area. 
LAN_ENTER_USRNAME_EMAIL
LAN_PWD_REQUIRED

tgtje:
will do

CaMer0n:
Thanks! :+1:

PR #1108, opened 2015-07-01
Update forum_template.php

mcpeace-maw:
"Forum Statistics" link visible only for logged in users.


CaMer0n:
IF statements inside template files should be avoided. This is a candidate for a forum preference.

mcpeace-maw:
OK, no problems if we will have that in preference.

PR #1112, opened 2015-07-04
Update on LANs for DOWNLOAD, UPLOAD, CONTACT and Fix for UPDATE_ROUTINES page_theme

MikeyGMT:
Rationalise on LANs for DOWNLOAD, UPLOAD, CONTACT
and Fix for UPDATE_ROUTINES page_theme


CaMer0n:
Sorry, I can't let this one through. The use of MPREFIX is incorrect. There should be no field called e107_page_theme. There is obviously a problem, but this is not the correct fix for it. Also, English.php is starting to get too large. It is only for common terms used throughout the front AND back end.  Please try to keep pull-requests small (a few files), so that I don't need to refuse ALL the work due to a few minor issues. The rest is great ,@MikeyGMT it's just that I am unable to make a simple merge now due to those issues that have been introduced.

PR #1135, opened 2015-07-12
update

tgtje:
Added LAN's to mailoutphp
(tabs CAPTIONS) for admin>create
added lan defs to lan_admin php

tablanchange


CaMer0n:
Thanks. But these terms do not recur that much. At this stage it would be better to add them to lan_mailout.php instead.

tgtje:
Correct about recurring. It was decommented (development  stage) as a TODO lan for lan_admin..Hence..

Issue #1141, opened 2015-07-15
hardcoded Search in Book/Chapter page plugin e_search.php


Jimmi08:
Hi,
hardcoded string. 
'text' => "Search in Book/Chapter"

![image](https://cloud.githubusercontent.com/assets/5429548/8697699/76be7b3c-2af7-11e5-8f83-b8852c0cf122.png)

tgtje:
Nice find ;-) would be a candidate for a LAN, 
I have to test Git anyway so i make a commit.
Thanks..

tgtje:
As mentioned above : it is solved, can be tested and closed ?

Jimmi08:
With next site. I am not able to test it now. Sorry

PR #1142, opened 2015-07-15
Created LAN to replace hardcoded txt as described in issue #1141

tgtje:
files involved : plug/page/e_search php;  lang/English/lan_search php

### DIFF
--- a/e107_languages/English/lan_search.php
+++ b/e107_languages/English/lan_search.php
@@ -115,3 +115,4 @@ define("LAN_SEARCH_98", "News");
 define("LAN_SEARCH_201", "Please redefine your search query");
 
 define("LAN_SEARCH_202", "Toggle Advanced Mode");
+define("LAN_SEARCH_BOCHAPT", "Search in Book/Chapter");
--- a/e107_plugins/page/e_search.php
+++ b/e107_plugins/page/e_search.php
@@ -85,7 +85,7 @@ class page_search extends e_search // include plugin-folder in the name.
 		if(!empty($catList))
 		{
 			$search['advanced'] = array(
-								'cat'	=> array('type'	=> 'dropdown', 		'text' => "Search in Book/Chapter", 'list'=>$catList),
+								'cat'	=> array('type'	=> 'dropdown', 		'text' => "".LAN_SEARCH_BOCHAPT."", 'list'=>$catList),
 						//		'date'=> array('type'	=> 'date',			'text' => LAN_DATE_POSTED),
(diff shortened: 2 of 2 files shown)


Moc:
I'm pretty sure that only the English.php files are supposed to have the LAN_NAME structure. All other files still follow the LAN_XXX_## structure, such as LAN_SEARCH_001. 

@CaMer0n Correct?

CaMer0n:
@Moc, yes, that's correct. English.php and lan_admin.php are the only files which should contain the format LAN_(DESCRIPTION). Also, @tgtje, in this particular case, the LAN should be defined in page/English_global.php - not lan_search.php , as it is being called by:  e107_plugins/***page***/e_search.php.  You can use Plugin Builder's language checking utility in this case too.

tgtje:
Aha, learned something new. 
About page/English_global.php > does not exist
 went ahead and did some stuff > would this be the way to?? .. sorry but an image explains it better ( i hope : bit piled)
Still again the question remains ; how the LAN should carry how and what.. (Note : there is NO language folder and no lang files in PAGE plugin)

runall

CaMer0n:
@tgtje Excellent! Just a small modification to make.  forgot to mention that English_global.php is the other file which has it's own format. 

    LAN_PLUGIN_(PLUGIN_FOLDER)_(UNIQUE_DESCRIPTION)

tgtje:
Changed the global file to use LAN_PLUGIN_PAGE_BOCHAPT (thinking this would be the proper one now) and yes it appears as above imgage in the plugin builder check.

However : out of my league now :  Using this method (forget to check? its final routine yesterday),
the LAN is displayed on search page NOT the defined txt. This (imo) tells me there is something else to do to get it going...(calling the def txt from within lang folder) (tried at least 20 combos> still LAN).....( wonder if this decommented line holds the issue // Removes the need for search_parser.php, search_advanced.php and in most cases search language files.)

Reverting back to LAN_SEARCH_203  (instead of descr from comment Moc) it displays fine, but plugin builder tells it is missing... ( i can trick the builder by adding that line ALSO in global, but that is not to be the solution).

CaMer0n:
You'll want to check in the preference editor to see that 'page' is in 'lan_global_list'. 
Running a plugin scan might also help.

tgtje:
The plugin rescan did the trick and added auto an example file ? (Testing 1 step at a time) (not noticed before).
Thank you. 
Again something to remember to look...Thought v1 was already well thought about, the new stuff in v2 keeps me amazed (lets call them 'hidden' features ) Great job!...have to mention it..
Back to work : testchanges to commit..

CaMer0n:
thank you @tgtje Positive feedback is always appreciated! :+1:

PR #1144, opened 2015-07-16
Update infopanel.php

mcpeace-maw:
Responsive fixes = ul removed, it is not needed. Removed span 6 old class. Commented a div that add a bottom gap in that panel, it is not used, is empty.

### DIFF
--- a/e107_admin/includes/infopanel.php
+++ b/e107_admin/includes/infopanel.php
@@ -191,7 +191,7 @@ class adminstyle_infopanel
 				}
 			}
 	
-			$mainPanel .= "<div class='clear'>&nbsp;</div>";
+			// $mainPanel .= "<div class='clear'>&nbsp;</div>";
 			$mainPanel .= "</div>
 	      
 			</div>";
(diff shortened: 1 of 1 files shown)


CaMer0n:
Thanks @mcpeace-maw . How does it look now in bootstrap2?( I'm trying to make all changes backwards compatible, as to not break the existing bootstrap2 admin area)

mcpeace-maw:
I'll try bootstrap 2 admin to see. I'm working to make responsive the admin layout in bootstrap 3 theme. After I make all changes, the theme will work OK at all screen sizes on infopanel style that is more used.

mcpeace-maw:
I'll explain, inside div#right-panel is ul.thumbnails (that is in this file) and li.span6 (that is in admin_theme file). ul and li break the layout at sizes under 1000px., are not used in any way and need to be removed.

CaMer0n:
Thanks, and if I understand well, 'span6' will need to be added, to make it bootstrap2 compatible.

mcpeace-maw:
For center column and right column are already 2 divs with class .span6.col-md-6 that you find here in this file, so no need to add something here or in admin_theme after ul and li remove..

PR #1145, opened 2015-07-16
Update admin_theme.php

mcpeace-maw:
Responsive fix, commented li, no use.


CaMer0n:
Not sure I understand this one. Placing // inside an echo statement will usually echo '//' ?

mcpeace-maw:
The wrapper li need to be removed. The ul was removed first in infopanel.php

PR #1147, opened 2015-07-16
Update admin_style.css

mcpeace-maw:
Responsive fix for middle icon buttons.

### DIFF
--- a/e107_themes/bootstrap3/admin_style.css
+++ b/e107_themes/bootstrap3/admin_style.css
@@ -44,10 +44,38 @@ body						{ padding-top: 75px;  }
 .tab-border 				{ border-left: 1px solid #ddd; }
 .tab-content 				{ overflow: visible }
 
-.core-mainpanel-block 		{ text-align:center; width:140px; float:left; height:100px; display:block; padding:1px; margin: 0;  }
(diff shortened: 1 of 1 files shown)


mcpeace-maw:
On my test site bootstrap3 admin works well with these changes.

CaMer0n:
@mcpeace-maw Thanks! Will commit and test here. Cheers.

mcpeace-maw:
The only thing that is not responsive now is the top right icon navigation bar on admin layout of bootstrap3.

CaMer0n:
@mcpeace-maw I think these are the 2 remaining issues now (but probably a single issue affecting both). Missing right-margin: 

bootstrap3-missing-right-margin


infopanel-alignment-issue

mcpeace-maw:
What you see is the effect of the row div that I added in #right-panel div.
The warning wrapper div need span12 in bootstrap and col-md-12 in bootstrap 3. That will solve that issue.

CaMer0n:
@mcpeace-maw actually, the right-margin is gone from every page of the admin area.

CaMer0n:
@mcpeace-maw I see what's going on, will fix and commit.

mcpeace-maw:
All divs that are direct child of the added row (inside #right-panel div) need col-md-12 or other size to compensate the negative margin added by that row div.

PR #1148, opened 2015-07-16
Update admin_theme.php

mcpeace-maw:
For missing right margin issue.


CaMer0n:
@mcpeace-maw Please upgrade and test, I believe it is okay now?

mcpeace-maw:
I updated and tested and it is OK now, please close it.

CaMer0n:
Thanks :+1:

PR #1150, opened 2015-07-17
LAN added (e_search) and language folders including lan def file (Eng…

tgtje:
…lish_global) issuenumber #1141


CaMer0n:
Sorry! It's a detail, but the LAN has to match the folder name exactly.  It's 'page' not 'pages'. 

LAN_PLUGIN_***PAGE***_BOCHAP

tgtje:
Ok, will try again (will probably take some time off so it can take a while; mainboard blown )

Issue #1139, opened 2015-07-15
Feature: e107::url() modification to support query parameters


lonalore:
I did some modifications on the url() method of e107 class so now, it supports the query parameters. Using:

```
e107::url('plugin', 'list', $row, null, array('foo' => 'bar'));
```

If SEF URLs are enabled, it generates URL like this:

> something/path/key?foo=bar

... if disabled:

> something/path.php?id=key&foo=bar

I push my changes soon, and if you are satisfied, I create a pull-request.

lonalore:
Now, it supports hashtag too. Using:

```
$options = array(
	'query'    => array(
		'foo' => 'bar',
	),
	'fragment' => 'hashtag',
);

e107::url('plugin', 'list', $row, null, $options);
```

If SEF URLs are enabled, it generates URL like this:

> something/path/key?foo=bar#hashtag

... if disabled:

> something/path.php?id=key&foo=bar#hashtag

lonalore:
I modified index.php because the query parameters were not appended to the "new location" path, so the $_GET variable only contained the the query parameters are defined in the "redirect" path in e_url.php

so in this case: something/path/key?foo=bar
I got only: {e_PLUGIN}qna/questions.php?id=key
instead of: {e_PLUGIN}qna/questions.php?id=key&foo=bar

CaMer0n:
@lonalore Did you take a look at 'legacy'? (see forum/e_url.php) - it achieves the same thing to a large extent.

lonalore:
I tried everything, but in the $_GET variable I only got the query parameters are defined in the "redirect" path. I want to make a simple pager with SEF URL..

this is my SEF implementation in e_url.php:
```php
$config['questions_by_category'] = array(
	'regex'    => '^qna/questions/(.*)$',
	'sef'      => 'qna/questions/{qna_sef}',
	'redirect' => '{e_PLUGIN}qna/questions.php?c=$1'
);
```
The requested URL is: 
> qna/questions/core/?frm=1

and dumping the $_GET variable I only get this:
> array(1) { ["c"]=> string(11) "core/?frm=1" }

I think, I should get something like this:
> array(2) { ["c"]=> string(4) "core" ["frm"]=> int(1) }

What do I do wrong? o.O

CaMer0n:
Your regex is incorrect. Try this: 

        'regex'    => '^qna/questions/core/(.*)$',

lonalore:
[...]

PR #1152, opened 2015-07-22
Issue #1139: Feature: e107::url() modification to support query param…

lonalore:
…eters


lonalore:
Ok, I am working on it. :)

PR #1153, opened 2015-07-23
"tags" form element with selectize.js (normal edit and inline edit)

lonalore:


CaMer0n:
:+1: Nice work! Would you be able to style it to look the same the previous tags input we were using? ie. blue tags, less space between the 'x' and the word, and increasing the width of the input box?

lonalore:
Yes, I would. I'm going to wrtie some css for it tomorrow. Thank You!

CaMer0n:
Thank you!! :+1:

PR #1154, opened 2015-07-23
reworked version on e-search and added alng folders/file

tgtje:


CaMer0n:
Thanks. btw. the "Merge pull request #1 from e107inc/master … " is normally due to not having 'rebase' enabled in your git config file. Would be good if you can enable it, so we can avoid those entries after merge in the future.

tgtje:
Although i am trying to do my best with Git, i still have some issues with it. Probably related to my own usage. I had to do a speed course and still not comfortable with. It's always related to a certain step before or after...Thanks.

CaMer0n:
I use TortoiseGit, I find it to be the easiest to use.

PR #1170, opened 2015-08-22
Placeholder (required) field on login menu (menu) AND signup field

tgtje:
Required placeholder field for Password (translation) hardcoded txt replaced by LAN, hover field admin icon (re-order), and latest commit to be able to use forum plugin translated defines (not fully implemented at previous state?)

### DIFF
--- a/e107_handlers/form_handler.php
+++ b/e107_handlers/form_handler.php
@@ -3152,7 +3152,7 @@ class e_form
 					{
 						$mode = preg_replace('/[^\w]/', '', vartrue($_GET['mode'], ''));
 						$from = intval(vartrue($_GET['from'],0));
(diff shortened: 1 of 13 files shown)


CaMer0n:
Thank you, fully tested already? (specifically the LANs)

tgtje:
Yes i did, want to take a look at front ? http://tgtje.eu/teste107/ Dutch is running, However there are stiil 'little' things that need to be translated ( means : LAN vs hardcoded ; mostly admin related) Not all seems at this point be fully there ( expected at stage). On forum front/parent/child..., again "by" as in news.. and the word "ago".
Note: while adding the forum LAN calls (e107 :: i was/still am unsure if 'true' is absolutely at place.
While Dutch is nearly completed, it becomes more and more easier to have a visual aid too.
Focus is thesedays more inside; admin piece.
If you want to have admin access please mention.

tgtje:
Did run testing on upgrade (x 3) (afterwards AND as kind of installpackage for upgrade. Have to inform, see no strange behaviour . outcome = alike fresh install.
Note : those using custom templates when upgrading (forum icons instead of txt./bs buttons) will suffer (visual) because that can not be dealt with (custom code). The links beneath however + alt function.

PR #1194, opened 2015-09-22
translation for messages when installing/deinstalling plugins admin s…

tgtje:
…creen (most common used)

(i am aware of using double quotes (double); but too much change has effect on following lines so walked the path between; so it becomes functional, and throws no errors (lack in coding skills, or missing things, since all LAN's need different approaches  depending position in code line..)

### DIFF
--- a/e107_admin/plugin.php
+++ b/e107_admin/plugin.php
@@ -1573,7 +1573,7 @@ class pluginManager{
 				}
 
 				$opts['delete_ipool'] = array(
-					'label'			=>'Remove icons from Media-Manager',
+					'label'			=> LAN_REM_ICONS,
 					'preview'		=> $iconText,
 					'helpText'		=> EPL_ADLAN_79,
 					'itemList'		=> array(1=>LAN_YES,0=>LAN_NO),
--- a/e107_handlers/plugin_class.php
+++ b/e107_handlers/plugin_class.php
@@ -531,7 +531,7 @@ class e107plugin
 			if (vartrue($this->unInstallOpts['delete_ipool'], FALSE))
 			{
 				$status = ($med->removePath(e_PLUGIN.$plugin, 'icon')) ? E_MESSAGE_SUCCESS : E_MESSAGE_ERROR;
-				$mes->add('Removing Icons from Media-Manager', $status);
+				$mes->add( LAN_REM_ICONS, $status);
 			}
 			return;
 		}
@@ -661,7 +661,7 @@ class e107plugin
 		$type = $this->ue_field_type($field_attrib);
 		$type_name = $this->ue_field_type_name($type);
 		
-		$mes->addDebug("Extended Field: ".$action.": ".$field_name." : ".$type_name);
+		$mes->addDebug("".LAN_EXTF." ".$action.": ".$field_name." : ".$type_name);
 		
 		// predefined
 		if($type == EUF_PREFIELD)
@@ -1609,7 +1609,7 @@ class e107plugin
 		{
 			if ($function == 'install')
 			{
-				$text = "Installation Complete.";
+				$text = LAN_INST_COMP;
 
 				if ($this->plugConfigFile)
 				{
@@ -1693,7 +1693,7 @@ class e107plugin
 						$query .= $tableData['data'][$k];
 						$query .= "\n) ENGINE=". vartrue($tableData['engine'][$k],"InnoDB")." DEFAULT CHARSET=utf8 ";
 
-						$txt = "Adding Table: <b>{$v}</b> ";
+						$txt = "".LAN_ADD_TABLE." <b>{$v}</b> ";
 						$status = $sql->db_Query($query) ? E_MESSAGE_SUCCESS : E_MESSAGE_ERROR;
 						break;
 
@@ -1701,7 +1701,7 @@ class e107plugin
 						if (!empty($options['delete_tables']))
 						{
 							$query = "DROP TABLE  `".MPREFIX.$v."`; ";
-							$txt = "Removing Table: {$v} <br />";
+							$txt = "".LAN_REM_TABLE." {$v} <br />";
 							$status = $sql->db_Query_all($query) ? E_MESSAGE_SUCCESS : E_MESSAGE_ERROR;
 
 						}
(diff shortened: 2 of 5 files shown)


CaMer0n:
Thank you! Would you mind fixing this code issue before I merge it?
Example: (but there are many like this)

    "".LAN_ADD_TABLE

  should be

    LAN_ADD_TABLE

tgtje:
As mentioned (lack of skill), i will try to eliminate the use of the double double quotes..
I will make a new pull request and close this one.

tgtje:
Did try to eliminate those double quotes..
Partially (skill level) i can not succeed : 
example : 
$text = LAN_INST_COMP ;  (works)

$txt = "Adding Table: <b>{$v}</b>";  (original)
changing to $txt = LAN_ADD_TABLE <b>{$v}</b>; (in different versions with and without "or ' etc.. not posible) it looks as if html (strong in place also does not work) use cripples it. (it also might be the value call here, deleting the html (b) cripples again the rest.. and error happens on value { 

As long as i keep getting errors here (point) i can not check for others..

Moc:
@tgtje Try this:

```php
$txt = LAN_ADD_TABLE.": <b>{$v}</b>";
```

tgtje:
I will try, thank you.. Will keep an eye out for the use of  :

tgtje:
Got it  working, but before i make a pull request > discussion ???> 

Text line  (system Information  main blue bar in admin fields...) 

Settings not saved as no changes were made.......

WHAT is the actual meaning (translated) for SETTINGS in this case.
It has nothing to do with plugins etc.. as changes ARE done (install/uninstall).
Is it to be understood that there are NO SYSTEM CORE settings influenced ???
In translation or interpretation it does make (sorry) no sense....( or triggers a thought of a mistake in the system)

Moc:
It's a message saying that the user made no changes, so nothing has been updated in the database. This happens for example when editing a news item and hitting the update button without actually changing anything. 

I think there's a bug somewhere that causes this message to pop up a little too often though. 

In Dutch I'd say something like: "Er zijn geen wijzigingen aangebracht."

tgtje:
So as a general it best could display (has to be a new define too) as : 
   Settings on (system) core level not saved as no changes were made.?

In Dutch : Geen wijzigingen op (systeem) core level aangebracht....

Although (system) core level also is an open 'term/item' , it 
differentiates from basic, and poses that ( an above posted message)  
did not affect the system further (more understandable imo.) Use of word 
(system) is a debate but, it does post another level of.. that way.


Op 21-10-2015 om 13:00 schreef Tijn Kuyper:

tgtje:
Just another question regarding statements (advice) 
Do NOT translate debug information. (eg. anything inside addDebug();
Working on definitions inside plugin_class (that is what previous pull had inside), stumbled on this statement. A lot of message related stuff in inside these.. being addDebug, addError..
Although it all seems to show it as it should when applied (LAN by def.), the question remains.. do i do it or yes/no... (about 75% of all messages seems to be evoked from within here..plugin class)

Moc:
Debug no. Info, warnings and errors, yes.

Moc:
To clarify: addDebug: do not translate but do translate addInfo, addWarning and addError.

Moc:
> So as a general it best could display (has to be a new define too) as :
   Settings on (system) core level not saved as no changes were made.?

No, not really, since 'no changes' really mean no changes anywhere. For example: when you edit a news item, but you do not change any of the values (like the news title, body and extended fields, etc.) and you hit the update button, this message will show up saying that nothing has been changed so nothing has been saved to the database. Basically "No changes were made". In dutch: "Er zijn geen wijzigingen aangebracht". 

Possibly @CaMer0n  can clarify this further though.

tgtje:
Thanks, more clarification is needed (as viewpoint as noobie, entry user imo) Language is  psychological  of impact. The better the understanding, the less misinterpretation, the better positive experience etc..... ( i always feel when trying to share or contribute, i also am kind of a salesman ;-) )

Moc:
I'm not sure how to clarify it furter though. The message simply means "No changes were made so nothing has been updated in the database.". Nothing meaning, no data (in general). This could be preferneces, news items, welcome messages, custom pages, anything, depending on which page you are. For example, when you go to the preferences section and hit the update button without actually changing any preferences, you get the same message. 

Could you try to explain where the confusion is?

tgtje:
Trying : think very basic (entry level)

reading (English common message) Settings not saved as no changes were 
made..

simple enough > understandable : yes (for me, since i know a little 
about the system)...,.

But getting this message whilst uninstalling/installing plugins for 
example.. (which in itself IS a change on whatever 'level' )
the message than is non 'usuable' because (thought)....... huhhh.. i 
just did something, but reports > nothing...
Did it actually work ?? or not...

So the 'concept' or thought 'below' the message is too 'narrow of base' 
and (whatever language translation follows after), the english present one
is íncorrect ?  as 'general' message (not the idea behind it).

So in above meaning even a new def like : Quo : No changes were made so 
nothing has been updated. (or something similar)

does have a too narrow 'base' as again 'changes' occurred ( not on level 
of system (core) ).

Since i do not have a total overview of where and when it happens 
(message) and again thinking how to 'create'a  more 'general' message.

Would a sentence like:  example : (english) No changes to system 
settings made, no effect by above action. Or : Above action has no 
effect on changing and saving other settings to system.

Imo this has a more general idea behind it ('the feel') but is more 
complete in its meaning and still acurate..

(Dutch short : ) Interpretatie wanneer dit bericht wordt getoond > als 
er niets bewaard wordt (instellingen.. huhh..welke ???, ik verander toch 
iets)
Hoe dit op te vangen middels een nieuwe zin (interpr.) die algemener is, 
en toch eigenlijk alles 'aanraakt'; en toch onderscheid 'voorstelt' dat 
een en ander op een ander niveau pas invloed heeft.. (zonder dit 
specifiek te gaan benoemen), en daarom is de engelse term dus soms 
wel/soms niet 'correct' in zijn 'uitleg'op DAT moment.. Streven is > 1 
melding > 1 vertaling ( niet creeren van specifieke extra defines...).

Moc:
I get your point now. 

> But getting this message whilst uninstalling/installing plugins for
example.

This is exactly what I meant with "I think there's a bug somewhere that causes this message to pop up a little too often though." in one of my previous reply. 
I think that this message should not pop up when installing/uninstalling plugins for example. This needs to be verified by @CaMer0n though. 

If this message does not show up in this particular case (install/uninstalling plugins), does the message make more sense? 

I think we could simplify it a lot... for example:

- No changes were made

(Dutch: Er zijn geen wijzigingen aangebracht). 

/cc: @CaMer0n What do you think?

tgtje:
Ok, thanks for answer, but still the 'relationship'  to/from the message remains than just the 'wording'changed
Ii will set these messages/LANs and defines on hold for present and focus on other parts for now. 
This partical file has  a lot of messages (plugins class) and sad to mention it : will increase numbers of new LANs.
(if those will appear on action taken, displayed and need explanation for the user/admin)..

CaMer0n:
It's good, but I believe the LAN names need to be improved. Example:
LAN_ADD ,  = "Add", "Added" and "Adding", have different meanings. The same for "REMOVE" and "REMOVED". 

A more generic, ("LAN_ADDING_X", "Adding [x] ") and ("LAN_REMOVING_X", "Removing [x]"). 
Followed by: 
LAN_TABLE  = Table
LAN_LINK = Link
etc. etc. 
.. maybe be more efficient and involve less duplication.

Issue #1183, opened 2015-09-14
https issue: browsers refuse to load unsafe http data (like bootstrap)


aducom:
I have a site setup with E107 and https. Because some data like bootstrap is retrieved using http, firefox and Ie are complaining that there is secured data and unsecure data and refuse to load the unsecure data causing a corrupted website. I can click to verify and after that it's ok, but this looks like an issue to me. I can set the http to bootstrap to https so there's no reason why e107 is using http for these. After all, it's no problem to load secure data from within a http site, but the other way around is a problem under the most recent browsers. But perhaps I'm doing something wrong?

CaMer0n:
Thanks. Is this also in the admin area, or just the front-end?

aducom:
Good question. I checked and the admin area seems to be working fine. So I guess it's only the front-end.

Moc:
@aducom Please update from Github and see if there are any remaining issues.

aducom:
This was fixed in one of the updates of Cameron. So this issue can be closed.

Moc:
@aducom Perfect! Thanks for checking.

Issue #805, opened 2014-11-05
problem using ssl + normal protocol


willem010:
My server has both http and https, and for now i want to use both.
therefore i cant check the "use ssl" box as this makes all connections use ssl.

one of the problems i'm having is the (jq) scripts are always loaded over http
resulting in an error when my user is using https.

can you guys add a detection, and output the scripts using the correct protocol for the request?

(else i would have to copy all scripts to local server, disabling the jquery options all togeher. and load them manually .. )

Moc:
There have been many changes with regards to http/https. If there are any outstanding problems, please open up a new issue.

willem010:
works perfectly. thnx!

PR #1197, opened 2015-09-30
To avoid cross domain conflict when using https

OTroccaz:


lonalore:
It would be better to use protocol-relative urls like:

```php
e107::js('url',"//ajax.googleapis.com/ajax/libs/jqueryui/1.10.1/jquery-ui.min.js");
```

output:
```html
<script src="//ajax.googleapis.com/ajax/libs/jqueryui/1.10.1/jquery-ui.min.js"></script>
```

If the browser is viewing that current page in through HTTPS, then it’ll request that asset with the HTTPS protocol, otherwise it’ll typically request it with HTTP.

Technically, this is called a “network-path reference” according to [RFC 3986!](http://tools.ietf.org/html/rfc3986#section-4.2).

willem010:
@lonalore 
that would also solve #805 
perfect!

OTroccaz:
Great, it works!
Thanks.

lonalore:
and also #1183 :)

Issue #1193, opened 2015-09-21
not translated string in install.php


Jimmi08:
Next hardcoded strings

![image](https://cloud.githubusercontent.com/assets/5429548/9995643/7f9c2db4-6083-11e5-980a-d3fb14ecdf02.png)

Jimmi08:
![image](https://cloud.githubusercontent.com/assets/5429548/9995708/d8a44b44-6083-11e5-894b-1a4e068ec0a1.png)

![image](https://cloud.githubusercontent.com/assets/5429548/9995716/e3da0bf2-6083-11e5-8fe5-9af86ae3143b.png)

Jimmi08:
install.php

![image](https://cloud.githubusercontent.com/assets/5429548/9997057/71eabef4-608a-11e5-8d1f-1a8e12a4182f.png)

![image](https://cloud.githubusercontent.com/assets/5429548/9997073/7f46a676-608a-11e5-9e91-fb81cbf0af92.png)

Jimmi08:
Used missing lan string

![image](https://cloud.githubusercontent.com/assets/5429548/9997122/ceb865dc-608a-11e5-9418-1b64219a88f7.png)

![image](https://cloud.githubusercontent.com/assets/5429548/9997103/b8d147fc-608a-11e5-91ae-51a82b5a5a6a.png)

Jimmi08:
Not used string (duplicate)
![image](https://cloud.githubusercontent.com/assets/5429548/9997703/f8595b96-608d-11e5-9252-a77bf76eff38.png)

This is replaced by:
![image](https://cloud.githubusercontent.com/assets/5429548/9997724/15164550-608e-11e5-91ef-2cc1068d241e.png)

Jimmi08:
Otherwise installation process without any problems.

tgtje:
Comment on LANINS_048 > no longer in use with that translattion the new one ( about 3 days old) is define("LANINS_048", "EXIF extension");

Jimmi08:
I used latest languages files from github (Czech version)
![image](https://cloud.githubusercontent.com/assets/5429548/10017642/89907aac-6130-11e5-9559-4047276edddd.png)

tgtje:
In older versions lanins 48 was decommented (not in use)<  > 3 days old now  > lanins_048 has new description.. The Czech version is 20 days, so does not have it...

Jimmi08:
[...]

PR #1233, opened 2015-11-14
Text to LAN's #issue6  #issue1193

tgtje:
use of & # 58 for the character : (colon)

do mention : all installs (prev versions too) language 'turnover' starts at Step 3 (or refresh pages from step 3 on).
covers issue #1193 too partially : hence (lang) working from step 3>  step 1 and step 2 translation is not seen > testing not possible..(refresh pages does not activate locale language use)


tgtje:
Found method to incorporate lan's as of step 2, will send new pull request
Info : when step 1 language is chosen and step 2 appears > refresh page > normally it will revert (browser should react asking for new 'info') back to step 1; but now should display the displayed text in chosen language.

PR #1234, opened 2015-11-15
Text to LAN's issues #6 and #1193

tgtje:
new pull request to cover most textual appearence.
As described normal installalion does not trunover the chosen language untill step 2 is reached and either  > go back is used OR refresh of page.. (going back from step 2 (lang was chosen on 1) puts step 1 in new lang mode..)

Note : 
logLines....  are NOT transformed to LAN's
template >SetTag..lines   .. are NOT transformed to LAN's
(edit : since this is v2 * > some of the new LAN's used;  where at a point in time in use in v1 version ** (not in recent v2 master files )).. as is.. using upgrade v1 to v2 implies using v2 language files!


tgtje:
have to close the commit, have found irregularity, need to test further before share

Issue #1177, opened 2015-09-04
Forum LAN_FORUM_2003


hanklu:
Version:e107 V2.0
Error: Forum LAN_FORUM_2003
Status:I see LAN in admin.
link:
e107_plugins/forum/forum_admin.php?mode=mods&action=list

tgtje:
Ok, a slipup (1 lan not called in forum php : (i will make a commit for it)
e107::lan('forum','front', true);  (forum_admin .php)
has to be in place so it can call the LAN (description) from English_front

tgtje:
https://github.com/e107inc/e107/pull/1178
please retest, thank you

hanklu:
Re-install github file. I see same information.
Link:e107_plugins/forum/forum_admin.php?mode=mods&action=list
Status:Forum LAN_FORUM_2003

I find other new:
Status:Forum WMGLAN_5
Link:e107_plugins/forum/forum_admin.php?mode=main&action=rules

Forum FORLAN_116
Link:e107_plugins/forum/forum_admin.php?mode=report&action=list

tgtje:
Ok, can not see the LAN_FORUM_2003 with new file, and do not notice the 
WMGLAN_5 (being Set Forum rules)

As for the FORLAN_116 (Reported Posts). happens now and then??, has to 
be more strict setting i guess
could you manually add (change) the following line in forum_admin.php ?  
(around line 20 )
e107::lan('forum', 'admin',true);
the piece >      ,true   is added) and recheck. if it works i can commit 
it.. (then at least 2 people would have checked it first)
Thank you

Moc:
With all the LAN changes I'm pretty sure this has been addressed by now. If not please re-open.

tgtje:
Yes, it was waiting for a response from the poster ( test)



---
Dit e-mailbericht is gecontroleerd op virussen met Avast antivirussoftware.
https://www.avast.com/antivirus

PR #1235, opened 2015-11-15
redefined LANs (install) #6 #1193 (latest)

tgtje:
Amended LAN defines>  third version;  as i found out that NOT all texts inside can carry LANs at certain positions > lines involved  up to 350.(install php)
 (most > die fatal error lines :  'hidden' when 5.3 is running and all is as should. 
Up to that point ( around 350) LANs are not being called ( testcase php 4.*  with added lan's display the lans, not the defines)...(as mentioned hidden if all is correct on newer version of php).

(Note : unsure for LAN 142,143,144 as i can  not access/trigger them)

Not translated are : 
logLines.... are NOT transformed to LAN's (reason : international sharing of logs when errors appear)
template >SetTag..lines .. are NOT transformed to LAN's.
( if i am correct and also above (log/TAG) is ok,, this would be the last revision for install php and language use). 

Second commit : forum  since no reaction and working method #1177 is covered  + update caption/help file

### DIFF
--- a/e107_languages/English/lan_installer.php
+++ b/e107_languages/English/lan_installer.php
@@ -171,3 +171,17 @@ define("LANINS_126", "For security reasons you should now set the file permissio
 define("LANINS_127", "The database [x] already exists. Overwrite it? (any existing data will be lost)"); 
 define("LANINS_128", "Overwrite");
 define("LANINS_129", "Database not found.");
+define("LANINS_130", "Installation");
+define("LANINS_131", "of");   //single time use installer only as in .1 of 8  not replacing by LAN_SEARCH_12
+define("LANINS_132", "Deleted existing database");
+define("LANINS_133", "Found existing database");
+define("LANINS_134", "Version");
+
+define("LANINS_141", "Please fill in the form below with your MySQL details. If you do not know this information, please contact your hosting provider. You may hover over each field for additional information.");
+define("LANINS_142", "IMPORTANT: Please rename e107.htaccess to .htaccess");
+define("LANINS_143", "DEBUG: Rename from e107.htaccess to .htaccess was successful");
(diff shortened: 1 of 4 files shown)


CaMer0n:
@tgtje Great stuff!! Just a couple of small things to change. 
1) LANINS_131 should be "[x] of [y]" . 
2) We don't make LANs for DEBUG information. so LANINS_143 should be removed.

tgtje:
@CaMer0n  no problem, but point 1 defend  > of is here meant as a common term where the SC's are already set.on their own..

[x] of [y] implements a choice> this is not so . A choice is already preset.

Hard to explain (translated) > example : 2 of a kind  (...of.. needs no explanation, only a translation. Of.here.. does not change ( number is always > 8  ('also' fixed)  )  SC 1 = Step  SC 2=2,3,4,,5,6,7,... OF.... 8
(hence decommented txt > single use) For each language only the 'wording 'is translated to locale, does not have an impact on coding.  Dutch of = van  German of = von  French of = de  Rus of = из  Swed of = af

CaMer0n:
@tgtje Okay, let's leave the [x] of [y] as it is for now. (not sure how it works with right-to-left languages). 

For point 2, only LAN 143 is debug info, the others (142 and 144) are errors. Would you mind fixing that? Then I'll commit this one.. cheers!

tgtje:
*Overworking no point, i will close these commits and **rewrite them again**.* ( i want to update all files to latest so i know i use no older pieces  (to eliminate any errors).
I will remove LAN 143. (def + coding)
As for the [x] of [y]  it (displayed as  install php line 2020 : Installation: {stage_pre} {stage_num} of 8
in changed form >'.LANINS_130.' &#58 {stage_pre} {stage_num} '.LANINS_131.' 8 </.....

Tough one to crack, since it is written by bar (template SetTag) that piece (correct me if i am wrong) would have RTL support already at that stage (lines 1282-1287) to re-arrange the {stage_pre} {stage_num} output  (arabic :  التركيب: الخطوة 3 من 8 )  = loosely 8 of 3 Step :Installation > 8 LANINS_131 {stage num} {step pre} LANINS_130..

PR #1305, opened 2016-01-17
amended LANs and issue #1177 17-01-2016 refreshed

tgtje:
rework on using fresh files ( coding as was in previous commit .with 1 change.)
Change was on prev commit for  define and coding  LANINS_143 (line 1355 /install php) > removed as being part of debug (no translation for message)
(typo in commentary header txt  for commit 2 > LAN_143 = LANINS_143; )!


tgtje:
Wow .. just discovered some defines are/were in use in db php..Closing and reworking

PR #1306, opened 2016-01-17
amended LANs and issue #1177 17-01-2016 refreshed

tgtje:
rework on fresh file  (unchanged in comparison. to old commit november)


CaMer0n:
Thanks! :+1:

PR #1310, opened 2016-01-25
BOOK_DESCRIPTION sc not returning proper value

Serios:
Fixing BOOK_DESCRIPTION sc not returning proper value on Chapters list due missing field in the db query


CaMer0n:
Thank you! :+1:

Issue #1309, opened 2016-01-23
Links with e_BASE in path doesn't work with SEF URL 


Jimmi08:
I noticed weird thing with my plugins 
~~(Just only with mine, so I do something wrong). The same with new plugin or with rewriting old one.~~

e_BASE  is empty. With other plugin (gallery, download) there si value ../..

I found the reason here (code is from e107_class.php). 
![image](https://cloud.githubusercontent.com/assets/5429548/12532979/0f576b58-c222-11e5-8cd6-4ff4b688cc41.png)

~~With other plugins currect directory is correct 
C:\xampp\htdocs\e107V2\e10720150123\e107_plugins\gallery
with my plugins 
C:\xampp\htdocs\e107V2\e10720150123 
but url is f.e
http://localhost/e107V2/e10720150123/links_page/links~~

Otherwise everything works. I noticed it just because using lastnews_menu first time and there is path done old way with e_BASE and with empty e_BASE is incorrect.

Without e_url.php path is correct, but e_url.php code seems to be allright, works very well. 

I don't know where or what to look. Any idea? Thanks

Jimmi08:
After more testing. Gallery plugin with enabled SEF URL has the same problem:
without SEF URL:
![image](https://cloud.githubusercontent.com/assets/5429548/12536562/44b5e54a-c2a9-11e5-9cb2-54a94b585d37.png)

with SEF URL:
![image](https://cloud.githubusercontent.com/assets/5429548/12536565/50b6ff3c-c2a9-11e5-9409-da9b25b11e52.png)

CaMer0n:
e_BASE is for files (includes etc.), not for URLs. The same for e_PLUGIN (file-system) vs e_PLUGIN_ABS (links). If you find them, please replace.

Jimmi08:
This code is in 12 files:
```<a href='".e_BASE."user.php```
I would replace it if I know how.

Jimmi08:
@CaMer0n waw, with this I think I am able to replace others. Just quick question, why did you use e107::getUrl()->create() and not e107::url()?

CaMer0n:
@Jimmi08 Some of the core is still using the older more complex/advanced method of creating SEF Urls. (news, page, search, system and user still use this). e107::url() is a newer, simpler developer-friendly method for doing the same thing.

Jimmi08:
[...]

PR #1312, opened 2016-01-27
Issue #1309. "user.php" in plugins replaced with SEF URL 

Jimmi08:


CaMer0n:
Nice work! Thank you! :+1:

PR #1313, opened 2016-01-30
Translation of hardcoded text, added new defines

tgtje:
Leftovers, added recently; translation for existing hardcoded text..


CaMer0n:
Nice work! Thank you! :+1:

PR #1316, opened 2016-02-01
Fixed: bug when you use "Quick Reply" twice (in forum).

lonalore:


lonalore:
Basically the problem was that e-token became invalid after submitting (first time) the form. So it was needed to create a new e-token and update it in the success callback.

PR #1319, opened 2016-02-01
Proper parsing of BOOK SC in List Chapters

Serios:
Book shortcodes doesn't get parsed in template-start and template-end portion of List Chapters template, which seems illogical.


CaMer0n:
Thank you! :+1:

Issue #1259, opened 2015-12-03
Extended JavaScript documentation 


lonalore:
# Classes and Methods
## Javascript

Including javascript in your plugin or theme may be achieved by using the following function: 

``` php
e107::js(type, value, dependency, zone);
```

_Example: Load a script in the 'faqs' plugin directory and auto-load jquery if not already loaded._

``` php
e107::js('faqs','js/faqs.js', 'jquery');
```

| Type | Value | Description |
| --- | --- | --- |
| core | path relative to the core folder | Include a core js file in the header of the page |
| url | full url to javascript filel | Include a remote js file in the header of the page |
| inline | javascript code | Include raw javascript code in the header of the page |
| theme | path to js file, relative to the current theme's folder | Include a theme js file |
| (any plugin folder name) | path to js file, relative to the plugin's folder | Include a plugin js file |
| settings | php array | Adds settings to e107's global storage of JavaScript settings. |

**'settings'**: An associative array with configuration options. The array is merged directly into e107.settings. All plugins should wrap their actual configuration settings in another variable to prevent conflicts in the e107.settings namespace. Items added with a string key will replace existing settings with that key; items with numeric array keys will be added to the existing settings array.
# Javascript
## JavaScript Behaviors

Behaviors are event-triggered actions that attach to page elements, enhancing default non-JavaScript UIs. Behaviors are registered in the e107.behaviors object using the method 'attach' and optionally also 'detach' as follows:

``` javascript
var e107 = e107 || {'settings': {}, 'behaviors': {}};

(function ($) {
  e107.behaviors.myBehavior = {
    attach: function (context, settings) {

    },
    detach: function (context, settings, trigger) {

    }
  };
})(jQuery);
```

e107.attachBehaviors is added to the jQuery ready event and so runs on initial page load. Developers implementing Ajax in their solutions should also call this function after new page content has been loaded, feeding in an element to be processed, in order to attach all behaviors to the new content.

See the **"e107_web/js/core/all.jquery.js"** file for more information.
## Using jQuery

jQuery is now namespaced to avoid conflicts with other Javascript libraries such as Prototype. All your code that expects to use jQuery as $ should be wrapped in an outer context like so.

``` javascript
(function ($) {
  // All your code here.
})(jQuery);
```

If you don't, you may see the error Uncaught TypeError: Property '$' of object [object DOMWindow] is not a function or similar.
## jQuery Once

[...]

PR #1322, opened 2016-02-04
Rewrited JS code to use e107.behaviors. Behavior to enable Smooth Scr…

lonalore:
Rewrited JS code to use e107.behaviors. 
Behavior to enable Smooth Scrolling if URL has a fragment.

### DIFF
--- a/e107_web/js/core/front.jquery.js
+++ b/e107_web/js/core/front.jquery.js
@@ -1,357 +1,486 @@
-/* global $ */
+var e107 = e107 || {'settings': {}, 'behaviors': {}};
 
-$(document).ready(function()
+(function ($)
 {
-		$(":input").tooltip();	
-	/*	
-		$(":input,label,.e-tip").each(function() {
-			
-			var field = $(this).nextAll(".field-help");
-		
-			if(field.length == 0)
+	// In case the page was opened with a hash, prevent jumping to it.
+	// http://stackoverflow.com/questions/3659072/how-to-disable-anchor-jump-when-loading-a-page
+	if(window.location.hash)
+	{
+		$('html, body').stop().animate({scrollTop: 0});
+	}
+
+	/**
+	 * Behavior to initialize Smooth Scrolling on document, if URL has a fragment.
+	 *
+	 * @type {{attach: Function}}
+	 */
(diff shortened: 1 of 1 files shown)


lonalore:
I think we would provide ability for the user to customize "Smooth Scrolling". For example: enable/disable it, to change duration for animation, and most importantly, to set a top-offset in the case theme has a fixed top-navigation bar.

I thought that we would create additional tab called "Global preferences" in theme-manager:
kepernyokep - 2016-02-04 - 09 44 42

We can send these settings to front-end using e107::js('settings', ...);

... so users can customize the JS behavior without overriding the JS code.

Moc:
Great addition. If this gets merged in, please document it so people know how to use it! :)

lonalore:
The "global" Smooth Scrolling was @CaMer0n's idea. :)
Here is a documentation about JavaScript Behaviors: https://github.com/e107inc/e107/issues/1259
(If grammar is not correct, please fix it...)

Moc:
Perfect, thanks!

CaMer0n:
@lonalore Great work! However I'm a bit nervous about such a large change to the front-end javascript, at this time. I believe this pull-request should be merged AFTER 2.0.1 so it can be tested in different environments for some weeks/months before being in the following official release. A few extra lines for the scroll animation would have been fine, but this is a rewrite of the entire file - like your subject says.  :-)

lonalore:
Thank You! :) Anyway, almost the whole code is unchanged inside the behaviors, I just copied them from the "document ready" event. But okey, it can wait for the next release, because as I noticed, the "comment" section of the code is heavily buggy. The "Edit" button does nothing, The "Vote Up/Down" and "Reply" buttons can be used only once. So it needs of thorough testing and rewriting.

CaMer0n:
Thanks. Only one vote per use was intended. ;-) The 'edit' button seems to be failing though.

lonalore:
Ohh, so after voting (or a wrong click), user cannot change it? :P :D

lonalore:
I created a new pull-request only with the "Smooth Scrolling". #1338

CaMer0n:
Thank you!, will commit it now.

PR #1354, opened 2016-02-13
Library manager

lonalore:


CaMer0n:
I will make some changes and comments. It may break some things, but I'll sure you'll be able to fix. Thanks :+1:

PR #1362, opened 2016-02-15
Fix for non-Latin characters in textareas

Serios:
Fixing and issue with non-Latin characters breaking in textareas if html tags are used

### DIFF
--- a/e107_handlers/e_parse_class.php
+++ b/e107_handlers/e_parse_class.php
@@ -4139,7 +4139,7 @@ return;
 	//    libxml_use_internal_errors(true); // hides errors.
         $doc  = $this->domObj;
 	    libxml_use_internal_errors(true);
-        @$doc->loadHTML($html);
+        @$doc->loadHTML($html, 'HTML-ENTITIES', 'UTF-8');
 		// $doc->encoding = 'UTF-8';
 
      //   $doc->resolveExternals = true;


CaMer0n:
@Serios Thank you! Could you elaborate a little more.. was this a problem when using TinyMce? How did you test it?

CaMer0n:
btw: https://gitter.im/e107inc/e107

CaMer0n:
@Serios Unfortunately, I had to revert this change as it was emptying out the content of the TinyMce text area when loading a record in the admin area. Others using non-latin characters are reporting that there is no issue. Could you share more about your database charset, server info etc?

Serios:
@CaMer0n if I use non-latin characters with html code it's fine. But when Cyrillic is imputed even if TinyMce is disabled the text returned form the DB is scrambled for example "ÐÐ°Ð¹Ð»" it's saved scrabled in the DB. 

The collation of DB is set to utf8_general_ci. It's fresh install. In e107_confiig.php $mySQLcharset is set to utf8 too. 

Cyrillic gets scrambled only if html tags are added too, otherwise works just fine, so I traced the problem to cleanHtml function in e_parse_class.

CaMer0n:
Thank you @Serios for your help with that. Yes please, if that "EDIT 3" solution does not cause any issues with loading/saving content in Cyrillic and English, please go ahead and submit a pull request. Thank you! :+1:

Serios:
Well I tested on my local and dedicated servers and was working fine :) Pull request on the way...

PR #1376, opened 2016-02-19
Field type "datepicker" enhancements: allow to use "disabled" attribu…

lonalore:
Field type "datepicker" enhancements: allow to use "disabled" attribute, and convert (default) value to the proper format.


CaMer0n:
Will take a look at this on Wednesday. Thanks

lonalore:
Okey, thank you!

PR #1384, opened 2016-02-25
Fixing http_build_query causing return of &amp

Serios:
Fixing http_build_query function causing return of &amp instead of & on some servers when using {MENU: xxxxx} with parameters, thus breaking parameters passing to the menu.

Eg. {MENU: path=news/latestnews&test1=true&test2=false&test3=44}
Will return call the menu with test1=true&amptest2=false&amptest3=44 instead of est1=true&test2=false&test3=44
 causing only the first parameter key passed, to be valid


CaMer0n:
Thank you! :+1:

Issue #1342, opened 2016-02-09
Cleanup: Replace all hardcoded <img> tags with $tp->toImage();


CaMer0n:
<img> tags are inconsistent throughout the core.
For icons we already have toIcon() but for everything else they are currently hard-coded.
Having a central method for rendering will provide enhancements for rendering on hi-res mobile devices/tablets as well as easier updates. (eg. 'img-responsive' will become 'img-fluid' in Bootstrap 4).

I understand this is a HUGE task, so pull-requests are VERY welcome. Just focus on one plugin/area at a time to avoid merge conflicts and test every change to ensure the image is not broken or lost some functionality. 

@Moc @lonalore @tgtje @Jimmi08

Jimmi08:
I take Gallery plugin. Did you mean this?
![image](https://cloud.githubusercontent.com/assets/5429548/12928276/cc51fdfc-cf6c-11e5-84e2-d4d6cd3a1ecf.png)

CaMer0n:
@Jimmi08 Yes, something like that. But $att needs to be merged with array('alt'=>'whatever');

Jimmi08:
@CaMer0n 
I did it by e_rss.php. I don't see diffs. What I am missing?
![image](https://cloud.githubusercontent.com/assets/5429548/12928962/b6a05216-cf70-11e5-9d13-7f2ac354c68f.png)

CaMer0n:
gallery parses some special parms to thumbUrl(), which are in $att. So you need to build on that array by adding alt and class.

Jimmi08:
Yes, I know what you mean. There are reasons why I can't do it and I let it to you  (I can be wrong, of course, I wrote this just to show that I tried :) )
- paths in gallery images (default from installation) includes {e_PLUGIN}, but thumbUrl() is called only for path with e_THEME or e_MEDIA, so they never be hashed with toImage();

-$att is not array in gallery plugin 
![image](https://cloud.githubusercontent.com/assets/5429548/12930334/ecc06e56-cf77-11e5-8af1-356658b7a4c5.png)
and toImage() needs width and height as separate array fields

- toImage() doesn't use x parameter (to be hashed path) . ThumbUrl() is called with options=null, others options are setted by f.e.  $tp->thumbWidth(), but to get hashed path, option x is still needed to be set somehow. 

When somebody other will do this, I will see where I am wrong. 

[...]

PR #1386, opened 2016-02-26
Cleanup: Replace all hardcoded <img> tags with $tp->toImage(); #1342

Jimmi08:
to finish cleaning gallery plugin


CaMer0n:
Thanks. Unfortunately, it looks like "data-src='holder.js/".$w."x".$h."'" was removed. You should be able to add it as a param. If not, please let me know.

Jimmi08:
Sorry, I missed this. But I don't see what param to use in toImage():
image

CaMer0n:
@Jimmi08 I'll need to add something in there for data-xxxx attributes.

PR #1404, opened 2016-03-02
New Ajax API.

lonalore:
As I promised, here is a new (full-featured) Ajax API, which can be used for all kind of HTML elements. The new Ajax API is backward compatible, so it supports the old ways of usage.

**Ajax API usages in core (HTML elements with ".e-ajax" class and "data-xxx" attributes):**
- "Create" button on page: "Tools / Database / Backup" **(Tested, works properly)**
- "Next/Prev" links on news pages for comment-pagination **(Tested, works properly)**
- "search.php" dropdown **(Tested, works properly)**

**Custom usage:**
- "Generate" button to get a random encryption key. [See code here.](https://github.com/lonalore/backup_dropbox/blob/master/admin_config.php)

**New feature:**
Now, we can use multiple (different kind of) commands in a single response for an Ajax request.

**Supported commands:**
- **'insert'** - Command to insert new content into the DOM.
  - **'replaceWith'**
  - **'append'**
  - **'prepend'**
  - **'before'**
  - **'after'**
  - **'html'**
- **'remove'** - Command to remove a chunk from the page.
- **'alert'** - Command to provide an alert.
- **'css'** - Command to provide the jQuery css() function.
- **'settings'** - Command to set the settings that will be used for other commands in this response.
- **'data'** - Command to attach data using jQuery's data API.
- **'invoke'** - Command to apply a jQuery method. The 'invoke' command will instruct the client to invoke the given jQuery method with the supplied arguments on the elements matched by the given target. Intended for simple jQuery commands, such as attr(), addClass(), removeClass(), toggleClass(), etc.

**Example to build Ajax response (with multiple commands):**
```php
$ajax = e107::getAjax();

$commands = array();
// Merge array into e107.settings javascript object.
$commands[] = $ajax->commandSettings(array('foo' => 'bar'));
// Remove 'disabled' attribute from the '#object-1' element.
$commands[] = $ajax->commandInvoke('#object-1', 'removeAttr', array('disabled'));
// Insert HTML content into the '#object-1' element.
$commands[] = $ajax->commandInsert('#object-1', 'html', 'some html content');

// This method echos data in JSON format. It sets the header for JavaScript output.
$ajax->response($commands);
```


lonalore:
The last step (that remains): allow to use Ajax API (where it is possible) in **form_handler.php** (text. textarea, password, checkbox, radio, etc...). Ajax API is only allowed for "dropdowns" at this time.

lonalore:
Hi @CaMer0n! If you have time, please check this out. I hope you find it good. :) Thank you!

CaMer0n:
@lonalore It looks great! The only thing that needs to change is the use of e107::ajax to load the object. This format: `e107::xxxx(foo,bar)` is for static functions that don't return objects. Since it is returning an object,  it should follow the standard and use: `$ajax = e107::getAjax(); `

lonalore:
@CaMer0n thank you! I changed the method name. :)

CaMer0n:
@lonalore Thank you! For the Docs, you should now be able to add them on e107.org yourself directly.

lonalore:
Wow, thank you! :)

lonalore:
@CaMer0n it doesn't work. :/ 

kepernyokep - 2016-03-08 - 10 53 17

CaMer0n:
Sounds like some of the code was blocked due to mod_security. Try a very small change, and see if it is accepted.

lonalore:
It's still the same: Error 500 - Malformed Header :/ 

I tried to insert this (into a "pre" block):
```javascript
var e107 = e107 || {'settings': {}, 'behaviors': {}};

(function ($) {
  e107.behaviors.myBehavior = {
    attach: function (context, settings) {

    },
    detach: function (context, settings, trigger) {

    }
  };
})(jQuery);
```

PR #1410, opened 2016-03-09
Ability to use CDN files for libraries.

lonalore:


CaMer0n:
How does this work with the JS/CSS cache?

lonalore:
I think it uses "System cahe" now, and not "JS/CSS cache". Because it only caches the information arrays provided by config() methods in e_library.php files. Before caching, it completes these information arrays with the version numbers (retrieved from the specified library files). Without caching, it would always read the files to get version number, and it would cause performance issues.

You can find caching here:
https://github.com/e107inc/e107/blob/master/e107_handlers/library_manager.php#L300-L314

I think, the best choice would be that we provide a "Library cache" category for Library Manager.

PR #1424, opened 2016-03-14
Update user_template.php

mcpeace-maw:
It is a try that need some fixes, lan texts render.... The header background image can be blurry if user image/avatar is small, so I think maybe is a better solution to use a background color on the header, a background color by user category maybe.


CaMer0n:
@mcpeace-maw Closing this in favor of the more recent pull-request.

Issue #1459, opened 2016-03-27
installation v2.1 in national language


Jimmi08:
Hi, what I am doing wrong?

Default screen in English
![image](https://cloud.githubusercontent.com/assets/5429548/14064541/2bdef61c-f407-11e5-9e1e-50587a5f1fa8.png)

Select my language, button Continue
![image](https://cloud.githubusercontent.com/assets/5429548/14064542/4772dd4e-f407-11e5-93a9-f7f328694f9a.png)

Step is in English, go back
![image](https://cloud.githubusercontent.com/assets/5429548/14064543/566c54d8-f407-11e5-861b-66cf65e60c88.png)

Step 1 is in Slovak, but selected language is English. After repeating select language, installation continure correctly.

Thanks

tgtje:
It was mentioned already a few times before (behaviour). 
You are not doing something wrong... install kicks in at step 3 
(and you can THAN go back and step 2 (sometimes 1) is in your native language... (like it has to first write to a temp thing (file or db ..??) and than is overruled by language, could not figure it out by myself yet).

Jimmi08:
yes, you are right. Should I close this?

Moc:
No, this needs sounds as if it needs some improvements.

CaMer0n:
@Jimmi08 Maybe you can test and solve it? Probably just a missing line somewhere. try using this url: install.php?debug

tgtje:
It (imo) should be equal to v1 ; at step 1 when changing language, it should act at the moment. ( page refresh and step 1 complete in langauge chosen).

Jimmi08:
https://github.com/e107inc/e107/pull/1464

PR #1464, opened 2016-03-27
Fix for wrong language in stage 2 #1459

Jimmi08:
This test was in stage_2() function, it was too late.


CaMer0n:
I knew you'd find it! Thanks @Jimmi08 !!

PR #1469, opened 2016-03-29
Gallery plugin fixes and enhancements

lonalore:
- prettyPhoto library moved to "e107_web/lib" folder (from Gallery's folder)
- jQuery Cycle library moved to "e107_web/lib" folder (from Gallery's folder)
- Implement e_library.php to provide information about third-party libraries
- Use Library Manager to load prettyPhoto and jQuery Cycle files
- Removed JS codes from PHP files, moved to separated JS files, and use JS behaviors
- Additional plugin pref: "Order images by" (Media ID ASC/DESC, Media Name ASC/DESC, Media Caption ASC/DESC)
- Use LAN files ([LANGUAGE]_global, _admin, _front), moved hardcoded texts to LAN files
- Use "Images per page" pref instead of hardcoded amount


CaMer0n:
Thank you!! I'm going to hold-off on the merge until after v2.1.1 is out (which should be mostly a bug-fix) in the next couple of days.

lonalore:
Okay, until then, I'll add more prefs to customize prettyPhoto and jQuery.Cycle. For example deep-linking setting and theme support for prettyPhoto, 'pause on hover' setting for Cycle... etc.

lonalore:
Under "prettyPhoto settings" tab, it is now possible to completely configure prettyPhoto:
screencapture-demo-lonalore-hu-e107v2-e107_plugins-gallery-admin_gallery-php-1459340571251

lonalore:
Done! :)

CaMer0n:
Great work!! Just checking - did you test prettyPhoto with the {PORTFOLIO} shortcode in the 'marketing' layout of Bootstrap3? And the JS/CSS only loads when needed, correct?
Also, it would be good to replace the 'markup' key used in the new templates. All templates are 'markup', so "item" or some other term used elsewhere in other core templates would be better.  These are small things. I haven't installed it yet.. but I looked through the code and it looks good.

lonalore:
> did you test prettyPhoto with the {PORTFOLIO} shortcode in the 'marketing' layout of Bootstrap3?

No, I didn't. But I'll test it tomorrow.

> And the JS/CSS only loads when needed, correct?

Originally, prettyPhoto was always loaded by **e_shortcode.php**. But I removed JS/CSS loading and I created an option (**Load prettyPhoto globally**) to let the user be able to choose... :)

> it would be good to replace the 'markup' key used in the new templates.

I decided to use prettyPhoto's parameters (from [documentation](http://www.no-margin-for-errors.com/projects/prettyphoto-jquery-lightbox-clone/documentation)) as template keys. I thought it would be clear to users what exactly should be overwritten. But I'll rename keys tomorrow, it ultimately doesn't matter. :)

Thank You! :)

lonalore:
It's done. :)
I tested with {GALLERY_PORTFOLIO}, it works:
kepernyokep - 2016-03-31 - 09 10 16

But there is an issue about thumb.php. It generates images without extension like:
`thumb.php?id=c3JjPWVfUExVR0lOJTJGZ2FsbGVyeSUyRmltYWdlcyUyRmxha2UtYW5kLWZvcmVzdC5qcGcmYW1wO3c9MTAyNCZhbXA7aD03Njg=`

So prettyPhoto cannot display thumbnails if "Overlay gallery" is enabled. [Issue is described here...](http://stackoverflow.com/a/27419571/2416835)

lonalore:
Anyway, you can test it here: http://demo.lonalore.hu/e107v2/
I sent account details on gitter some days ago.

PR #1471, opened 2016-03-29
Change newscategory shortcode

rica-carv:
If i'm allowed, i would like to suggest that the NEWSCATEGORY shortcode could also use the nolink parm as is used by NEWS AUTHOR shortcode....

### DIFF
--- a/e107_core/shortcodes/batch/news_shortcodes.php
+++ b/e107_core/shortcodes/batch/news_shortcodes.php
@@ -308,11 +308,17 @@ class news_shortcodes extends e_shortcode
 	function sc_newscategory($parm=null)
 	{
 		$category_name = e107::getParser()->toHTML($this->news_item['category_name'], FALSE ,'defs');
-		$category = array('id' => $this->news_item['category_id'], 'name' => $this->news_item['category_sef'] );
-		$categoryClass = varset($GLOBALS['NEWS_CSSMODE'],'');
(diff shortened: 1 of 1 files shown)


Jimmi08:
Doesn't do {NEWS_CATEGORY_NAME} exactly the same thing? Just replace {NEWSCATEGORY} with it in your template. 
PS. I just made update my list of these shortcodes and noticed new ones, I wouldn't know this otherwise.

CaMer0n:
Yes, closing this. I made a change to the new shortcode. Use {NEWS_CATEGORY_NAME: link=1} for a link, or just {NEWS_CATEGORY_NAME} for no link. 
https://github.com/e107inc/e107/commit/1281c70714985f0c9f8343c5b06a6bb088d2a8ef

rica-carv:
Ok, thanks, i didn't noticed that neither...
So i can assume that the {NEWSCATEGORY} shortcode is only present for legacy and backward compatibility purposes, right?

rica-carv:
If im allowed another suggestion, why not also merge this way (ie: {NEWSCOMMENTS: link=1}) the two shortcodes {NEWSCOMMENTS} + {NEWSCOMMENTLINK}????
Or even better, the 3: {NEWSCOMMENTS} + {NEWSCOMMENTLINK} + {NEWSCOMMENTCOUNT} ???
As far i see it, both 3 situations could be handled via param, no?

PR #1475, opened 2016-03-30
Avoid "Missing file" message for deprecated files in language-Packs verification

OTroccaz:
Deprecated files were taken into account only for language other than
English


CaMer0n:
Thank you!

PR #1510, opened 2016-04-09
reworked : using language folder + use of standard e107::lan

tgtje:
BS3 language use : site pref tab ; 1 thing at a time to keep up with changes ;-)

### DIFF
--- a/e107_themes/bootstrap3/languages/English_admin.php
+++ b/e107_themes/bootstrap3/languages/English_admin.php
@@ -1,0 +1,23 @@
+<?php
+/*
++ ----------------------------------------------------------------------------+
+|     e107 website system - 
+|
+
++----------------------------------------------------------------------------+
+*/
+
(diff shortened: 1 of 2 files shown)


CaMer0n:
Thank you!

tgtje:
@Cameron. Just discovered (pointed out by third party) that allthough everything SEEMS to work as intended the new LANs for left/right; top/bottom  do create the login/user to disappear when used.
AKA lan does de-activate (does not react) on NAVBAR.  I am looking for a solution, but get confused why $brandingopts does work and rest not.

Thinking if the 2 arrays could be called by another $brandingopts2 and $brandingopts3 which replace the arrays might do it. But i can not/know not (lack of coding) how to adjust the file to acccept those 2 new strings ( so my contribution turns out to be buggy).

cameron:
fyi, @CaMer0n != @cameron :)

Issue #1514, opened 2016-04-10
signup shortcodes  -  ignoring latest changes, 


Jimmi08:
New user registration system has 3 values (0,1,2)
![image](https://cloud.githubusercontent.com/assets/5429548/14409435/9cb90866-ff13-11e5-8026-8540de579d5c.png)
But in signup shortcodes (login table or login menu) there are tests 
	`if ($pref['user_reg'])`
       ` if(empty($this->userReg))`
so button is displayed for login only setting

Jimmi08:
There left just one place where signup link is always displayed.
![image](https://cloud.githubusercontent.com/assets/5429548/14573909/93a04ebc-0358-11e6-8c68-6f02b229140f.png)

![image](https://cloud.githubusercontent.com/assets/5429548/14573916/9f6324c2-0358-11e6-8d94-621c7665185b.png)

Log in and Sign up is in one lang string.

Jimmi08:
@CaMer0n Can I split CHATBOX_L3  to solve last place with this problem?

CaMer0n:
Thanks @Jimmi08 . Yes please..2 new LANs. and please be sure to use [ ] 
ie. "Please either sign in or if you are not registered click [here] to sign up".

Jimmi08:
@CaMer0n  Do you mean this?
```
define("CHATBOX_L3", "You must be logged in to post comments on this site - please either log in from the Login box or from [here].");
define("CHATBOX_L3b", "If you are not registered click [here] to signup");
```

```
	if($pref['user_reg'])
	{
		$text1 = str_replace(array('[',']'), array("<a href='".e_BASE."login.php'>", "</a>"), CHATBOX_L3);
		if($pref['user_reg'] == 1 ){
		  $text1 .= str_replace(array('[',']'), array("<a href='".e_BASE."signup.php'>", "</a>"), CHATBOX_L3b);
		}
		$texta = "<div style='text-align:center'>".$text1."</div><br /><br />";
	}
```

results:
![image](https://cloud.githubusercontent.com/assets/5429548/14930122/d2367908-0e61-11e6-8d65-b77e58ffc5ed.png)
![image](https://cloud.githubusercontent.com/assets/5429548/14930129/dc45f752-0e61-11e6-9767-5871c860274c.png)

CaMer0n:
@Jimmi08 Yes, exactly!. Just be sure to use `e_SIGNUP` and `e_LOGIN` for the urls instead of  `e_BASE."login.php` etc.

PR #1542, opened 2016-04-15
signup link only for value 1, part fix for  #1514

Jimmi08:


Jimmi08:
https://github.com/e107inc/e107/issues/1514

PR #1545, opened 2016-04-15
hardcoded/not translated string in login.php FIX #1465

### LINKED ISSUE #1465: hardcoded/not translated string in login.php
In login.php is not translated message:

image

Can I add new string LAN_LOGIN_32 in lan_login.php ?

Jimmi08:


CaMer0n:
It is also used on signup and forgot password. (if not, it should be ). Yes, please add to English.php 
LAN_LOGGED_IN

PR #1546, opened 2016-04-15
#1515 - hardcoded string in e107_admin/prefs.php 

### LINKED ISSUE #1515: hardcoded/not translated string in e107_admin/prefs.php
Result:
image

Reason:
`e107::getNav()->admin("Basic ".LAN_OPTIONS.'--id--prev_nav', 'core-prefs-main', $var);`

Jimmi08:


tgtje:
Sorry to mention ; Basic is already a define (lan_admin php line 426) Just not adapted there...

>?   	e107::getNav()->admin(LAN_BASIC .LAN_OPTIONS.'--id--prev_nav', 'core-prefs-main', $var);

Jimmi08:
I've already checked this, but it's not goog way (to combine strings). Two reasons:

1. Second string begins with capital letter, maybe it's good in English (Basic Options), but translation in my language is Základné možnosti (low letter for Options).  But this is heading, I can live with big letter, it's not correct, but understable.

2. Basic has 4 translantion in my language (as heading, in sentence there is more forms). Result depends on the word that is going after it, so it's hard to tell how to translate it without context.  (I think it's called in English Descriptive Adjectives).   The worst is if fist word is translated as single and second word is in plural and masculine gender - combine them and user tell you go to back to school.  Oh, almost forgot, basic can be noun too, again different translation.   

I checked using LAN_BASIC, it's used only for tabs, what is correct. I can use the same translation for options, so can be solved for now with your solution.  

Let me know. I will pull new request.

tgtje:
I personally have no problems with the new LAN, and translations. There 
are points in your comment..
I merely made comment that Basic is in use.. (and how it would be used 
normally)
All in concept of keeping the amount of LAN's small. (arrays will not be 
sufficient for this, because of same symptom)

As going through the files, it does and will happen more, but this will 
be always a though one.
Adjusting a word to a 'general' usuable code, which is language correct 
for many languages.
Will it ne netiquette, or proper use..
Of course your new LAN is correct for use, and why not, hence comment 
(obvious for endresult; maybe even better)

---

Jimmi08:
And I have no problems with discussion, because this is really tricky (you solve my problem, than with next lang you will new one). General solution (maybe) is in reordering LAN_OPTIONS:  LAN_BASIC.  There is only Russian trans with similar grammar and they haven't translated this yet.

tgtje:
Ha... why not out of the box : a Font that is small, and uses small 
capitals all the time... problem solved ;-)


---

rica-carv:
@tgtje Or use [php ucfirst](http://php.net/manual/en/function.ucfirst.php)...

tgtje:
@rica-carv similar as discussion based, will not function as wished...
Take German for example; each noun has a capital i.e. das Ungeheuer...  
So however usefull function, will not be correct spelled/written.
Helas.
The font issue is a small joke ;-)  if everything is capital, it will 
not matter.....

---

CaMer0n:
The menu title should just be "Preferences".

PR #1528, opened 2016-04-12
Minor changes to newscategory and download_shortcode, updated core plugins news icons

rica-carv:

PR #1551, opened 2016-04-16
#1528 Update newscategory shortcode

rica-carv:
#1528 Proposal to enable the newscategory shortcode to be displayed with or without links...


rica-carv:
Link hardcoded tag maybe candidate for templating???

CaMer0n:
I believe we should focus on the newer {NEWS_CATEGORY_NAME} and use 'url' and 'tag' modes. (see other shortcodes for similar behavior) . The default should be no link.

rica-carv:
@CaMer0n I focused on the newscategory shortcode because the [news_template.php](https://github.com/e107inc/e107/blob/master/e107_plugins/news/templates/news_template.php) only includes the newscategory shortcode, and doesn't include the newest news_category_name....

PR #1554, opened 2016-04-16
Add img-rounded class to images

rica-carv:
Add img-rounded class to images, for site wide image format consistency (like classes used for example by plugin news images...)

### DIFF
--- a/e107_plugins/banner/e_shortcode.php
+++ b/e107_plugins/banner/e_shortcode.php
@@ -99,12 +99,12 @@ class banner_shortcodes extends e_shortcode
 					if($row['banner_image'][0] == '{')
 					{
 						$src = $row['banner_image'];
-						$ban_ret = $tp->toImage($src,   array('class'=>'e-banner img-responsive', 'alt'=>$row['banner_clickurl']));
+						$ban_ret = $tp->toImage($src,   array('class'=>'e-banner img-responsive img-rounded', 'alt'=>$row['banner_clickurl']));
 					}
 					else
 					{
 						$src = e_IMAGE_ABS.'banners/'.$row['banner_image'];
-						$ban_ret = "<img class='e-banner img-responsive' src='".$src."' alt='".$row['banner_clickurl']."' style='border:0' />";
+						$ban_ret = "<img class='e-banner img-responsive img-rounded' src='".$src."' alt='".$row['banner_clickurl']."' style='border:0' />";
 					}
 
 


Jimmi08:
Mayve is better to add class parameter  to shortcode, no next hardcoded html code.Look at gallery thumb  img for examle. You can set default value this way but there needs to be way how to use custom classes.

rica-carv:
@Jimmi08 I really agree with you, but that would require heavy code change. I can do it, if i have clearance for it. I would even go a step further and sugest a bootstrap check (i that's even possible inside the shortcode...) to return or no the bootstrap specific img- classes...

Jimmi08:
@rica-carv   just add $class as param, and if it's not filled, use default.  Look here: https://github.com/e107inc/e107/pull/1555/commits/42a4e9c38b49f95f53e77893ba629edad370146c

rica-carv:
@Jimmi08 Original shortcode doesn't use $parms, used $row, so got to do some variable changes....
Working on it...

Jimmi08:
@rica-carv see it now, I checked first shortcode, sorry.

rica-carv:
@Jimmi08 Really some work here, got to change some code... eheh

CaMer0n:
Please post screenshots of the default banners being displayed using themes 'jayya' and 'core' from e107 v1.x  with these code changes. Thanks.

rica-carv:
@Jimmi08 Line 84 on this file is a good candidate for your hardcoded language english strings hunt... LOL

rica-carv:
@cameron v1.4 'Jayya' theme + banner
jayya

The micro thumbnail render is normal render of the banner, since i can only work around it (without modify the php code) by using the `{SETIMAGE: w=468&h=60}` shortcode before the `{BANNER}` shortcode ([theme.php, line 184](https://github.com/rica-carv/e107_theme-selospt_v2/blob/master/theme.php))...

The v1.4 'core' theme i can't use it, since my 'core' theme folder only has 3 files: `admin_alt_nav.sc, admins_style.css & usersettings_template.php`...

CaMer0n:
@rica-carv This is why I only used `toImage() `when `if($row['banner_image'][0] == '{')` 
The full-path specification was introduced in v2.x

rica-carv:
@CaMer0n Ok, but what's the issue with my code??
The code currently present on the master e107 renders the banner image exactly the same way as mine does (a micro banner with the link).
And my code also only uses the full path if the `if($row['banner_image'][0] == '{'` is not met : 
`if($row['banner_image'][0] == '{')
 					{
 						$src = $row['banner_image'];
 					}
 					else
 					{
 						$src = e_IMAGE_ABS.'banners/'.$row['banner_image'];
						$style = "'border:0'";
 					}
						$class = vartrue($parms['class']) ? $parms['class'] : "e-banner img-responsive img-rounded";
						$ban_ret = $tp->toImage($src,   array('class'=>$class, 'alt'=>$row['banner_clickurl'], 'style'=>$style));
 				break;
 			}`

I only put the `toImage `render outside the` if loop `to remove redundant code.....

CaMer0n:
Okay great. So if you remove the full path from your banner table. (as it would be coming from v1.x) Does it render the full-size image?

rica-carv:
@CaMer0n Yes, it does render the full size image.
Original code:
untitled 1

My code:
untitled 2

By the way, why does the new added banners are stored in another location (ie: `e107_media\xxxx\images\yyyyyy` ????
Only the original plugin banner image is stored under `e107_plugins/banner/images` ....
I can see this is due to the use of media manager, but can't this be more consistent (ie, store banner images in `e107_media` or `e107_plugins/banner` ) instead of having banners images scattered around???

CaMer0n:
All custom content is stored in e107_media and e107_system. It is more consistent than media spread through plugin folders. This is especially true when e107 is running multiple sites from the same script files.

rica-carv:
@cameron Following my comment [here](https://github.com/e107inc/e107/pull/1555#issuecomment-211539827), my changes implies two ways of sending the parm to the shortcode:
The old way, single parm: {BANNER=campaign} & new way, multiple parm as array {BANNER: campaign=value&class=value}....
Therefore, i need to check if $parm is array or string, and process accordingly, right?

CaMer0n:
That's correct @rica-carv  :+1:

PR #1555, opened 2016-04-17
enhancement for better templating

Jimmi08:
example of use:
`{LOGIN_TABLE_USERNAME: class=form-control input-lg}`

### DIFF
--- a/e107_core/shortcodes/batch/login_shortcodes.php
+++ b/e107_core/shortcodes/batch/login_shortcodes.php
@@ -56,6 +56,7 @@ class login_shortcodes extends e_shortcode
 		$placeholder =  $ulabel[$allowEmailLogin];	
 		
 		$parms = eHelper::scParams($parm);
+		$class = vartrue($parms['class']) ? "class='".$parms['class']."'" : "class='tbox form-control input-block-level'";
 		return "<input ".$class." type='text' name='username' id='username' size='40' maxlength='100' placeholder=\"".$placeholder."\"  />";
 
 	}
--- a/e107_core/shortcodes/batch/login_shortcodes.php
+++ b/e107_core/shortcodes/batch/login_shortcodes.php
@@ -68,8 +68,11 @@ class login_shortcodes extends e_shortcode
 			return null;
 		}
 
+		$parms = eHelper::scParams($parm);
+		$class = vartrue($parms['class']) ? "class='".$parms['class']."'" : "class='tbox form-control input-block-level'";
+
 		$pref = e107::getPref();
-		$text = "<input class='tbox form-control input-block-level' type='password' name='userpass' id='userpass' size='40' maxlength='100' placeholder=\"".LAN_LOGIN_2."\" />";
+		$text = "<input ".$class." type='password' name='userpass' id='userpass' size='40' maxlength='100' placeholder=\"".LAN_LOGIN_2."\" />";
 		
 		if (!USER && e107::getSession()->is('challenge') && varset($pref['password_CHAP'],0)) 
 		{
--- a/e107_core/shortcodes/batch/login_shortcodes.php
+++ b/e107_core/shortcodes/batch/login_shortcodes.php
@@ -55,8 +55,9 @@ class login_shortcodes extends e_shortcode
 		$ulabel = array(LAN_LOGIN_1,LAN_LOGIN_28,LAN_LOGIN_29);
 		$placeholder =  $ulabel[$allowEmailLogin];	
 		
-		
-		return "<input class='tbox form-control input-block-level' type='text' name='username' id='username' size='40' maxlength='100' placeholder=\"".$placeholder."\"  />";
+		$parms = eHelper::scParams($parm);
+		return "<input ".$class." type='text' name='username' id='username' size='40' maxlength='100' placeholder=\"".$placeholder."\"  />";
+
 	}
 	
 	function sc_login_table_password($parm='') //FIXME use $frm


CaMer0n:
Thanks @Jimmi08 Could you please test your code with this line removed: 
`$parms = eHelper::scParams($parm);`
(replaced with `$parms = $parm;` to test)

When we use the format `{XXXXXX: key=value}` $parm is always an array. Normally, there's no extra processing required.

Jimmi08:
@CaMer0n 
It looks it works. I noticed this code in gallery shortcodes, so I thought it's needed. Or maybe I should use them different way in template?

You can close this, I will create new request but I would like to use the same way for signup shortcodes - so I could try to change this to use form-handler.

But how to pass custom class to signup shortcodes?  ( I know how to pass it futher to form handler, but can I add $parm as input parameter to some signup shortcodes?) Otherwise they work, just I need to use my own classes.

CaMer0n:
Yes, you can add `$parm` to shortcodes that don't have it. For those that do, and are being utilized, you may need to add a check with `is_string()` and do a `parse_str()` to make sure the `$parm` is always an array.  Such as in the case of using this format: `{SHORTCODE=xxxx}  `  You can find other shortcodes in the core that do this check.

Jimmi08:
@CaMer0n  Do you mean this:

```
	if(is_string($parm))
	{
		parse_str($parm, $parms);
	}
	else
	{
		$parms = $parm;
	}
```

a then using $parms[key]  ?

CaMer0n:
@Jimmi08 Yes, that the code! :+1:    It is only required if you find a shortcode using $parm as a string already. If there is no history of $parm usage with a particular shortcode, it is not necessary.

rica-carv:
@Jimmi08 @CaMer0n What about a single line code, using scParams???
`$parms = ((is_array($parm))? eHelper::scParams($parm) : $parm );`

By the way, what's the difference between e107 core function `eHelper::scParams($parm)` and php native function `parse_str` ???

CaMer0n:
scParams should only be used when there is a history of using this format for the shortcode in question:  `{SC=key=val&key1=val1}` 

The format above is deprecated in favor of: `{SC: key=val&key1=val1}`

rica-carv:
@CaMer0n So, in #1554 i used the correct one (`eHelper::scParams($parm)`), since we only have to use `parse_str` on new shortcodes, right?

If i'm allowed, just a suggestion, why not convert all legacy shortcodes with parm to the new style, on parsing???

CaMer0n:
Actually, I believe you introduced a bug in 1554 with  `$parms = eHelper::scParams($parm);`

From application.php

* @param string $parmstr
* @return array associative param array

public static function scParams($parm)

ie. scParams() is for OLD shortcodes that use SC= and which have a STRING value by default. (not an array)

rica-carv:
@CaMer0n Err, somehow, it works for me, without any error..... So you think is better to use `parse_str` ???

CaMer0n:
Please take a look at this: 
```

	public static function scParams($parm)
	{
		if (!$parm) return array();
		if (!is_array($parm))
		{
			$parm = str_replace('\&', '%%__amp__%%', $parm);
			$parm = str_replace('&amp;', '&', $parm); // clean when it comes from the DB
			parse_str($parm, $parm);
			foreach ($parm as $k => $v)
			{
				$parm[str_replace('%%__amp__%%', '&', $k)] = str_replace('%%__amp__%%', '\&', $v);
			}
		}

		return $parm;
	}

```
From your banner commit: 
```
if(is_array($parm))
{
	$parms = eHelper::scParams($parm);
	$campaign = $parms['campaign'];	
}
```

There's no error, but it is unnecessary. Do you see what is wrong with this scenario? You're sending an array to a function that handles strings.

rica-carv:
@CaMer0n You're right, got to revise that... Thanks....

Jimmi08:
@CaMer0n 
I noticed weird thing. With shortcodes {SIGNUP_PASSWORD1} {SIGNUP_PASSWORD2} you can't use parameters. When I remove 1 (number) from shortcode, everything works. I suppose some regex problem with parsing.

CaMer0n:
Okay, let's summarize, because this may be getting confusing.  There are a few scenarios. 

### 1.  Old and no history of $parm usage. 
- Method:   sc_shortcode()  
- Usage:  {SHORTCODE}  
- Parm: null

#### Upgraded
- Method: sc_shortcode($parm=null);  
- Usage: {SHORTCODE: x=y}
- Parm: array
- Processing: none

### 2. Old and $parm is a string
- Method:   sc_shortcode($parm='')  
- Usage:  {SHORTCODE=value}  
- Parm: string

#### Upgraded
- Method: sc_shortcode($parm=null);  
- Usage: {SHORTCODE: x=y} OR {SHORTCODE=value}  
- Parm: array or string
- Processing:  $parm = (is_string($parm)) ? array('x'=>$parm) : $parm; 

### 3. Old and $parm is a string qry. 
- Method:   sc_shortcode($parm='')  
- Usage:  {SHORTCODE=key=value}  
- Parm: string

#### Upgraded
- Method: sc_shortcode($parm=null);  
- Usage: {SHORTCODE: key=value} OR {SHORTCODE=key=value}  
- Parm: array or string
- Processing:  $parm = (is_string($parm)) ? eHelper::scParams($parm) : $parm; 


Hopefully that covers all the scenarios, but if I missed one, please let me know.

CaMer0n:
@Jimmi08 Anything using simpleParse() on the template cannot use $parms. Perhaps that is what is happening in this case.

Jimmi08:
@CaMer0n  I meant this:
https://github.com/Jimmi08/e107_snippets/issues/6

rica-carv:
@cameron What about multiple parms???
**Method 2:** {SHORTCODE: _x=y&w=k_} OR {SHORTCODE=_value&value2_}
    Processing `$parm = (is_string($parm)) ? array('x'=>$parm) : $parm;` will always evaluate as array (right?), but on the second case ({SHORTCODE=_value&value2_}), will return only an array of keys with empty values.

**Method 3:** {SHORTCODE: _key=value&key1=value1_} OR {SHORTCODE=_key=value&key1=value1_}
    Processing: `$parm = (is_string($parm)) ? eHelper::scParams($parm) : $parm;` will always evaluate as array....

Or am i missing something here?

CaMer0n:
@rica-carv I have updated the summary above. I hope it is clearer now.

Moc:
I'm reopening and labeling this as 'documentation' as I think that this (or the result of this discussion) should be documented once settled upon. 

Just FYI : D

CaMer0n:
Thanks @Moc. I don't think it should be in the developer Docs though, because it's really specific to old core in need of upgrading in the core. Better to display only the relevant NEW methods on e107.org. However, this should go somewhere. Maybe on the wiki here?

rica-carv:
@CaMer0n Err, probably `{SHORTCODE=value&value2}` is not used anywhere, i was just showing the possible parms possibilities... :blush:

CaMer0n:
Well, rather than the high possibility of breaking existing shortcodes (inside the core and outside). We deprecated the old way, and use {SHORTCODE: x=y...}  which does exactly that.

rica-carv:
@CaMer0n So really good is to revise all old shortcodes and upgrade them to the new style, correct?

CaMer0n:
Yes, but I suggest we discuss each batch on a case-by-case basis if you wish to do that. There may be special considerations for some shortcodes.

rica-carv:
Ok, i'll keep that in mind.... Thanks....

PR #1557, opened 2016-04-17
Proposed change to newssummary shortcode

rica-carv:
After seeing comment `//TODO Add Pref?`, decided to propose change to shortcode newssummary, and allow use of more parms for it's render.
If extended parms use are allowed inside the shortcode itself, it will allow theme developers to render accordingly to their preferences, and not to depend on the core prefs....

Just my opinion....

### DIFF
--- a/e107_core/shortcodes/batch/news_shortcodes.php
+++ b/e107_core/shortcodes/batch/news_shortcodes.php
@@ -510,23 +510,38 @@ class news_shortcodes extends e_shortcode
 		elseif($this->news_item['news_body']) // Auto-generate from first 2 sentences of news-body. //TODO Add Pref?
 		{
 			$tp = e107::getParser();
-			$text = $tp->toHtml($this->news_item['news_body'],true);
-			$text = str_replace("<br />","\n",$text);
-			$text = strip_tags($text);	
-			$tmp = preg_split('/(\.\s|!|\r|\n|\?)/i', trim($text));	
-			$tmp = array_filter($tmp);
+
+			$ellipses = ($parm['ellipses']?$parm['ellipses']:true);
+			$readmorelink = ($parm['readmorelink']?$parm['readmorelink']:true);
+			$strip_html = ($parm['strip_html']?$parm['strip_html']:true);
+    			$length = ($parm['length']?$parm['length']:500);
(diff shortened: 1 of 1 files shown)


CaMer0n:
Thanks @rica-carv Unfortunately, it's not quite what I had in mind for `{NEWSSUMMARY}` . The "Add Pref" was about having an option in admin or hard-coded for 1,2,3,4 lines of the news-body. 
News-summary by default does not contain HTML.(news_summary from the news table)
The news-body check is a fall-back option when the news_summary value in the DB is empty. (such as coming from v1.x) . So html tags should always be stripped in order to have consistent text-only output from this shortcode be it coming from news_summary or news_body 

The "read more" is another shortcode, so it shouldn't be here. 
The "merge" option can be done by using `{NEWS_BODY}{NEWS_SUMMARY}`

In it's current form I can't merge this code with this branch. 
btw. if there is a specific bug/reason why you chose not to use text_truncate(), please let me know.. there may be room for some improvement in there.

rica-carv:
@CaMer0n I respect your ideias, but if i'm allowed, let me explain mine's:
- strip_tags: i use them because i've got some news texts that came from v1.x, and have html tags in it, so for safe, i strip them (the strip_tags function only deals with html tags, so if a text is clean, nothing will happen...)
- readmorelink: Of course i can use the readmore shortcode, but and if a text is smaller than the cut lenght, and i don't want to display a read more button (or link) after the end of it, as it is by using a shortcode?? With the readmore shortcode present after the news_summary one, the button will always be displayed, even if the string is cut or not.....
- merge: I use it, because if i count the cut lenght of both news_summary (when it exists) + news_body together, and if i count the cut lenght only when i have news_body to cut, both lenght will be equal.... With separated shortcodes you would ever cut only the news_summary, if it's long enough...
A picture for my explanation:
untitled 4

On the current code, the text_truncate function is only called if the limit param exists, so if it's called the simplest [NEWS_SUMMARY} shortcode, it will never be called...
How it renders with the same news text as above:
untitled 5

Besides, the text_truncate function has almost 20 lines of code, while i do the same with two lines of code....

As far i see it, my code ads more flexiblility for display purposes...
But that's just my ideia.......

PR #1561, opened 2016-04-18
Rica carv patch 3 banner shortcode

rica-carv:


rica-carv:
Sorry, duplicated....

PR #1570, opened 2016-04-22
possibility to use lang string in placeholder

Jimmi08:
try to be able to use lang string in placeholder option in text fields


CaMer0n:
Thanks @Jimmi08 What happens if you just use regular text? ie. placeholder=Optional 
Perhaps deftrue() would be better than constant() ?

Jimmi08:
You are right (I needed at first to understand what you meant, but yes, there is better solution, so I am closing this)

PR #1575, opened 2016-04-23
added parameters to some signup shortcodes

Jimmi08:


CaMer0n:
This one parsed the auto-test, but there are a couple of small things to improve. 

`vartrue($parm['class']) ?  $parm['class']  : ''; `
.. and similar.. should be replaced with: 

`vartrue($parm['class'], '') ;`

`shortcode_function($parm) `should have a default value.. 

`shortcode_function($parm=null) `

(otherwise it will throw an error when no $parm is used). 

That's all, the rest looks great.

Jimmi08:
Thanks a lot

PR #1576, opened 2016-04-23
Login table username shortcode

Jimmi08:
This is more for discussion if this is right way. This shortcode is done similarly as signup shortcode and it needs to add signup lang strings. Or I can remove that string, but it's nice title there.  Just first attempt.Thanks


Jimmi08:
Core things should stay simple, now I am able to override this in theme. I am closing this.

PR #1579, opened 2016-04-24
added parameters to some signup shortcodes

Jimmi08:


CaMer0n:
Thank you @Jimmi08 !!

PR #1591, opened 2016-04-28
bug rework, array translation causes to login/user on NAVBAR to disap…

tgtje:
…pear

Behaviour of (arrays) left;right; top and bottom (when LAN's) in use causes the navbar 'area' for login/user to disappear  (admin area for preferences do work fine, but using the LAN,s (field)  as 'command' triggers non functional state.

All other LAN's and functions show no 'bug'


CaMer0n:
@tgtje I believe the answer is this: 

    $frm->select('nav_alignment', array('left'=>LAN_LEFTXXX, 'right'=>LAN_RIGHTXXX), e107::pref('theme', 'nav_alignment', 'left'));

.. please note the removal of "useValues".

tgtje:
I will run some tests with proposal ( if success new commit)

Thanks


---

tgtje:
rework; the array description was already tested> no results at an earlier stage.
The issue arises as ( proposed) useValue is in place..
New commit preparing

PR #1593, opened 2016-04-29
Bug rework navbar > user/login disappearing when language is in use, …

tgtje:
…positioned switching; (value is removed, array is txt defined) No change on language file needed.


CaMer0n:
Thank you @tgtje !!

PR #1595, opened 2016-04-29
extending newstags shortcode

Jimmi08:
I needed this markup for tags:
```
            <div class="blog__footer">
              <ul class="blog__tags">
                <li><a href="#">Bootstrap</a></li>
                <li><a href="#">HTML5</a></li>
                <li><a href="#">CSS3</a></li>
                <li><a href="#">jQuery</a></li>
              </ul>
            </div> <!-- / .blog__footer -->
```
I was able to get it with this code in template and changes in core shortcode:
```
            <div class="blog__footer"> 
              <ul class="blog__tags">
	        <li>{NEWSTAGS: separator=</li> <li>}	</li>
                <li>{NEWSCATEGORY}</li>
              </ul>
            </div> <!-- / .blog__footer -->
```
Result:
image

I didn't find using label param with this shortcode in e107 code.  If there is other way how to do it, let me know.

### DIFF
--- a/e107_core/shortcodes/batch/news_shortcodes.php
+++ b/e107_core/shortcodes/batch/news_shortcodes.php
@@ -948,12 +948,15 @@ class news_shortcodes extends e_shortcode
 		return $info;
 	}
 
-	function sc_newstags($parm='')
+	/* example {NEWSTAGS} */
+	/* example {NEWSTAGS: separator=</li> <li>} */
+	/* example {NEWSTAGS: type=label&separator=</li><li>} */
+	function sc_newstags($parm=null)
 	{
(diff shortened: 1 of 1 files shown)


CaMer0n:
@Jimmi08 Thanks! It looks good, but what about the people who are using `{NEWSTAGS=label} ` ?
You need to add something like: 

    if(is_string($parm) && $parm == 'label')
    {
        $parm = array('type'=>'label');
    }

Jimmi08:
I think that they should correct their code to make things simple but I accept your opinion :)

Jimmi08:
Uff. I take it back. It's reductant now with theme override system and I let this be as it. Ok?

PR #1603, opened 2016-05-02
Patch forum.php for breaked bootsrap css searchbox

rica-carv:
Moved hidden inputs due their present location breaks bootstrap searchbox css format (see picture)
untitled 1


CaMer0n:
Thank you @rica-carv

Issue #1608, opened 2016-05-03
Hardcoded lans in forum plugin files


rica-carv:
[Hardcoded lan in view_forum template `<div class='panel-heading'>Status Keys</div>`](https://github.com/e107inc/e107/blob/master/e107_plugins/forum/templates/forum_viewforum_template.php#L356)
Status keys.....

[Hardcoded lan in forum_viewforum.php `<div class='panel-heading'>Viewable by</div>`](https://github.com/e107inc/e107/blob/209d68a6fe67464446998d99adad476ae8e5ef1b/e107_plugins/forum/forum_viewforum.php#L373)
Viewable by....

Moc:
Feel free to submit pull-requests for this :)

rica-carv:
@Moc I can do it, but where's the guidelines to how to create new LAN strings???

Moc:
http://e107.org/developer-manual/plugin-basics/languages#languages
https://github.com/e107inc/e107/wiki/e107-Coding-Standard#loading-language-files

rica-carv:
@Moc Frankly, i don't know if i should do it, mainly because is not all those hardcoded HTML a candidate for templating????

Moc:
The part in forum_viewforum.php needs to go into a template eventually yes. It isn't an easy task though (see the lines below). Regardless, these hardcoded strings need translation. You can find LAN in templates as well. 

Cameron recently introduced a shortcode that can be used as well: {LAN: LAN_XXX)

tgtje:
Just noticed post; there are more than these 2.. Admin needs some more too, @rica-carv Like to try or on my own todo (rework needed) list ?

rica-carv:
@Moc @tgtje Just proposed the lan change here #1609

Moc:
I think this one can be closed by now, right?

tgtje:
Yes can be closed (looking for items from post 1 : they are present as LAN's ( yesszus committed on GitHub on 15 Sep 2016)

PR #1609, opened 2016-05-04
#1608 Removed harcoded language

rica-carv:
Removed harcoded language.

Will try to see a solution tio the templating issue...


rica-carv:
The templantig issue will not be easy....
Moving the HTML from the function to the template won't do, since if `{VIEWABLE_BY}` is empty, it will display all the divs around it anyway.
That's why i understand that the HTML still is inside the function, since there, if the string is empty, it will return nothing...

But this shortcode could work in template if somehow i could figure out a way to make shortcode wrappers work in forum_viewforum.php file.....

rica-carv:
As suggested here, i've managed to temporarly workaround the templating of the `{VIEWABLE_BY}` shortcode, but since i still got some issues, it's only a temporary workaround....
Please check the changes proposal here (also commented by me):
- [forum_viewforum.php](https://github.com/rica-carv/e107/commit/5dad8e2f94ea3582b60e6daf897e8890cc588613)
- [view_shortcodes.php](https://github.com/rica-carv/e107/commit/7039763bdf89f6e2e7e22047e316ce6c883b5365)
- [forum_viewforum_template.php](https://github.com/rica-carv/e107/commit/94c39dde3c705a58da799a018774c68f190d97c1)

There's also a lot work to be done on the forum_viewforum.php, as it's commented: 
> ///TODO  XXX All these $fVars items need to be put into a shortcode class so they can be parsed with parms and wrappers. Big Job!

CaMer0n:
Unfortunately, this pull-request has conflicts with the existing branch.

rica-carv:
@CaMer0n ??? Anyone doing the same??

CaMer0n:
@rica-carv No, just sync with the latest changes and then resubmit the pull-request. Thanks.

PR #1611, opened 2016-05-04
Download shortcodes update

rica-carv:
Styled edit button to be more site wide consistent look (adden class for buttons), removed forced right alignment (can be donne via template, more user friendly...) and removed hardcoded language string....


CaMer0n:
Thank you @rica-carv

PR #1612, opened 2016-05-04
Language use for plugin Social, and XUP LAN's

tgtje:
Social plugin language use,
Messages for Xup etc

### DIFF
--- a/e107_core/shortcodes/batch/signup_shortcodes.php
+++ b/e107_core/shortcodes/batch/signup_shortcodes.php
@@ -88,8 +88,8 @@ class signup_shortcodes extends e_shortcode
 					
 					// 'signup' Creates a new XUP user if not found, otherwise it logs the person in. 
 					
-					$button = (defset('FONTAWESOME') === 4) ? $tp->toGlyph('fa-'.$ic, array('size'=>$size)) : "<img class='e-tip' title='Register using your {$p} account' src='".e_IMAGE_ABS."xup/{$p}.png' alt='' />";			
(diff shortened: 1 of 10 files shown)


CaMer0n:
Thank you @tgtje !!  - just a small change needed - we should never split a sentence into 2 LANS. eg. LAN_XUP_SIGNUP." {$p} ".LAN_XUP_SIGNUPb. (and never use lowercase. eg. 'b')

Instead we would use one LAN with [x] and then replace [x] with $p. 
eg. "Register using your [x] account".

tgtje:
Sorry ; maybe i do not get all that is needed to accomplish it (as mentioned before, i am not that good with coding). After 3 hours of experimenting (str replace/arrays) i decide NOT to assign that line to a LAN. I just do not get what to do to get it function properly. > alternative

how about this as LAN : 

> define("LAN_XUP_SIGNUP", "Sign-in with your account");
and as code : 

> $text .= " <a title=' {$p} : ". LAN_XUP_SIGNUP." ' role='button' class='signup-xup  btn btn-primary' href='".e107::g....................   (both for button and text)

It will display in the end as user as > ex :    facebook : Sign in with your account.

Moc:
@tgtje Maybe something like this:

From;
`title='Register using your {$p} account'`

To
`title='".str_replace("[x]", $p, LAN_XUP_REG)."'`

With 
`define("LAN_XUP_REG", "Register using your [x] account");`

tgtje:
Thank you , i just saw with this example what i did wrong..

It works doing it this way :-)

---

Moc:
Haha I thought I was quick enough. 

Use the lanVars method. They're essentially the same in this basic scenario but the lanVars method allows for easy customization in the future.

tgtje:
Thank you, will do

tgtje:
new pull created (amend version)

PR #1624, opened 2016-05-06
Language use for plugin Social, and XUP LAN's reworked to use lanVar

tgtje:


CaMer0n:
@tgtje It's much better!! But I'm sorry to say that you have translated some commented-out terms: 
Basic rule: If the line starts with // don't translate it.

tgtje:
Seems amend does work 1 time only ? If not ok, please inform.

CaMer0n:
Great stuff @tgtje ! Just noticed that there's a lot of terms in the English_global.php file. 
Normally it shouldn't contain much more than this.. 

    define("LAN_PLUGIN_SOCIAL_NAME",  "Social");
    define("LAN_PLUGIN_SOCIAL_DIZ" .... 

It would be better to add another 'English_front.php' file and load it when needed. ie. in the _menu.php file etc. ..and place the other terms in that.

tgtje:
will come back to it, had some first result> , but site switch language gets screwed with last changes ( part native, part english)...(have to find cause first)
I close the pull for the moment..

PR #1626, opened 2016-05-07
Language use Plugin Social

tgtje:
rework done, some extra defines in E..global needed! These LAN's (defines) are only needed when social is active .No other file (except main English php) suited . Chose to not clutter main English.

batch signup shortcode > carries signup (xurl)
user handler > carries exception messages  > English/lan user > holds defines
English/lan_email > carries define for (social) mail (email to someone) tested that LAN will display text).

### DIFF
--- a/e107_plugins/social/admin_config.php
+++ b/e107_plugins/social/admin_config.php
@@ -9,7 +9,9 @@ if (!getperms('P'))
 	exit;
 }
 
-
+e107::lan('social',true, true);
+e107::lan('social',false, true);
+e107::lan('social','menu',true); 
 class social_adminarea extends e_admin_dispatcher
 {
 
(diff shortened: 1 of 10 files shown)


CaMer0n:
Great work @tgtje !! I wasn't sure if you were aware, but the only LANs that don't use numbers are the global LANs. everything else should use LAN_PLUGIN_NUMBER or 
LAN_PLUGIN_ADMIN_NUMBER. (or similar) 

So I suggest. 
LAN_SOCIAL_ADMIN_00 // for the admin
and 
LAN_SOCIAL_000 // for the frontend

Since there is no main "social" page.. as all elements are spread around.. so we shouldn't require a separate English_menu language file. I suggest we do something similar to the forum plugin: 

LAN_SOCIAL_000 to LAN_SOCIAL_099 // Common parts of the frontend
LAN_SOCIAL_100 to LAN_SOCIAL_199 // Facebook specific 
LAN_SOCIAL_200 to LAN_SOCIAL_299 // Twitter specific 

Sorry for all the changes, but hopefully this is mostly just a quick search/replace for you. 
Much appreciated!

tgtje:
Ok, will adjust files, i like to close this commit, because i noticed somewhere along the line  the 4 systemfiles are not commited either (the described ones...) ??
I'll make a new pull for social anyway, 
but i have to test first if reassigning the LAN's for Facebook and Twitter (text displayed in menus) can be transferred... ( have in mind that those lan's in E front did not work...have to retest anyway).

If i understand correct you would like to see the same method done FOR forum ( and 'ditch' the E _menu from forum.?? in time.

May i point to this : plugins /blank/e_menu .php although decommented  would point to the use of..
	> {
		// e107::lan('_blank','menu',true); // English_menu.php or {LANGUAGE}_menu.php
	}

tgtje:
New commit

PR #1636, opened 2016-05-10
Make it possible for non admins to download more files with send()

waskosky:
There seemed to be a bug that only admins can download files located in the system import folder. It is possible that this was some kind of security mechanism at one point although calling this function would require PHP access to modify what you are attempting to download anyway, so if there is something sensitive in that files location it would be vulnerable anyway.


waskosky:
Per other communication, the import folder is not intended to be a final resting place for any files or media and the "e107_media" folder is supposed to be used instead.

PR #1664, opened 2016-05-21
Social plugin language use (05-21-16)

tgtje:
reworked, latest files


tgtje:
See code climate finds issues, I do not believe i can add/adjust. a) system files b) supplied file.
These are out of my league.

PR #1687, opened 2016-06-06
Fix for datepicker.

lonalore:
According to this: https://github.com/e107inc/e107/commit/82beb3385c099d1b202674e5daa3475ef2bc9c81#commitcomment-17752290

PR #1694, opened 2016-06-08
Fix for datepicker. Issue #1687

lonalore:


lonalore:
Please test it. I tested with news, and in my custom plugins.

PR #1695, opened 2016-06-09
Fix for datepicker(). Tested!

lonalore:
Issue #1687 
Issue #1694 

I did some tests while drinking my morning coffee because this issue bothered me... here is my test page: http://demo.lonalore.hu/e107v2/e107_plugins/_blank/_blank.php

Before submit:
kepernyokep - 2016-06-09 - 10 13 28

After submit:
kepernyokep - 2016-06-09 - 10 14 05

The point is that, we have to set timestamp (initial value) in the hidden field in the case where timestamp is selected.

### DIFF
--- a/e107_handlers/form_handler.php
+++ b/e107_handlers/form_handler.php
@@ -966,28 +966,26 @@ class e_form
 	 */
 	function datepicker($name, $datestamp = false, $options = null)
 	{
-		
 		if(vartrue($options) && is_string($options))
 		{
-			parse_str($options,$options);	
-		} 
-		
-		$type		= varset($options['type']) ? trim($options['type']) : "date"; // OR  'datetime'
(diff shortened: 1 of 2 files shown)


lonalore:
Furthermore, I recommend moving javascript code into a separated file, and using javascript behaviors, because datepicker is not initialized on form elements are load via Ajax. If you allow me, I will move JS code into a JS behavior.

lonalore:
@CaMer0n can you merge this final fix? Thank you! :)

CaMer0n:
Thanks for the code. Unfortunately it broke some functionality. Try editing a 'string' value directly in the text-input field and hit enter, it should update the hidden value, but it no longer does.

lonalore:
Thank you! This should fix it!

PR #1708, opened 2016-06-16
Fixing an issue

Serios:
Fixing an issue with field type number and passed option min not get applied properly to the field


CaMer0n:
@Serios Thank you. But what if min =0 ?

PR #1729, opened 2016-06-23
$form->button() - Ability to use any kind of button class for the selected action.

lonalore:
The main problem is that I can not create `<button type="button" ... >` with `class="btn-primary"`, because Form API always appends the `btn-default` class to my class attribute. So, `btn-default` class always overrides `btn-primary`.

Reproducation:
```php
$options = array(
	'name'    => 'foo',
	'value'   => 'Action button with class btn-primary',
	'action'  => 'action', // because I want to create <button type="button" ... >
	'label'   => '',
	'options' => array(
		'class' => 'btn-primary',
	),
);

e107::getForm()->button($options['name'], $options['value'], $options['action'], $options['label'], $options['options']);
```

I did some testing with my changes:
kepernyokep_2016-06-23_22-07-53

### DIFF
--- a/e107_handlers/form_handler.php
+++ b/e107_handlers/form_handler.php
@@ -2527,15 +2527,15 @@ class e_form
 		return $text;	
 
 	}
-	
-	
-	
-	
-	
-	
-	
+
+
+
+
+
+
+
 	/**
-	 * Admin Button - for front-end, use button(); 
+	 * Admin Button - for front-end, use button();
 	 * @param string $name
 	 * @param string $value
 	 * @param string $action [optional] default is submit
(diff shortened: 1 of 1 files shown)


rica-carv:
Weird. Currently, on the frontend, i can render a button with the style i want, without any hassle...
This issue happens where?

lonalore:
Yes, you can render button with custom style, but if you want to render something like this:

`<button type="button" name="" value="" class="btn btn-primary">...</button>`

The result will be:
`<button type="button" name="" value="" class="btn btn-primary btn-default">...</button>`

And the button will appear on the site with 'btn-default' styles, and not with 'btn-primary'.

lonalore:
Of course, you can set the proper '[action](https://github.com/e107inc/e107/blob/master/e107_handlers/form_handler.php#L2568-L2622)' to achieve the desired CSS class, but in this case, the button always has 'type="submit"' [because of this](https://github.com/e107inc/e107/blob/master/e107_handlers/form_handler.php#L2549). But I want to use button with 'type="button"'. :)

rica-carv:
Sorry, my fault....
My rendered buttons are link tags....: `<a class="btn btn-primary btn-large " href="/news.php">News</a> or <a class="btn btn-default btn-large " href="/contact.php">Contact</a>`

CaMer0n:
@lonalore Thank you!

PR #1735, opened 2016-06-25
Some bug fixes

lonalore:


lonalore:
@CaMer0n Could you please merge these changes? If you have time. Thank you!

lonalore:
My last commit is a fix for XmlDependencies method in order to check dependencies properly. The main problem was that, `$dv `was an indexed array in `foreach ($tag as $dt => $dv)`

I `var_damp()`-ed `$dv` variable inside foreach and I got this:
```php
array(1) {
	[0]=> array(2) {
		["@attributes"]=> array(2) {
			["name"]=> string(6) "nodejs"
			["min_version"]=> string(3) "1.0"
		}
		["@value"]=> string(0) ""
	}
}
```

So I put another foreach into the main foreach in order to check (all) dependencies properly.
```php
foreach ($tag as $dt => $da) // $da is an indexed array
{
	foreach ($da as $dv) {
		if (isset($dv['@attributes']) && isset($dv['@attributes']['name'])) {
			...
		}
	}
}
```

Now, dependency-checking works properly:
kepernyokep - 2016-06-29 - 12 11 24

dobem:
This will not work because $pref is not loaded at XmlDependencies($tag) method.

Also, it's better to make sure to avoid the following situation:
1. plugin b requires a;
2. install plugin a and b;
3. uninstall plugin a;

At least a confirmation message that plugin b might stop working while uninstalling plugin a.

Moc:
The last time I looked at the dependencies issue, it was solely the looping through which was the issue and which has now been corrected by lonalore. As far as I remember, I don't recall there being an issue with the prefs not loaded. The screenshot above by lonalore shows that his fix is working. Anyhow, before it's merged in it will be tested anway.

lonalore:
@dobem thanks for your review! Indeed, `$pref` variable was undefined. 
Furthermore, I added some code to check that the plugin is used by another plugin or not... before uninstalling it.

Moc:
@dobem: good catch!

rica-carv:
Sorry to bring this back up, but just a few questions:

1. If plugin B only depends of plugin A for, let's say, informational purposes (ie, doesn't need to get code from plugin A, just needs data), 
2. If plugin B depends of plugin A, but it's not critical for plugin B to work (ie, plugin B has a fallback so if plugin A is not present, it won't show anything related to plugin A), 

should plugin A be on plugin B dependencies on any of the above situations?
I think not, but would like to ear some other opinions....

PR #1756, opened 2016-07-01
Fixed Form API issues on Admin UI. New feature (CDN variants) for Library Manager.

lonalore:


CaMer0n:
Thanks! This is a lot for one commit. I believe it should work in reverse. When the theme defined BOOTSTRAP, the core loads the library file. Requiring every bootstrap theme to include a library file is counter-productive.

PR #1802, opened 2016-08-15
Update comment_class.php

mostvotedplaya:
Fix XSS issues.

You can verify this by entering

```
<script>alert(1);</script>
```

in a comment on news, you should see the alert box rendered to screen.


CaMer0n:
I don't think this addition would hurt. Could be helpful in the case of an admin posting script.

PR #1816, opened 2016-08-27
#6 Info Panel + User Admin Text to LANS

MikeyGMT:
#6 Info Panel Text to LANS

### DIFF
--- a/e107_admin/includes/infopanel.php
+++ b/e107_admin/includes/infopanel.php
@@ -209,11 +209,11 @@ class adminstyle_infopanel
 	//  ------------------------------- e107 News --------------------------------
 
 		$newsTabs = array();
-		$newsTabs['coreFeed'] = array('caption'=>'General','text'=>"<div id='e-adminfeed' style='min-height:300px'></div><div class='right'><a rel='external' href='".ADMINFEEDMORE."'>".LAN_MORE."</a></div>");
-		$newsTabs['pluginFeed'] = array('caption'=>'Plugins','text'=>"<div id='e-adminfeed-plugin'></div>");
-		$newsTabs['themeFeed'] = array('caption'=>'Themes','text'=>"<div id='e-adminfeed-theme'></div>");
+		$newsTabs['coreFeed'] = array('caption'=>LAN_GENERAL,'text'=>"<div id='e-adminfeed' style='min-height:300px'></div><div class='right'><a rel='external' href='".ADMINFEEDMORE."'>".LAN_MORE."</a></div>");
+		$newsTabs['pluginFeed'] = array('caption'=>LAN_PLUGIN,'text'=>"<div id='e-adminfeed-plugin'></div>");
+		$newsTabs['themeFeed'] = array('caption'=>LAN_THEMES,'text'=>"<div id='e-adminfeed-theme'></div>");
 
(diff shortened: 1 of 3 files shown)


MikeyGMT:
Testing 

image

image

MikeyGMT:
Sorry this got added to the existing PULL - thought it would create a new!

Addition of User TODO Lans 

Extended User Test 
image

Moc:
To create a new pull request, you'll need to create a new branch on your fork and then submit a pull request (for that new branch).

MikeyGMT:
Thanks @Moc, I think I'll just wait until @CaMer0n reviews and merges. 
I have done another 10 files cleaning up lans today.

CaMer0n:
Thanks for this Mikey. I believe for the extended field messages we need something to indicate it is for user_extended files and not just regular user fields. (helps in hunting down future issues too). Perhaps placing "(Extended User Fields)" at the end of the message. ? ie. LAN_UPDATED." ( ".ADLAN_78.") ";

MikeyGMT:
@CaMer0n but then it's not generic lan? Confused, because the user will be on the Extended User Fields page, no?

CaMer0n:
@MikeyGMT You're editing 107_admin/users.php (admin editing of 2 linked database tables e107_user and e107_user_extended).
I'm just suggesting to combine 2 generic terms to add clarity, especially in the case of errors.

MikeyGMT:
Oh ok ... you mean something like this 

> e107::getMessage()->addSuccess(LAN_UPDATED.': '.EXTLAN_34);

MikeyGMT:
I have made some improvement

image
 
But I cannot see where the success message is generated from for the Users Table.

MikeyGMT:
@CaMer0n added the entity at the end of the the failure messages and success for UE table - but I still cannot locate where the success message for Users table is generated. So it looks a little odd - as per the image in the post above.

CaMer0n:
@MikeyGMT the main user table part is automatic and part of the admin-ui. I suggest we remove the 'success' for the user-extended, but keep it if there is an error. (in which case it will be rendered in it's own red box)

tgtje:
Just searched before last post was made :  (work done , so anyway)
here was the info (_maybe helpfull or not; in regards latest post_..)
LAN_SUCCESS is present in ..admin/db.php but has NO reference in lang defines (...export xmlfile)

LAN_UPDATED =  Successfully updated = bottom line
and
update_routines.php  (e107::getMessage()->addSuccess(LAN_UPDATED." : ".$path);)  = line 234 (function updatePlugin) = top line

rica-carv:
Maybe i'm wrong, but as far i can see, [LAN_SUCCESS is legacy](https://github.com/e107inc/e107/blob/60f1d78dce59b5937ddfd85146d3fcae95d2c688/e107_admin/db.php#L74)...

Now it's [E_MESSAGE_SUCCESS, right](https://github.com/e107inc/e107/blob/4a40104073c2a9cfda91a2d1d000d56e01f4dab6/e107_handlers/message_handler.php#L27)?

MikeyGMT:
Thanks folks, I'll rework later this week.

MikeyGMT:
Couldn't sleep, so here it is. 
image

Have replaced "success" messages with debugs instead, which may be helpful for admins.

@rica-carv LAN_SUCCESS is current, LANs are for translation into localisations. 
E_MESSAGE_SUCCESS is, I think, for use in the programming, and not rendered on the front end, as LANS are... correct me if wrong.

PR #1825, opened 2016-08-31
#6 User settings Text To Lans

MikeyGMT:
#6 User settings Text To Lans


MikeyGMT:
Testing "Class Structure"
image

I don't use user ratings anywhere.

MikeyGMT:
Just found this in user.php LAN_112 needs to be re-enabled via the deprecated lans tool. 

>// BC for v1.x template
>$bcList = array(
>	'LAN_112'   => 'LAN_USER_60',// Email Address

So I need to do a bit more work on in this pull.

MikeyGMT:
Checked the LAN_112 but don't think it is an issue. 
I replaced LAN_USER_60 on my user template with LAN_112 
Email address is displayed ok

MikeyGMT:
image

CaMer0n:
Thanks @MikeyGMT It looks good, other than it broke e107 standards. Only lan_admin.php and English.php (and English_global.php)l should contain the "LAN_ENGLISH_TERM" format. Everything else should use LAN_ABBREV_XX where XX is numeric.

MikeyGMT:
No problem, I hadn't known that. I thought we had agreed in prior conversations that we were moving away from the numbered LANs, as they seem to get reused which causes confusion, plus they're easier to read in the code itself. If lan_install_99 is disused or redefined with a new English term then it's not easy to spot if the foreign version needs revising. Happy to rework as needed.

PR #1830, opened 2016-09-03
Sept16

tgtje:
Text to LAN's

since system core files change rapidly here are a few txt to lan's

### DIFF
--- a/e107_admin/admin.php
+++ b/e107_admin/admin.php
@@ -148,7 +148,7 @@ class admin_start
 		
 		if(deftrue('e_MEDIA') && is_dir(e_MEDIA) && !is_writable(e_MEDIA))
 		{
-			$mes->addWarning("The folder ".e_MEDIA." is not writable. Please correct before proceeding.");			
+			$mes->addWarning(LAN_CHMOD_MEDIA);			
 		}	
 		
 		if(deftrue('e_SYSTEM') && is_dir(e_SYSTEM) && !is_writable(e_SYSTEM))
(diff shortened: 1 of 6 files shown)


CaMer0n:
Thank you! It's almost usable as is. However some LANs contain paths (eg. e_MEDIA) and {SHORTCODES}. These need to be replaced with [x] and then $tp->lanVars() used to replace them within the context of the script using them.

Moc:
@tgtje Something like this is what is needed: 

```php
define("LAN_CHMOD_FOLDER", "The folder [x] is not writable. Please correct before proceeding.");
```

And then:
```php
$message = $tp->lanVars(LAN_CHMOD_FOLDER, e_MEDIA); 
$mes->addWarning($message);
```


@CaMer0n  Are you referring to CUSLAN_80 with a shortcode? If yes, would this be solution? 

```php
define("CUSLAN_80", "Will be listed in the Menu-Manager under this name or may be called using [x] in your theme. Must use ASCII characters only and be all lowercase.");
```  	

```php
		'menu_name' 		=> array('title'=> CUSLAN_64, 		'tab' => 2,	'type' => 'text', 		'width' => 'auto','nolist'=>true, "help"=> $tp->lanVars(CUSLAN_80, "{CMENU=name}"),

tgtje:
Shortcode ? 
CUSLAN_80
Will be listed in the Menu-Manager under this name or may be called using {CMENU=name} in your theme   (CMENU=  is pure textual (no linkage) = Create Page/MEnu ..TAB MENU > Menu Name > hover text....

CaMer0n:
@Moc Thanks. Yes, that looks right. (as long as it is defined in the init() method)

PR #1838, opened 2016-09-13
New pharses

yesszus:
Add view forum pharses


CaMer0n:
Thank you @yesszus ! This one is good! It helps me if the LAN definitions and where they are inserted/used are in the same pull-request.

PR #1842, opened 2016-09-13
Update

yesszus:
Add new pharses


CaMer0n:
@yesszus Thank you! All good. :+1:

PR #1843, opened 2016-09-13
Update post_shortcodes.php

yesszus:
Add pharse


CaMer0n:
@yesszus Thank you!! btw.  There's no point using this: `return "".LAN_FORUM_8013.""; ` it's better to just use this: `return LAN_FORUM_8013; `  (otherwise it's  a bit like saying: return 0 + 1 + 0; )

PR #1844, opened 2016-09-14
Update  e107_plugins/pm/admin_config.php

yesszus:
Add pharses.


CaMer0n:
Thank you @yesszus ! I believe we need to check which of these terms are also used in the PM menu on the frontend and the PM page on the frontend.. and then replace accordingly to avoid duplicates.

yesszus:
LAN_PM_1 not work in admin menu send new private message.
Not conflict PM menu on the frontend and PM page.

[http://hun-social.dtsh.hu/pm.png](url)

PR #1845, opened 2016-09-14
Update  e107_plugins/pm/languages/English_global.php

yesszus:
Add new pharses.


CaMer0n:
I believe we have some duplicates in there which need to be carefully looked at. eg. define("LAN_PM_35", "Send new message");  (English.php).

yesszus:
All codes restructured of all files, remove duplicates.

PR #1848, opened 2016-09-14
Update e107_languages/english/admin/lan_db.php

yesszus:
Add new pharses


CaMer0n:
@yesszus Excellent! Thank you.

PR #1851, opened 2016-09-14
Update forum_viewforum.php

yesszus:
Add lang codes


CaMer0n:
Thank you @yesszus !! It's close to prefect. There's just a couple of small things to fix. 

     Delete ".$tp->toGlyph('trash');  // note the empty space after 'Delete'. 

should become: 

    ".LAN_DELETE." ".$tp->toGlyph('trash'); 

The same applies to Line 999, but also you're also missing the close '>' on the '<a' tag. 

With this in mind, please be sure to test changes before submitting. 
Thank you again, once corrected these pull-requests are a huge help.

PR #1857, opened 2016-09-15
Update forum_class.php

yesszus:
Add lang codes


CaMer0n:
Nice work. Thank you! :+1:

PR #1869, opened 2016-09-18
Some corrections

yesszus:
Add lang codes


CaMer0n:
@yesszus Thank you. Did you test your changes? I believe there is a syntax error on line 350.

PR #1882, opened 2016-09-22
Update e_marketplace.php

yesszus:
Add new pharses


e107_handlers/e_marketplace.php
```diff
@@ -347,7 +347,7 @@ public function download($id, $mode, $type)
 		
 		if(!file_exists(e_TEMP.$localfile))
 		{
-			$mes->addError( "Automated download not possible. Please <a href='".$remotefile."'>Download Manually</a>"); 
+			$mes->addError( "".EPL_ADLAN_235." <a href='".$remotefile."'>".EPL_ADLAN_236."</a>");
```
CaMer0n: Thank you for this pull-request. 
`"". ` should be removed before EPL_ADLAN_235 - it is useless code.

yesszus: Fixed.

Issue #1883, opened 2016-09-23
wrong action in form with userlanguage menu


Jimmi08:
I noticed that with SEF-URL (via e_url.php) this code:

	`$action = (e_QUERY && ! $_GET['elan']) ? e_SELF.'?'.e_QUERY : e_SELF;`

return wrong path (adding e_QUERY to URL, path URL is already correct thanks SEF-URL)

Changing this to 
`$action = e_REQUEST_SELF;` 
helped

I am not sure it's correct or problem was somewhere else.

CaMer0n:
@Jimmi08 That looks like the correct fix.

Moc:
Fixed by pull-request #1908

PR #1884, opened 2016-09-24
wrong action in form with userlanguage menu #1883

Jimmi08:


CaMer0n:
@Jimmi08 This is good, but now it removed any queries that were in the URL. Normally just replacing e_SELF with e_REQUEST_SELF should work?

Jimmi08:
It doesn't work.  I did this:
`$action = (e_QUERY && ! $_GET['elan']) ? e_REQUEST_SELF.'?'.e_QUERY : e_REQUEST_SELF;`

original url:
`http://v2demo.e107.sk/news/month/201602`
Then changed lang via menu:
result:
`http://v2demo.e107.sk/news/month/201602/?month.201602.0`

CaMer0n:
@Jimmi08 Then it is probably better to use e_REQUEST_URI for backward compatibility.

PR #1888, opened 2016-09-26
Update forum_admin.php

yesszus:
Add lang codes.


CaMer0n:
Thank you!! e_PAGETITLE usage needs to be replaced with LAN_PLUGIN_FORUM_TITLE (if I remember well, from English_global.php )

PR #1893, opened 2016-09-29
Update e107_languages/English/admin/lan_plugin.php

yesszus:
Add new pharses.


CaMer0n:
@MikeyGMT this may require some of your LAN optimization if you're up for it.

MikeyGMT:
@CaMer0n is it possible to assign these?

PR #1901, opened 2016-09-30
Update e107_admin/image.php

yesszus:
Add lang codes.


tgtje:
duplicate definition (already available) define("IMALAN_21", "Used by");
 In commit used as IMALAN_131 >media_usedby

PR #1905, opened 2016-10-01
Update e107_admin/image.php

yesszus:
Add lang codes.

### DIFF
--- a/e107_admin/image.php
+++ b/e107_admin/image.php
@@ -195,7 +195,7 @@ class media_cat_ui extends e_admin_ui
 		);
 		
 		// FIXME LAN
-		$this->fields['media_cat_type']['writeParms'] = array('image' => IMALAN_143, IMALAN_144 => 'File', 'video' => 'Video');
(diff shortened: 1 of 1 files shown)


e107_admin/image.php
```diff
@@ -195,7 +195,7 @@ function init()
 		);
 		
 		// FIXME LAN
-		$this->fields['media_cat_type']['writeParms'] = array('image' => 'Image', 'file' => 'File', 'video' => 'Video');
+		$this->fields['media_cat_type']['writeParms'] = array('image' => IMALAN_143, IMALAN_144 => 'File', 'video' => 'Video');
```
CaMer0n: changing a code value with a LAN will cause problems. For arrays in this "x => y" format, only "y" should ever be replaced with a LAN. 
`$this->fields['media_cat_type']['writeParms'] = array('image' =>IMALAN_143, 'file' => IMALAN_144, 'video' => OTHER_LAN_DEFINITION);`

Issue #1819, opened 2016-08-28
File Inspector > Setup > would be better named Scan


MikeyGMT:
Hi All, 
Opening the File Inspector, the left menu options are 
"Setup" 
"Results"

Setup is not really correct, it would be better to rename this to 
"Scan"

Does anyone disagree ?
Thanks
Mikey

CaMer0n:
@MikeyGMT Okay to change it ! :+1:  Just make sure if you edit the LAN that it is not used anywhere else (as "Setup").

MikeyGMT:
Slight mod proposed... Scan Now moved to the left menu, Scan Now was on the (renamed) Go button.

![image](https://cloud.githubusercontent.com/assets/2847651/18147132/29adb776-6fcb-11e6-81ff-9a69b563c856.png)

About On and Off, in prefs we use yes/no - shall we use yes/no instead?

MikeyGMT:
>	function create_image($dir) {

I have not changed this section, no idea if it is used, but we should update the copyright data. Any thoughts? 

>	function snapshot_interface() 

Lots of TODO-LANS ... again, no idea if used / what it really does. 

>	function parsesize($size, $dec = 0) {

https://en.wikipedia.org/wiki/Megabyte
File Size in file inspector is lower case for KB, MB, GB etc., SI unit nomenclature should be corrected e.g. kB, MB, GB, TB etc. I does not need to be translated, being an international standard. 

Comments?

MikeyGMT:
> function fileinspector_adminmenu() //FIXME - has problems when navigation is on the LEFT instead of the right. 
 
Unclear, is this still relevant seems ok to me?

MikeyGMT:
define("FC_LAN_5", "Core Files");

define("FR_LAN_4", "Core files");

Why?

MikeyGMT:
![image](https://cloud.githubusercontent.com/assets/2847651/18148657/ec45e55e-6fd2-11e6-9245-779a6b60db32.png)

How's that?

MikeyGMT:
TODO LANS on snapshot_interface Done, but it isn't working - on unmodified site. There's a number of duplicate defines. Commenting those.

MikeyGMT:
Redone the LANS work in File inspector. 

In Dev Mode - I played with the PREG String match function. Seems to do something odd, depending on the string - examples from the php page link in the help text. 

Sometimes never completes 

[...]

PR #1914, opened 2016-10-10
Issue #1819 Clean up for Lans on File Inspector

MikeyGMT:
Issue #1819 Clean up for Lans on File Inspector


CaMer0n:
Nice work @MikeyGMT !  Cheers.

PR #1924, opened 2016-10-15
Update e107_languages/English/admin/lan_image.php

yesszus:
Add new pharses


CaMer0n:
Thank you. There's HTML in the LANs which need removal. Any links should be hardcoded in the script and can be inserted using an [x] in the LAN or wrapping the linkable text with [ ] and using str_replace().

yesszus:
Removed HTML codes:

5f2f42a

yesszus:
Updated e107_languages/English/admin/lan_image.php:
a1722b3 

Recoded e107_admin/image.php:
6e2f6d1

PR #1927, opened 2016-10-16
Update lan_submitnews.php

yesszus:
Add pharses.


MikeyGMT:
Better to use LAN_SUMMARY ... etc from English.php - then we only need translate one phrase :)

CaMer0n:
@MikeyGMT You may wish to review these changes after merging.

PR #1931, opened 2016-10-16
Update e107_languages/English/lan_notify.php

yesszus:
Add pharses


CaMer0n:
Thanks. This format: "NT_LAN_NOTIFY_DISABLED" which contains no numeric values should only ever be used in lan_admin.php, e107_languages/English.php and English_global.php (in plugins) . All other LANS should end in numeric values.

Issue #1887, opened 2016-09-25
Minor bug: Updating User Settings having no Userclass set generates a warning in_array 


MikeyGMT:
When in the front end, go to user settings, a warning happens, if the user has no userclasses set.  

Warning in_array() expects parameter 2 to be array, string given, Line 738 of /home/nplsc2k1/public_html/e107_handlers/user_handler.php

I have tried this for a solution and it seems to work

```php
		if ($incInherited)
		{
			$classList = e107::getUserClass()->get_all_user_classes($userData['user_class']);
		}
		else
		{
			if ($userData['user_class'] != '') 
			{
				$classList = explode(',',$userData['user_class']);	
			}
			else 
			{
			$classList = array ('0' => '');
			}
		}

```

MikeyGMT:
Should the userclass not already be populated? 

Also notice that Main Admin is not checked.

![image](https://cloud.githubusercontent.com/assets/2847651/18818576/d485223a-8374-11e6-876f-ab757ad2c553.png)

Moc:
Can you share your PHP version?

MikeyGMT:
PHP Version 5.3.27

CaMer0n:
@MikeyGMT Would you mind making a pull-request? btw. please update to `!empty($userData['user_class'])` from `$userData['user_class'] != ''`

MikeyGMT:
Main Admin remains unchecked - commit does not fix this.

PR #1935, opened 2016-10-17
#1887 User Setting Fix for Users with No Userclass plus some Notice Removals

MikeyGMT:
#1887 User Setting Fix for Users with No Userclass plus some Notice Removals
Tested before submission


CaMer0n:
Thank you @MikeyGMT !! :+1:

PR #1941, opened 2016-10-18
Update  e107_admin/upload.php 

yesszus:
Add lang codes.


CaMer0n:
Thank you @yesszus . Are you able to include 2 files in a single pull-request? eg. e107_admin/upload.php and lan_upload.php ?

Moc:
You can do this by pushing both commits to the same branch (e.g. also commit the upload.php file to patch-1 instead of comitting to a new patch-2 branch).

Issue #1944, opened 2016-10-19
[bug] e107::getComment()->render() not ready to display mutiple times on same page


nesjett:
As the tittle says, It seems that if you call reder() multiple times on the same page, It´s default behaviour will be broken.

It adds the comment as espected, and renders them, but the javascript is not done in the right way and it shows the resulting new comment on the wrong element.

Explanation:
The problem is that you handle a lot of things with ID in order of using Classes and find them trought parent (or somthing dynamic)

Scenario:
We would need a lot of comment forms on a social news feed, for example like facebook one.

I´ve already solved that, but I´m still working on making prev next buttons work, and edit one

Ill make a pull request when finished, hope it helps.

nesjett:
Update: Im driving crazy with next prev buttons, can someone confirm that they are working as expected? (in my case they just doesnt do nothing)

Moc:
@nesjett Have you been able to solve this?

CaMer0n:
@Moc @nesjett @Jimmi08 @SimSync  This commit above is a partial fix that hopefully won't break the comments. If you guys can test my changes thoroughly over the next 48 hours, that would be very helpful. I have tested locally, and haven't found any issue so far.

Moc:
So far, adding comments still seem to work. Haven't tested with multiple comment forms yet.

Evgura:
@CaMer0n The whole structure of the comments was broken. Maybe because of CSS or Javascript.
Rolled back the update.

Jimmi08:
Hi all, 
I don't know what to test because I hadn't any problem with comments (except the position of new reply comment with nested comments)

_before update:_
So I added nested comments to news, no problems, I changed setting to set nested off, I got paginated comments, button prev and next works.

_I updated_
Everything looks normal, both nested and not nested. 
I only noticed a bit slow reaction on next,prev buttons but it can my installation itself. 

_I added comment menu to news sidebar..._
everything works

It surprised me that Moderate comments took me to admin area...

Ok, I maybe have the problem:

-  single news
- click on reply -  works, the position of reply is fixed now
- click on reply on another comment - no reaction
after refresh, I can reply, but it takes visible time to display textarea (again it could be my problem) 

- I can open more reply textarea at once after refresh... just note, maybe it's ok. 

-  I have set Comment sorting - newest at first, but if I am on next page (not first), and a new comment is posted (not reply, first level comment), it's not visible, you need to manually click on pagination...

CaMer0n:
[...]

PR #1945, opened 2016-10-19
Fix for displaying multiple comment forms on the same page

nesjett:
Because the use of IDs in order of Classes as identifiers, the comment
form was not able to work properly.
NOTE: Next and Prev buttons doesnt work, but I think its not related to
this issue. Please, confirm that they are not working on the e107
previous this commit.

### DIFF
--- a/e107_web/js/core/front.jquery.js
+++ b/e107_web/js/core/front.jquery.js
@@ -80,12 +80,15 @@ $(document).ready(function()
 
     $(document).on("click", ".e-comment-submit", function(){
 			
+			var commentsParent = $(this).parent().parent().parent().parent().parent().parent().parent();
 			var url		= $(this).attr("data-target");
 			var sort	= $(this).attr("data-sort");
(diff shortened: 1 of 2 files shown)


nesjett:
Fix related to #1944

e107_web/js/core/front.jquery.js
```diff
@@ -80,12 +80,15 @@ $(document).ready(function()
 
     $(document).on("click", ".e-comment-submit", function(){
 			
+			var commentsParent = $(this).parent().parent().parent().parent().parent().parent().parent();
```
lonalore: This won't work if you use custom templates for comments, because the HTML structure may be not the same. Use parents() instead. For example if you want to get the form, which contains your element, use something like this:

```javascript
var $form = $(this).parents('form:first');
```

CaMer0n:
Closing this as the proposed changes would break as per the comments mentioned above.

Issue #1936, opened 2016-10-17
Core update page displays incorrect status after update


rica-carv:
Live site is currently a v1.4 upgraded to v2.x, but for some reason database update is stuck in core database structure, and github update gives a lot errors.....

![untitled 2](https://cloud.githubusercontent.com/assets/17130530/19448099/416fad00-9498-11e6-98fd-1a207326a8b2.jpg)

![untitled 1](https://cloud.githubusercontent.com/assets/17130530/19448125/5e85e7d8-9498-11e6-950a-6e6bd5cfe6c9.jpg)

Can anyone help to clean up this mess? Thanks....

tgtje:
Seen it happen before, doing the sync again could help. (transfer from temp to dir looks failing sometimes > server glitches ???) (use ftp else).

All other related posts will suffer from this cause !! Visited selos, > example : forum has issues too... paths reveal old v1 forum files etc...; java errors  var showhide=new switchcontent > log (show/hide) etc.. Heavily adjusted theme construct, and some special plugins... hard to tell what all might have influence..
Do not forget (afaik) .tk is domain, but source is normally elsewere coupled.

rica-carv:
@tgtje Donne new sync, same error. So i'm going with heavy artillery via FTP, as last resort.

Meanwhile i'm trying to make a V2.x of it without upgrading from V1.4, so therefore all those errors will be gonne (i hope...)
The theme is currently being (slowly) upgraded to v2.x....
The special plugins are also on the "to upgrade to v2.x" list, so....But as v1.4 they work on v2.x, but who knows when they'll start to fail....

Moc:
Try to do a manual update. 

- Download the ZIP from Github, then upload and overwrite all files. 
- Run the database updates (in debug mode preferably, so we can see where it gets stuck if it does).

rica-carv:
@Moc Manual update works, site is now updated to 2.1.2 (git), but github update still gives the same errors....

CaMer0n:
These errors can often be ignored.  Consider it debug info. If files are unable to be unzipped then it's a server configuration issue.

rica-carv:
@CaMer0n Did a database check, had some errors had to be corrected directly via Phpmyadmin, now it's ok, but v2 forum (database sucessfully upgraded) and links plugins gives me errors now....

Moc:
Which errors?

rica-carv:
@moc Only have this:
[...]

PR #1966, opened 2016-10-28
Added CURLOPT_TIMEOUT to Fix GitSync #1936 and some fixme flags

MikeyGMT:
Added CURLOPT_TIMEOUT to Fix GitSync #1936 and some fixme flags - for file class and for BBCODE handler notice


CaMer0n:
Thank you @MikeyGMT !

PR #1986, opened 2016-11-02
Update e107_admin/image.php

yesszus:
Some corrections and add lang codes.


yesszus:
Updated e107_languages/English/admin/lan_image.php

669b8ae

PR #2005, opened 2016-11-06
Update e107_languages/English/admin/lan_admin.php 

yesszus:
Some corrections.


yesszus:
Renamed.

CaMer0n:
Thank you @yesszus !! Just a little change, let's use LAN_UI_FILTER for this particular set that applies to the filter dropdown. (Just like we do for LAN_UI_BATCH )

yesszus:
Changed.

CaMer0n:
Thank you! I will make some changes.

Issue #2020, opened 2016-11-10
Custom theme forumdescription shortcode not working...


rica-carv:
I wonder if anyone could give it a try on this.
I tried to customize the `forumdescription `shortcode on my `theme_shortcodes.php`, where i do already have some working news shortcodes, and this simple fuction doesn't work:
```
	function sc_forumdescription()
	{
    return "test test test";
    }
```

I simply got a blank output where the shortcode should output `test test test` ....

Further testing with any other overriden shortcode, when in the `forum.php` page, the overrides won't work, giving a blank output instead of the shortcode output.
But when in the` forum_viewforum.php` page, the overriden shortcodes works as desired....
I've just tested it with the overriden `repliesx` shortcode on both pages...

PR #2021, opened 2016-11-10
Patch forum.php for override shortcodes

rica-carv:
Patch forum.php for override shortcodes, issue #2020


CaMer0n:
Great! Thank you !

PR #2022, opened 2016-11-10
Allow forum_description to have core shortcodes and parse them...

rica-carv:
Allow forum_description to have core shortcodes and parse them...


CaMer0n:
I'm not sure about this one. Normally the shortcode should go into the template, not into the DB description.

rica-carv:
@CaMer0n So how can i deal with this on the forum description field:
`Área para a colocação de questões sobre a utilização do {SITENAME}`

I'm using the core shortcode `{SITENAME}` due to if i change the site name, it will be automatically updated on the forum descriptions using that shortcode.
I've got a lot of forum descriptions using the shortcode, so changing all them when i eventually change the site name (and yes, i've already change it some 20 times... :cry: ), is a litlle bit overworking...

Just my thoughts....

CaMer0n:
You could use a custom shortcode in your theme and add the sitename, or you could remove the shortcode completely, place your text with {SITENAME} in it.

rica-carv:
@CaMer0n If i added a shortcode to my theme and add the sitename, i still would need the description to render my custom theme shortcode indide the description properly, no?
I can understand the unflexibility to not use shortcodes outside templates, but if that's so, why not make them invisible if they're present in descriptions, etc, as i have?
Instead of displayng "`Área para a colocação de questões sobre a utilização do {SITENAME}`", the output could be simply "`Área para a colocação de questões sobre a utilização do (SHORTCODE NOT ALLOWED)`" or "`Área para a colocação de questões sobre a utilização do            `"...

Anyway, a workaround this personal issue is if i could somehow do a custom bbcode, kind of [SITENAME]. But how can i add custom bbcodes sitewide, not only for specific theme?

PR #2024, opened 2016-11-10
Viewable_by shortcode wrapper enabled

rica-carv:
Viewable_by shortcode wrapper enabled and hardcoded html removed...


CaMer0n:
Thanks @rica-carv ! :+1:

Issue #1616, opened 2016-05-05
Question: Can Shortcode wrapper handle Shortcode code inside itself?


rica-carv:
I'm wondering if its possible to have a shortcode wrapper with shortcodes inside it, and have it parsed correctly, ie like this:
`$DOWNLOAD_WRAPPER['view']['DOWNLOAD_VIEW_DATE'] = '<span class="e-tip" title="{DOWNLOAD_VIEW_DATE_LAN}: {DOWNLOAD_VIEW_DATE=long}">{GLYPH=date}&nbsp;{---}</span>`

My main issue is that i don't want to render the tootip or the glyph icon associated to the shortcode,if the shortcode is blank....

Any ideias?

CaMer0n:
No, it's not possible. Just use regular html markup to render the glyph.

rica-carv:
@CaMer0n The main issue is not the glyph, is the title render: `title="{DOWNLOAD_VIEW_DATE_LAN}: {DOWNLOAD_VIEW_DATE=long}"`

Moc:
Is this issue still relevant?

rica-carv:
Solved in #2029

PR #2029, opened 2016-11-11
Enable wrapper to have parseable shortcodes

rica-carv:
Enable wrapper to have parseable shortcodes, issue #1616

### DIFF
--- a/e107_handlers/shortcode_handler.php
+++ b/e107_handlers/shortcode_handler.php
@@ -1376,48 +1376,47 @@ class e_parse_shortcode
 		if(isset($this->wrappers[$code]) && !empty($this->wrappers[$code])) // eg: $NEWS_WRAPPER['view']['item']['NEWSIMAGE']
 		{
 			list($pre, $post) = explode("{---}", $this->wrappers[$code], 2);
-			return $pre.$ret.$post;
+//			return $pre.$ret.$post;
 		}
 
-		if(!empty($fullShortcodeKey) && !empty($this->wrappers[$fullShortcodeKey]) ) // eg: $NEWS_WRAPPER['view']['item']['NEWSIMAGE: item=1']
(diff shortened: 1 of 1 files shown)


CaMer0n:
It appears you're trying to parse every shortcode twice. I don't think this is a good idea.

rica-carv:
Corrected to only parse wrapper, and some minor code changes

CaMer0n:
I believe it would good for performance to check if `$pre` and `$post` have a shortcode in it before attempting to run `preg_replace_callback('#\{([A-Z][^\x02]*?\S)\}#', array(&$this, 'doCode'), $pre/$post) `   what do you think?

rica-carv:
Something like this?
```
$match='#\{([A-Z][^\x02]*?\S)\}#';

preg_match($match, $pre/$post)
?
preg_replace_callback($match, array(&$this, 'doCode'), $pre/$post)
:
$pre/$post;
```

CaMer0n:
Thanks. I just committed something similar. Please test.

rica-carv:
@CaMer0n Thanks, it works.
No we can get rid of those hardcoded HTML strings inside shortcodes (i hope...)

rica-carv:
@CaMer0n Sorry, but it seems it works only with shortcodes (new or overriden) from the theme, not with shortcodes from the plugins...
I've tested it with custom forum shortcodes in a theme (ie: `$SC_WRAPPER['THREADSX'] = "<a href='{FORUMTHEMESHORTCODE}'>{---}</a>";`), and it works, but the shortcodes form the plugin doesn't (ie: `$SC_WRAPPER['THREADSX'] = "<a href='{FORUMNAME}'>{---}</a>";`)...

CaMer0n:
Please try now. There's a commented-out test on line 139 of forum_template.php

rica-carv:
@CaMer0n Thanks, now it works as intended....
Proceeding with HTML cleanup... :smile:

PR #2031, opened 2016-11-12
Dashboard

lonalore:


CaMer0n:
Looks great. The only part I don't like is `$ADMIN_HEADER_FLEXPANEL` . Can you adapt `$ADMIN_HEADER` to function correct for the flexpanel and infopanel?

lonalore:
I did not want to use condition in a template file, but okay. Please check my last commit.

PR #2033, opened 2016-11-16
Forum shortcodes HTML removal

rica-carv:
Some HTML removal & new forum shortcodes for some more flexibility output....
HTML should be in templates, right?

Leaved shortcode {LASTPOST} for legacy reasons....


e107_plugins/forum/shortcodes/batch/forum_shortcodes.php
```diff
@@ -658,6 +671,20 @@ function sc_lastpost()
...
+
+	function sc_lastpostlink()
+	{
+        return $this->lastpostdata('link');
+	}
+
+	function sc_lastpostname()
+	{
+        return $this->lastpostdata('name');
+	}
```
CaMer0n: This is why we have $parms, to avoid situations like this..  lastpostxxxxx()
What we need here is sc_lastpost($parm)
check for $parm['type'] == 'link' etc.

CaMer0n:
We have $parms, to avoid situations like: ` sc_lastpostxxxxx()`
What we need here is `sc_lastpost($parm)`
check for $parm['type'] == 'link' etc. 

The ones you see there are due to backward compaibility.

rica-carv:
@CaMer0n Ok, got it. So i can rewrite my proposal with parms, inclunding new shortcodes to replace old ones, but keeping the legacy old ones, right?

CaMer0n:
@rica-carv Yes, I made a start on it for you, please see the last commit mentioned above.

PR #2036, opened 2016-11-16
Patch forum search shortcode

rica-carv:
Patch forum search shortcode,
Somehow when removing vars, the issue corrected by #1603 wasn't corrected...


CaMer0n:
Maybe I'm missing something, but it looks like you just moved the hidden tags inside the same form. How does it change anything?

rica-carv:
@CaMer0n Yes, the simple move of the hidden tags from the form end to the form start just makes this "bug", and that's what i did previously with #1603....

PR #2039, opened 2016-11-16
Enable templating of viewtopic caption

rica-carv:
Enable templating of viewtopic caption


e107_plugins/forum/forum_viewtopic.php
```diff
@@ -615,7 +616,7 @@ function forumbuttons($thread)
 
 if ($forum->prefs->get('enclose'))
 {
-	$forumTitle = e107::pref('forum','title', LAN_PLUGIN_FORUM_NAME);
+	$forumTitle = varset($FORUMCAPTION) ? $tp->parseTemplate($FORUMCAPTION, TRUE, $sc) : e107::pref('forum','title', LAN_PLUGIN_FORUM_NAME);
```
CaMer0n: Just use `!empty($FORUMCAPTION) `instead, and it should be good to go. Thanks.

rica-carv: Donne.
Thought that it was better to use the core varset function....
Is the varset core function to be deprecated?

CaMer0n:
Thank you! btw. For the future - I'm trying to use lowercase 'true' and 'false' and not "TRUE" and "FALSE"

rica-carv:
Err, i copy that code from somewhere else, and it was already uppercase.
Anyway, i'll keep in mind to lowercase the 'TRUE' & 'FALSE' i find out when changing code...
Sorry...

PR #2044, opened 2016-11-17
Fixes #1712 Plugin Download and Install from Market

### LINKED ISSUE #1712: Unable to install plugins or themes via admin web interface....
I'm having a "Can't continue" error when i try to install plugins or themes via the web interface on the admin area....
Any ideas what might be causing it???

MikeyGMT:
Fixes #1712 Plugin Download and Install from Market - Plus added debug statements around plugin download. Tested on 3 sites.

### DIFF
--- a/e107_admin/plugin.php
+++ b/e107_admin/plugin.php
@@ -2,7 +2,7 @@
 /*
  * e107 website system
  *
- * Copyright (C) 2008-2013 e107 Inc (e107.org)
+ * Copyright (C) 2008-2016 e107 Inc (e107.org)
  * Released under the terms and conditions of the
  * GNU General Public License (http://www.gnu.org/licenses/gpl.txt)
  *
@@ -34,12 +34,17 @@ $plugin = new e107plugin;
 $pman = new pluginManager;
 define("e_PAGETITLE",ADLAN_98." - ".$pman->pagetitle);
 
+e107::getDebug()->log("Plugin Manager.");
+
 if(e_AJAX_REQUEST) // Ajax 
 {
 	print_a($_POST);
 	print_a($_GET);
-	exit; 	
-	
+	exit;
+}
+else 
+{
+	e107::getDebug()->log("Not Ajax");
 }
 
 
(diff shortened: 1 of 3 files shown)


e107_admin/plugin.php
```diff
@@ -665,51 +680,59 @@ private function pluginDownload()
...
+		$this->id         = !empty($tmp[1]) 	? intval($tmp[1]) 	: "";//mikeyfixed
+		$this->id         =  empty($this->id) 	? $plugin_id 		: $this -> id;//mikeyfixed
 	//	print_a($_GET); 	
-		
 		$string =  base64_decode($_GET['src']);	
+
 		parse_str($string, $data);
 
-		if(!empty($data['plugin_price']))
+		if(!empty($plugin_price))
```
CaMer0n: Why not use $data, is it empty?

e107_handlers/e_marketplace.php
```diff
@@ -492,26 +538,30 @@ class e_marketplace_adapter_wsdl extends e_marketplace_adapter_abstract
...
-			'exception' 			=> true,
-		    "uri" 					=> "http://server.soap.e107.inc.com/",
-		    'cache_wsdl'			=> WSDL_CACHE_NONE,
-		    'connection_timeout' 	=> 60,
+			"trace"					=> true, 
+			'exception'				=> true,
+			"uri"					=> "http://server.soap.e107.inc.com/",
+			'cache_wsdl'			=> WSDL_CACHE_NONE,
+			'connection_timeout'	=> 60,
 		);
```
CaMer0n: Your tabs/spaces are showing as changes. Makes it difficult to track what you changed. :-(

CaMer0n:
@MikeyGMT Thank you. I'd prefer not to clutter the debug log if possible. Was the issue the lack of curly brackets on the `else `?

e107_admin/plugin.php
```diff
@@ -665,51 +680,59 @@ private function pluginDownload()
 		$frm = e107::getForm();
 		$mes = e107::getMessage();
 		
+		e107::getDebug()->log("Plugin Download.");
+		$tmp = explode('&amp;', e_QUERY); // #1712 we need to process new ampersand separated query
```
MikeyGMT: Best way to learn is have a go. 
What I found was $_GET only contained mode=download, there was no other data in there.
src is encrypted and in e_QUERY and that $tmp was already trying to grab values from exploding e_QUERY using .dot. separator. 
However, we got &amp; separators. So I added check to see id e_query contained &'s then exploded with that.

e107_admin/plugin.php
MikeyGMT: Yup data was empty ...

e107_admin/plugin.php
MikeyGMT: I was unsure about leaving debug calls - they only show up in debug mode... when we need them. Agree it adds clutter. Will trim them down. 

Good learning experience. I suppose, why was $_GET and $data are empty may be a better fix ? I'm online today. Busy this evening and weekend.

e107_admin/plugin.php
MikeyGMT: Check details here... 
 https://docs.google.com/document/d/1hbq645DZHW2Zy1oSos71yJMiqHbEk06BObM5HvzWNgQ

I'm back after the weekend :)

e107_admin/plugin.php
CaMer0n: I can't reproduce any of these issues on my local machine. If you can grant me FTP access to a machine that has this issue, I could run some tests. I have a hunch it is the use of '&amp' in the URL. I will commit that change and we can see how that affects things.

CaMer0n:
@MikeyGMT Do I understand that this issue is resolved?

PR #2052, opened 2016-11-21
Update forum_viewforum.php (replaces #2050)

rica-carv:
Permisions check should be on post, not thread.
With thread, the "create new topic" is displayed in the dropdown, and it shouldn't if user is not allowed to create posts....


CaMer0n:
thanks @rica-carv  - makes sense. What are the other changes you have added to this pull-request?

rica-carv:
I've made changes on the dropdown button, so that instead of displaying the "Start new topic" when a user isn't allowed, to display "Forum options"...

Another ideia would be to display the "Start new topic", but make it disabled if a user is not allowed to create new topics....

rica-carv:
Replaced #2050

PR #2064, opened 2016-11-25
Update e107_languages/English/admin/lan_plugin.php

yesszus:
Add pharses.


CaMer0n:
Thanks, but we should never translate anything inside addDebug() or any other e107::getDebug() command.

yesszus:
Ok. Restored!

CaMer0n:
Thank you @yesszus !

PR #2065, opened 2016-11-25
Forum userlist shortcode wrapping

rica-carv:
Removed hardcoded html and use wrapping instead, for userlist shortcode...


CaMer0n:
Thanks. Tested with bootstrap3 already?

rica-carv:
@CaMer0n I've tested already in bootsrtap3 theme, and it works, but it seems that the {USERLIST} shortcode is only used in old legacy themes (anyway, i'm personally using it on a bootstrap theme of my own.... 😉 ), so i'm on a dillema:
Should it use the new wrapper mode or the legacy wrap mode?

CaMer0n:
This will break for anyone using a pre v2.x template with this forum shortcode. Why don't you just create a new one called USERLISTX ?

rica-carv:
I've tested it with jayya, and the break seems to be not my changes, but the attempt to load theme_shortcodes from the jayya theme:
```
Couldn't Load: e107_themes/jayya/theme_shortcodes.php with class-name: theme_shortcodes and pluginName
Function: 	e_parse_shortcode->getScObject()()
File: 	G:\PortableApps\UniServerZ\www\e107-master\e107_handlers\e107_class.php
Line: 	1125
```

Anyway, i could do the extra shortcode, but it would simply duplicate code from the legacy shortcode....

PR #2079, opened 2016-11-29
Enable forum_post custom caption

rica-carv:
Enable forum_post custom caption

P.S.: Why does this forum_post.php file is such different form the other forum .php files?
It's quite hard to go throught this code, since it's a class inside the file, and it's quite a "overkill" for a forum post form.... Even the template handling is a "mess": after defined text, call internal render function that after all just calls core render... Real odd...
Or i miss something here?

### DIFF
--- a/e107_plugins/forum/forum_post.php
+++ b/e107_plugins/forum/forum_post.php
@@ -859,6 +859,11 @@ class forum_post_handler
 		$sc         = e107::getScBatch('post', 'forum')->setScVar('forum', $this->forumObj)->setScVar('threadInfo', vartrue($data))->setVars($data);
 		$text       = e107::getParser()->parseTemplate($template['form'], true, $sc);
 
+		if(!empty($template['caption']))
+		{
+      			$this->forumObj->prefs->set('title', $template['caption']);
+    		}
+
 		$this->render($text);
 
 		if(empty($data))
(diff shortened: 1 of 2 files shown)


CaMer0n:
I re-wrote it from scratch - this is why it is different to the others. Rest assured, everything there is for a reason. ;-)

rica-carv:
@CaMer0n That code won't take out my sleep, not a problem for me. Just though it could be "cleaner" & easier to mantain...

e107_plugins/forum/forum_post.php
```diff
@@ -859,6 +859,11 @@ function renderForm()
 		$sc         = e107::getScBatch('post', 'forum')->setScVar('forum', $this->forumObj)->setScVar('threadInfo', vartrue($data))->setVars($data);
 		$text       = e107::getParser()->parseTemplate($template['form'], true, $sc);
 
+		if(!empty($template['caption']))
+		{
+      			$this->forumObj->prefs->set('title', e107::getParser()->parseTemplate($template['caption'], true, $sc));
+    		}
+
 		$this->render($text);
```
rica-carv: Yes, i have donne it there because i got an empty $template variable if i try to get the template again before the table render inside [function render](https://github.com/rica-carv/e107/blob/58009dd3c54c353808d55617506a7177e52fbfd6/e107_plugins/forum/forum_post.php#L969)

If you have a workaround....

CaMer0n:
@rica-carv does my commit work for you? It works for me. (without overriding the title for every forum page too)

rica-carv:
@CaMer0n Yes, it does work...
It's getting recursive, you correcting my pulls.... 😆

PR #2081, opened 2016-11-29
Wrappers in forum post template

rica-carv:
Wrappers in forum post template.

P.S.: Forum_post_options wrapper replaces existing {FORUM_POST_OPTIONS_LABEL} shortcode. Is it safe to remove?

### DIFF
--- a/e107_plugins/forum/templates/forum_post_template.php
+++ b/e107_plugins/forum/templates/forum_post_template.php
@@ -13,21 +13,35 @@ if (!defined('e107_INIT')) { exit; }
 // New in v2.x - requires a bootstrap theme be loaded.  
 
 //$FORUM_POST_TEMPLATE['caption']		= "Custom caption";
+
+$SC_WRAPPER['FORUM_POST_AUTHOR'] = "<div class='form-group'>
+										<label for='name' class='col-sm-3 control-label'>".LAN_FORUM_3010."</label>
+										 <div class='col-sm-9'>{---}</div>
+									</div>
+";
+$SC_WRAPPER['FORUM_POST_SUBJECT'] = "<div class='form-group'>
+										<label for='name' class='col-sm-3 control-label'>".LAN_FORUM_3011."</label>
+										 <div class='col-sm-9'>{---}</div>
+									</div>
+";
+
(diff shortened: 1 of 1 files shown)


e107_plugins/forum/templates/forum_post_template.php
```diff
@@ -13,21 +13,35 @@
...
+";
+
+// Replaces legacy? {FORUM_POST_OPTIONS_LABEL} shortcode...
+$SC_WRAPPER['FORUM_POST_OPTIONS'] = "<div class='form-group'>
+										<label for='name' class='col-sm-3 control-label'>".LAN_FORUM_8013."</label>
+										 <div class='col-sm-9'>{---}</div>
+									</div>
+";
+
+
```
CaMer0n: $SC_WRAPPER is deprecated. If this only applies to the v2.x template we should use $FORUM_POST_WRAPPER

rica-carv: I know $SC_WRAPPER is deprecated, or better, for legacy themes, but i try to use the $FORUM_POST_WRAPPER and it doesn't works....
So i stick to the legacy one....

CaMer0n: Should be fixed now. Please try again. Thanks.

rica-carv: Thanks, it works now.

CaMer0n: Yes, just replace $SC_WRAPPER with $FORUM_POST_WRAPPER and it should be fine. Not sure how this change will impact old v1.x themes without bootstrap.

CaMer0n: I think you may be missing my point. If you have removed HTML from the shortcode and put it in a wrapper, and a v1.x theme is using a custom template, it will be missing that HTML unless they manually edit their theme and add those wrappers. This is my concern.

CaMer0n: Maybe a flag to a legacy template. and then pass it to the shortcode handler.

rica-carv:
Updated to v2.x wrappers...

Issue #2055, opened 2016-11-22
How to set a "scope" for Github provider in HybridAuth?


lonalore:
I enabled Github Auth, but it creates accounts without email address because of missing "user:email" scope. How can I set it without modifying core file:

https://github.com/e107inc/e107/blob/master/e107_handlers/hybridauth/Hybrid/Providers/Github.php#L15

If I am not wrong, it should be set in this constructor:
https://github.com/e107inc/e107/blob/master/e107_handlers/user_handler.php#L1102-L1119

Can I alter Hybrid Auth config somehow?

lonalore:
Btw it's unable to save usersettings without email... email field is missing completelly:

![kepernyokep_2016-11-26_12-49-28](https://cloud.githubusercontent.com/assets/6043821/20640163/51d0f934-b3d7-11e6-80d3-32228aa2b453.png)

lonalore:
@CaMer0n  Can I create a pull-request with these shortcode modifications?

Of course we can use `$options['disabled'] = true;` instead of `$options['readonly'] = true;`. But in that case need to change `usersettings.php` too.

CaMer0n:
Yes thanks. Okay for the pull-request. 👍  For the 'scrope' please see social/admin_config.php - we could consider adding a form field for 'scope'.

lonalore:
Tested on my website. Works properly! And the admin looks like this:

![kepernyokep_2016-11-30_19-24-59](https://cloud.githubusercontent.com/assets/6043821/20765736/eeff1efc-b733-11e6-9af8-7d225bbeb0bd.png)

PR #2084, opened 2016-11-30
Issue #2055 - Ability to set a "scope" for HybridAuth provider.

lonalore:
Issue #2055


CaMer0n:
Thank you!

Deltik:
This pull request causes #4115.  The email field is hard-coded to be required at all times, regardless of whether alternative login systems are enabled.

Issue #2083, opened 2016-11-30
'user_xup_login' event is not triggered?


lonalore:
I use Github authentication on my website, and I want to store/update Access Token when a user logs in... but 'user_xup_login' event is not triggered. :/

I use the new `e_event.php` addon file for registering my event callback functions (other events are triggered, so the addon file works properly):
```php
class PLUGIN_event
{

	function config()
	{
		$event = array();

		$event[] = array(
			'name'     => 'user_xup_login',
			'function' => 'PLUGIN_user_login_callback',
		);

		return $event;
	}

	function PLUGIN_user_login_callback($data)
	{
		e107::getLog()->add('XUP_LOGIN', (array) $data);
	}
}
```

CaMer0n:
Please feel free to add it where it needs to go. 

	e107::getEvent()->trigger('user_xup_login', $userdata);

PR #2088, opened 2016-12-01
Issue #2083 - Fix for "user_xup_login" event triggering... loginProvi…

lonalore:
Issue #2083 - Fix for "user_xup_login" event triggering... loginProvider() method call used 2 parameters, but method uses 1 parameter.

### DIFF
--- a/e107_handlers/user_model.php
+++ b/e107_handlers/user_model.php
@@ -1586,7 +1586,10 @@ class e_user extends e_user_model
 		
 		$userdata  = $userlogin->getUserData();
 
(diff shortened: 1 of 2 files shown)


e107_handlers/user_model.php
```diff
@@ -1757,6 +1760,8 @@ public function tryProviderSession($deniedAs)
 
 				if($sql->update('user', $updateQry) !==false)
 				{
+					$updatedProfile = array_replace($user, $userdata);
+					e107::getEvent()->trigger('user_xup_profile_updated', $updatedProfile);
```
CaMer0n: XUP = eXternal User Profile
So, `user_xup_updated` might be more appropriate than `user_xup_profile_updated`. 
All the rest looks great! 👍

CaMer0n:
Thank you!!! 👍

PR #2092, opened 2016-12-01
Theme Layout improvement - Allow to use "*" wildcard in a path for theme layouts.

lonalore:
kepernyokep_2016-12-01_14-36-52

It works with SEF and Legacy URL too:
- /e107_plugins/[plugin]/project.php
- /project/[user]/[project]
- /project/[user]

### DIFF
--- a/class2.php
+++ b/class2.php
@@ -1238,6 +1238,17 @@ if(!defined("THEME_LAYOUT"))
 					$def =  $lyout;
 					break 2;
 				}
+
+				// Allow to use "*" wildcard.
+				if(!empty($kpage))
+				{
+					$match = matchPath($c_url, $kpage);
+					if($match)
+					{
+						$def = $lyout;
+						break 2;
+					}
+				}
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


rica-carv:
Humm, maybe with some modifications it could also solve [this](http://e107help.org/2824/how-to-get-real-php-filename-if-sef-urls-are-enabled?show=2982#c2982), since currently i can't enable a menu with SEF urls...

lonalore:
@rica-carv did you try to use `e_URL_LEGACY`? It should also work when SEF URLs are enabled.

```php
if(defset('e_URL_LEGACY') == 'e107_plugins/plugin/file.php')
{
    // Load menu...
}
```

rica-carv:
@lonalore Just tryed.
For some reason, it will only be defined when a page is rendered via SEF url. When a page is rendered via regular url, it's not defined.
The opposite occurs for e_PAGE....
Guess i'll have to end up with both constants, ie:
`If (e_PAGE?:e_URL_LEGACY)...`

lonalore:
I think the best solution would be moving [this Simple URL ReWrite code](https://github.com/e107inc/e107/blob/master/index.php#L60-L160) to `class2.php`. And put a condition after it.

```php
if(!defset('e_URL_LEGACY', false))
{
	define('e_URL_LEGACY', ltrim(e_REQUEST_URI, '/'));
}
```

So, `e_URL_LEGACY` would be available globally, and it would always contain the legacy URL.

rica-carv:
@lonalore Well, for now ended up solving it this way, if it matters:
`$check_url = e_PAGE!="e_PAGE" ? e_PAGE : end(explode('/', parse_url(e_URL_LEGACY)['path']));`

CaMer0n:
@lonalore I'm just checking that you're aware that currently, without your changes, you could just use "project" (as per your example) and it should work for:
-  /project
- /project/something/
- project.php
- e107_plugins/project/something.php 

So I'm not quite seeing the full value of what you've proposed. Can you explain more?

lonalore:
Wow, indeed! It never ever occurred to me to try this. Thank you!

PR #2096, opened 2016-12-02
Banner Lans to Generic Terms #6

MikeyGMT:
Replace duplicates with common terms in Banner Plugin.
Created this TODO term also LAN_DEFAULT_TABLE_DATA - for use in other
plugins too.

### DIFF
--- a/e107_admin/core_image.php
+++ b/e107_admin/core_image.php
@@ -1175,6 +1175,7 @@ $core_image = array (
       'banner_sql.php' => '950a1eda9eed1f1d735db4231dafc09d',
       'banner_template.php' => '6cce6a757e0a9aa609ec52e04db9e527',
       'e_menu.php' => '8173c4db24ee86f28a6cc0416b7f42f1',
+      'e_help.php' => 'TODO-Newfile',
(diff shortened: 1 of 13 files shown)


MikeyGMT:
Testing 
image

image


image

image

image

Note - should this be save not close after configuration changes made...?

image

MikeyGMT:
Let me know if it is incorrect to comment this menu option out. My bad, happy to uncomment. 

I was going to shorten to "Configure Menu" so it fits in the left column, but saw that it's no longer done here... 

e107_plugins/banner/admin_banner.php 

`//	'main/menu'			=> array('caption'=> BNRLAN_36, 'perm' => 'P'), //NOW CONFIGURED IN MENU MGR`

MikeyGMT:
Reverted and shortened to fit. 
image

CaMer0n:
I think commented-out is better.  Maybe we need to work on the help file to explain where one can configure specific instances via Menu-Manager.

MikeyGMT:
Testing - this is the first stab at help - I think it can grow. 

image

Also I have added the filename and a TODO message in CORE_IMAGE.PHP which drives file inspector.

MikeyGMT:
File inspector entry (Result Core File Unchecked)
image

PR #2097, opened 2016-12-02
RSS Menu Lans to Generic Terms #6

MikeyGMT:
Replace duplicates with common terms in RSS Plugin


MikeyGMT:
Testing 

image

image

image

image

e107_plugins/rss_menu/admin_prefs.php
```diff
@@ -94,19 +94,19 @@ class rss_ui extends e_admin_ui
...
-		  'rss_path'        =>   array ( 'title' => 'Plugin-Dir',   'type' => 'text', 'data' => 'str', 'readonly'=>1, 'width' => 'auto', 'help' => '', 'readParms' => '', 'writeParms' => '', 'class' => 'left', 'thclass' => 'left',  ),
+		  'rss_path'        =>   array ( 'title' => LAN_PLUGIN_FOLDER,'type' => 'text', 'data' => 'str', 'readonly'=>1, 'width' => 'auto', 'help' => '', 'readParms' => '', 'writeParms' => '', 'class' => 'left', 'thclass' => 'left',  ),
 		  'rss_url'         =>   array ( 'title' => LAN_URL,        'type' => 'method', 'data' => 'str', 'width' => 'auto', 'help' => '', 'readParms' => '', 'writeParms' => '', 'class' => 'left', 'thclass' => 'left',  ),
-		  'rss_topicid'     =>   array ( 'title' => 'Topic id',     'type' => 'text', 'data' => 'int', 'width' => 'auto', 'help' => '', 'readParms' => '', 'writeParms' => '', 'class' => 'center', 'thclass' => 'center',  ),
+		  'rss_topicid'     =>   array ( 'title' => LAN_ID,         'type' => 'text', 'data' => 'int', 'width' => 'auto', 'help' => '', 'readParms' => '', 'writeParms' => '', 'class' => 'center', 'thclass' => 'center',  ),
```
CaMer0n: I believe this needs to be Topic Id, otherwise we have 2 fields with the same title.

CaMer0n:
Thank you @MikeyGMT for this! 👍 (please see small issue in comment above)

e107_plugins/rss_menu/admin_prefs.php
MikeyGMT: Ooo OK, will add Topic ID

MikeyGMT:
Fixed 
image

image

Not sure I understand how code climate assumes we can never get to this?  

> The method dboptions uses an else expression. Else is never necessary and you can simplify the code to work without else.

NEW

```
        {
            return LAN_NOCHANGE_NOTSAVED;
        }
```

Issue #2101, opened 2016-12-03
Not subforum after update v1 - v2


fizi:
I updated site from Github. Forums appears, but no subforums (blank page).
Server log: 
[Sat Dec 03 10:04:22 2016] [error] [client --------------] PHP Parse error:  syntax error, unexpected '[' in /chroot/home/------/------/html/e107_plugins/forum/shortcodes/batch/viewforum_shortcodes.php on line 536, referer: http://---------/e107_plugins/forum/forum.php[Sat Dec 03 10:09:27 2016] [error] [client -------------] PHP Parse error:  syntax error, unexpected '[' in /chroot/home/-------/-----------/html/e107_plugins/forum/shortcodes/batch/viewforum_shortcodes.php on line 657, referer: http://----------/e107_plugins/forum/forum.php

If I delete the part of code (viewforum_shortcodes.php): 
$caller = debug_backtrace()[1]['function']; - on line 536 
$caller = debug_backtrace()[1]['function']; - on line 657
subforum appears, but no last post, this cell is empty.

rica-carv:
Which PHP version are you using?

fizi:
PHP Verzió
5.3.3
mySQL
5.1.73

rica-carv:
Hum, you upgraded from v1.x to v2.
Have you copied the templates from `e107_core\templates` to `e107_themes\templates`?
Beware, if you customized the `e107_themes\templates` they'll be overwritten, so make a backup copy of them first, and afterwards replace the customization in the new template files...
Give it a try, but for now i can't remember anything more that could solve this "bug"...

Mine subforums are fully working on a clean v2 install, so i can only figure it out as old legacy code remains giving this "bug"...

fizi:
I deleted e107/templates folder.
I made a number of unique template and I placed into my theme/templates folder.

CaMer0n:
> Have you copied the templates from e107_core\templates to e107_themes\templates?
This should NOT be necessary. If it is, then there is a bug to be corrected. 

I suggest upgrading PHP to 5.6 (or at least 5.4)

fizi:
Unfortunately, the service does not want to upgrade. :(

rica-carv:
@CaMer0n 
The copy might be necessary, since i've had real issues with wrong templates from a upgrade from v1.4 to v2, and that copy process simply solve it.
@fizi Try to restore the deleted folder, just to see if it works or not.

tgtje:
[...]

PR #2104, opened 2016-12-03
Issue #2101 - Fixes for fatal error on subforum pages.

lonalore:


rica-carv:
Well, it seems that "simplified" PHP code (ie: less variables used) is no longer an option..
Too bad...

PR #2109, opened 2016-12-04
Lans Cleanup auth.php #6

MikeyGMT:
Lans Cleanup auth.php #6

### DIFF
--- a/e107_admin/auth.php
+++ b/e107_admin/auth.php
@@ -66,11 +66,11 @@ if (ADMIN)
 		// XXX LOGIN AS Temporary solution, we need something smarter, e.g. reserved message stack 'admin' which will be always printed
 		// inside admin area
 		if(e107::getUser()->getSessionDataAs())
-		{ // TODO - lan [logout]?
+		{  
 			$asuser = e107::getSystemUser(e107::getUser()->getSessionDataAs(), false);
 			
 			$lanVars = array ('x' => ($asuser->getId() ? $asuser->getName().' ('.$asuser->getValue('email').')' : 'unknown')) ;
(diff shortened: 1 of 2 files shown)


e107_admin/auth.php
```diff
@@ -66,9 +66,9 @@
 		// XXX LOGIN AS Temporary solution, we need something smarter, e.g. reserved message stack 'admin' which will be always printed
 		// inside admin area
 		if(e107::getUser()->getSessionDataAs())
-		{ // TODO - lan
+		{ // TODO - lan [logout]?
 			$asuser = e107::getSystemUser(e107::getUser()->getSessionDataAs(), false);
-			e107::getMessage()->addInfo('Successfully logged in as '.($asuser->getId()  ? $asuser->getName().' ('.$asuser->getValue('email').')' : 'unknown'). ' <a href="'.e_ADMIN_ABS.'users.php?mode=main&amp;action=logoutas">[logout]</a>');
+			e107::getMessage()->addInfo(ADLAN_164.' '.($asuser->getId()  ? $asuser->getName().' ('.$asuser->getValue('email').')' : 'unknown'). ' <a href="'.e_ADMIN_ABS.'users.php?mode=main&amp;action=logoutas">[logout]</a>');
 		}
```
CaMer0n: This needs an [x] in the LAN. ie. Successfully logged in as [x] and then [x] replaced with the variables. 
(Some languages may not place this info at the end of the sentence.)

MikeyGMT: Done but note there are duplicate messages coming from admin/users.php 
Note that the further instruction message is a little confusing. 

image

CaMer0n:
Great work. Thanks.

PR #2113, opened 2016-12-05
Correction for circular reference bug in wrapper

rica-carv:
This corrects the circular reference form {SHORTCODE} to SC_WRAPPER[SHORTCODE], when wrapper is like something this: $SC_WRAPPER['SHORTCODE'] = '{SHORTCODE}{---}';
Currently, it's a neverending loop....


CaMer0n:
Thank you! 👍

PR #2114, opened 2016-12-05
Error template misplaced styles?

rica-carv:
Error template misplaced styles?

In almost all the templates or HTML, the `alert-danger class` is on the `div `tag....


CaMer0n:
Thank you!

PR #2117, opened 2016-12-06
Just to avoid encoding problems with accented characters

OTroccaz:


yesszus:
Wrong characters this change: 2016 februÃ¡r 02 kedd.

OTroccaz:
In french, without : 12 d?cembre 2016
With : 
12 décembre 2016
I will keep it for myself!

OTroccaz:
... but there's still a problem ...

CaMer0n:
@OTroccaz Could there be a problem with French.php locale ?

OTroccaz:
Thanks for the reply @CaMer0n! I will update the language-pack tomorrow. I have made some tests but the problem remains ... and this is the only place where there is this problem.
I will search again.

CaMer0n:
Thank you! Maybe the server is missing the french locales?

rica-carv:
[Maybe some further reading for this issue here....](http://stackoverflow.com/questions/4095899/utf8-encoding-problem-with-good-examples)

OTroccaz:
I have progressed...
The server is under Windows and apparently it is not possible to define locales in utf-8.

> The set of available locale names, languages, country/region codes, and code pages includes all those supported by the Windows NLS API except code pages that require more than two bytes per character, such as UTF-7 and UTF-8. If you provide a code page value of UTF-7 or UTF-8, setlocale will fail, returning NULL

There would therefore be at least two possibilities:
. Or, like the function convert_date, use isUTF8:
```
public function toDate($datestamp = null, $format='short')
	{
		if(!is_numeric($datestamp)){ return null; }
    
		$dateSLR = e107::getDate()->convert($datestamp, $format);
		if (e107::getParser()->isUTF8($dateSLR))
		{
			$dateSLR = utf8_encode($dateSLR);
		}
		return '<span data-livestamp="'.$datestamp.'">'.$dateSLR.'</span>';	
	}
```
. Either use IntlDateFormatter
Your opinions?

CaMer0n:
@OTroccaz Did that change work for you? 
Ideally, I would suggest moving to a Linux server. Other problems may arise with using Windows.

OTroccaz:
@CaMerOn Yes it works for me.
I can not migrate this server to Linux because there are other priority Windows applications.

PR #2121, opened 2016-12-06
Forum shortcodes HTML removal (replaces #2033)

rica-carv:
Forum shortcodes HTML removal (replaces #2033)

Some HTML removal & new forum shortcodes for some more flexibility output....
HTML should be in templates, right?

Leaved shortcode {LASTPOSTUSER} & {LASTPOSTDATE} for legacy reasons....
Two more shortcodes created inside {LASTPOST}:

- type=date -> renders latest thread date without link tag
- type=name -> renders latest thread name


CaMer0n:
Thank you! 👍

Issue #1951, opened 2016-10-20
New private message admin form working with server time


rica-carv:
The new private message, on the admin area has the server time stamp, not taking in consideration the time offset setup on preferences....

![untitled 1](https://cloud.githubusercontent.com/assets/17130530/19578504/de660be0-9713-11e6-99c3-8edff99611a9.jpg)
![untitled 2](https://cloud.githubusercontent.com/assets/17130530/19578505/de6a5dd0-9713-11e6-8efe-34e27be46fe6.jpg)
![untitled 3](https://cloud.githubusercontent.com/assets/17130530/19578550/1610f42e-9714-11e6-9662-c7700fcd8fda.jpg)

lonalore:
@rica-carv "Time offset" setting is not in used, and it should be removed. Time offset is set in [class2.php](https://github.com/e107inc/e107/blob/master/class2.php#L1145) using "Timezone" setting. In your case, Server-time and PM-sent time are "equals" because these values calculated using the same timezone due to [date_default_timezone_set()](https://github.com/e107inc/e107/blob/master/class2.php#L1145) (it sets the timezone globally).

CaMer0n:
Closing.

PR #2139, opened 2016-12-12
Timezone settings improvement: Display time-offset in timezone dropdown.

lonalore:
kepernyokep_2016-12-12_16-17-28

### DIFF
--- a/class2.php
+++ b/class2.php
@@ -1140,7 +1140,7 @@ if (($_SERVER['QUERY_STRING'] == 'logout')/* || (($pref['user_tracking'] == 'ses
 *
 */
 
-$tz = vartrue($pref['timezone'],'GMT'); //TODO Adjust on the front-end based on user timezone value. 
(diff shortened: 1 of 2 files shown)


lonalore:
@CaMer0n 
I think "Time offset" setting is unnecessary, because in PHP it's only for backward compatibility:
http://php.net/manual/en/timezones.others.php

I did a search and I noticed that only rss.php uses that setting. In my opinion we should rewrite rss.php for using a time-offset, which is calculated from "Timezone" setting. And we could remove that setting completely.

lonalore:
Furthermore, "Timezone" (extended user field) is not in use currently, according this:
https://github.com/e107inc/e107/blob/master/class2.php#L1143-L1145

So we also need to modify [extended_timezone](https://github.com/e107inc/e107/blob/master/e107_core/sql/extended_timezones.php) file to use the same list as "Timezone" admin pref. Then we could set user-defined timezones in class2.php.

lonalore:
Related issue: #1951

Issue #1973, opened 2016-10-29
How to remove news "list" heading?


rica-carv:
I'm trying to use the news "grid" in my template, to display news via a grid system.
Therefore, i'm using this template code copied from news_menu_template.php and adapted:
```
// Grid Menu
$NEWS_TEMPLATE['list']['start']    = '<div class="row">';
$NEWS_TEMPLATE['list']['item']		= '<div class="{NEWSGRID} col-sm-4"><div>
												{SETIMAGE:w=250&h=250}{NEWSTHUMBNAIL}<div>
              									{NEWS_TITLE: link=1}
              									<p>{NEWS_SUMMARY}</p>
                                </div>
            							   </div></div>';
$NEWS_TEMPLATE['list']['end']      = '<div class = "clearfix visible-xs"></div></div>';
```

But unfortunately, when using the list model to display news, it appears a heading on the page top, wich i don't want:
![untitled 1](https://cloud.githubusercontent.com/assets/17130530/19832318/f3f2502a-9e18-11e6-8b69-8534a6e55921.jpg)

Can this be removed via template? How?

The weird thing is that if i use the default model to display news, it doesn't appear, but i can't use the grid system with the default model, since it only renders the` $NEWS_TEMPLATE['default']['item']` definiton, anything more is ignored....

rica-carv:
Forget it, just found out we can template the caption...

PR #2142, opened 2016-12-13
File Inspector: Update Core Image #1973 and Lans Clean up #6 

MikeyGMT:
File Inspector: Update Core Image #1973 and Lans Clean up #6

### DIFF
--- a/e107_admin/core_image.php
+++ b/e107_admin/core_image.php
@@ -3,7 +3,7 @@
 + ----------------------------------------------------------------------------+
 |     e107 website system
 |
-|     Copyright (C) 2008-2010 e107 Inc. 
+|     Copyright (C) 2008-2016 e107 Inc. 
 |     http://e107.org
 |
 |     Released under the terms and conditions of the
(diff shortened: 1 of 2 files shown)


MikeyGMT:
Testing 
image

CaMer0n:
Core-image is auto-generated during release creation. So, it's not expected that core_image will function accurately with the github version of e107.

MikeyGMT:
@CaMer0n oh ok - I did ask. .... never mind. 
You need it reverting as there's LAN updated too.

MikeyGMT:
@CaMer0n I have 2.1.4 Git released 2 days ago, but there are still many non-core files reported - e.g. the "flexpanel" files, but on checking core_image file date, it does not look like it was updated during the release process, as mentioned above.  

image

I am trying to clean up a very old site, hence it's a needle and haystack process :( 

It may be good too, to optionally exclude the system folders, and the old newspost images, i.e. places where users are permitted to upload stuff. 

Thanks
Mikey

MikeyGMT:
Ditto for 2.1.5

Issue #2150, opened 2016-12-14
Forum button bounce and change the size


fizi:
If I logout and see the forum button "New Topic" - the button bounce and change the size.
Mouse out:
![forum_button_of](https://cloud.githubusercontent.com/assets/86780/21197744/d68b309a-c23c-11e6-979c-5e0397df7915.jpg)
Mouse on:
![forum_button_on](https://cloud.githubusercontent.com/assets/86780/21197758/e4a56588-c23c-11e6-8742-b42bb81d05f6.jpg)
Now I delete data-toggle='tooltip' (line: 97) from viewforum_shortcodes.php and button OK (no tooltip), if the button has tooltip the button bounce.

rica-carv:
Bootstrap issue. Fetching a workaround...

rica-carv:
@fizi Just to make sure, which bootstrap library version are you using on that theme?

fizi:
I'm using 3.3.7

rica-carv:
From wich CDN? Or is it local hosted?

Jimmi08:
Confirmed.  
Local 3.3.7
I noticed that there is wrong html markup too. (not solved this) 
http://getbootstrap.com/components/#btn-groups
- dropdown should be in div with class btn-group too.

rica-carv:
[Tooltips in button groups, input groups, and tables require special setting](http://getbootstrap.com/javascript/#tooltips)

I've tryed the `container: 'body'`, but for any unknown reason for me, it didn't work.
The weird thing is that i've tryed it in bootply, and it worked.
So, i came up with th extra HTML tag to solve it...
Not elegant, but it's the only way i find out to solve it...
And yes, i even tryed jquery javascript also...

fizi:
I modified viewforum_shortcodes.php file (94):
`$text .=
			'<a href="'.($this->var['ntUrl'] ?:"#").
			'" class="btn btn-primary'.($this->var['ntUrl'] ?"":" disabled").'"'
			.($this->var['ntUrl'] ?"":" data-toggle='tooltip' title='".LAN_FORUM_0006."'
			style='cursor: not-allowed; pointer-events: all !important;'").'>'.LAN_FORUM_1018.'</a>
      <div class="btn-group">
			  <button class="btn btn-primary dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
		    	<span class="caret"></span>
			  </button>
		    <ul class="dropdown-menu pull-right">`

[...]

PR #2151, opened 2016-12-14
Patch for #2150

rica-carv:
Patch for #2150 
Not very elegant, but it's the only way i find out to solve the button pop...
Bootstrap seems to have no solution for it...

### DIFF
--- a/e107_plugins/forum/shortcodes/batch/view_shortcodes.php
+++ b/e107_plugins/forum/shortcodes/batch/view_shortcodes.php
@@ -804,7 +804,7 @@ function sc_buttonsx()
 	}
 		$replyUrl = "<a class='btn btn-primary".($url ?"":" disabled")."' "
 			.($url?"":" data-toggle='tooltip' title='".LAN_FORUM_0046."'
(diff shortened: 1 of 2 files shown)


Moc:
@fizi Can you please check if this solves the issue?

e107_plugins/forum/shortcodes/batch/view_shortcodes.php
```diff
@@ -804,7 +804,7 @@ function sc_buttonsx()
 	}
 		$replyUrl = "<a class='btn btn-primary".($url ?"":" disabled")."' "
 			.($url?"":" data-toggle='tooltip' title='".LAN_FORUM_0046."'
-	style='cursor: not-allowed; pointer-events: all !important;'")." href='".($url ?:"#")."'>".LAN_FORUM_2006."</a>";
+	style='cursor: not-allowed; pointer-events: all !important;'")." href='".($url ?:"#")."'>".LAN_FORUM_2006."</a>".($url?"":"<span></span>");
```
CaMer0n: <span></span> isn't valid. Can you add an &nbsp; inside without messing up the layout?

CaMer0n:
Thank you! 👍

PR #2153, opened 2016-12-15
Introduce e107::getError() for delivering an error page to the br…

lonalore:
…owser.


lonalore:
Here is a simple example for usage of `e107::setErrorPage()` in a custom plugin:
https://github.com/lonalore/batch/blob/master/batch.php#L26

CaMer0n:
Thankyou!!

PR #2173, opened 2016-12-20
Code cleanup. Fixed undefined variables.

lonalore:


e107_admin/image.php
```diff
@@ -2576,6 +2576,8 @@ function deleteFileXml($imgFile)
 	
 	function batchDelete()
 	{
+		$tp = e107::getParser();
```
rica-carv: Define the $tp variable to use only once inside the function?
Ocuppy some more server memory why?

Issue #1755, opened 2016-07-01
Site contact info field rejecting bbcode


rica-carv:
I'm trying to inser this bbcode inside the contact info field at the admin ui preferences page: 
`[img width:50px]e107_media/0548bbec44/images/2016-07/24_7.png[/img]`
After i click update, the info returned is that "_Settings not saved as no changes were made._"
Strangely indeed, in the contact info field it retires the **width:50px** text:
`[img]e107_media/0548bbec44/images/2016-07/24_7.png[/img]`

Is this intended to work this way?
If it is, how can i add images with custom width and height on the contact info field?

And by the way, isn't anywhere a core bbcode reference chart for e107???

rica-carv:
@CaMer0n If i recall correctly, at the issue post time, i did, but i haven't tried it recently...
I'll try to do it soon and report back...
Thanks...

Moc:
@rica-carv any update?

rica-carv:
@Moc Yes. My first version,` [img width:50px]e107_media/0548bbec44/images/2016-07/24_7.png[/img]` is now saved (the correct  updated message is diplayed, but redirected to the main preferences admin ui page.... Weird...), but the code is still truncated to `[img]e107_media/0548bbec44/images/2016-07/24_7.png[/img]`

@CaMer0n's version, `[img style=width:50px]e107_media/0548bbec44/images/2016-07/24_7.png[/img]` somehow works (succesfull save message displayed correctly, but again redirected to preferences maisn adim ui page...), but now the text is truncated to this: `[img style=width%3A50px]e107_media/0548bbec44/images/2016-07/24_7.png[/img]`

So, basically, problem still persists.....

CaMer0n:
@rica-carv 
Did you try with this? (your example with `[img width:50px]` is using the incorrect markup) 

`[img width=50]e107_media/0548bbec44/images/2016-07/24_7.png[/img]`

Moc:
@rica-carv Any feedback?

Moc:
Closing due to lack of feedback. Feel free to add additional information and I'll re-open.

PR #2174, opened 2016-12-20
Fix for #1755. Use the newer filter method instead of toDb.

lonalore:
Issue #1755


CaMer0n:
This could have serious repercussions on prefs. We'd need to compare the results of both.

CaMer0n:
Closing as it doesn't bring any real benefit at this time.

Issue #2183, opened 2016-12-22
problem with menu manager - sometimes buttons for position doesn't work


Jimmi08:
Sometimes buttons move up or top don't work.  I don't know if it related but Console looks like this:

![image](https://cloud.githubusercontent.com/assets/5429548/21442961/a10f2f8c-c8a1-11e6-9e20-89e751d7a929.png)

Thanks

MikeyGMT:
Probably my lans work. I'll look tomorrow.

MikeyGMT:
I've checked my menus pull and cannot see anything that could stop it from working.  

However, I found a fix for the first error in menus.php on line 428 it is being called as js not css. 
Was
`	e107::js('url',	"http://ajax.googleapis.com/ajax/libs/jqueryui/1.10.1/themes/base/jquery-ui.css");`

Change to 
`	e107::css('url',	"http://ajax.googleapis.com/ajax/libs/jqueryui/1.10.1/themes/base/jquery-ui.css");`

And that error goes away. 

Continuing investigation....

MikeyGMT:
I have manually reverted lans work in 2177 on one site, it does not fix the move up/down problem. I have a site update 26 Nov, move up down is working there. I will apply the lans work there to see what happens.

MikeyGMT:
Looks related to this as when I add it to the 26 Nov version of menumanager_class.php then menus don't move up and down. 
 
https://github.com/e107inc/e107/commit/0b78d551ee2852123bdd2a04973d2453c30a0791

```
		$file = e107::getParser()->filter($file);
		$newurl = e_PLUGIN_ABS.$file."?id=".intval($_GET['id']).'&iframe=1';

```

mortalites:
Having same issue with not being able to move menus up or down. Sometimes the down and up option doesnt even show up. Sometimes a page refresh fixes it other times the menus still will not move.

PR #2185, opened 2016-12-23
Menu Manager calling jquery-ui.css with js method #2183

MikeyGMT:
Menu Manager calling jquery-ui.css with js method part of #2183


CaMer0n:
Well spotted. Thanks

PR #2186, opened 2016-12-23
Replace jsapi with loader.js in chart_class

MikeyGMT:
This avoids 2 warnings when opening admin.
jsapi:22 A Parser-blocking, cross-origin script,
https://www.google.com/uds/?file=visualization&v=1&packages=corechart,
is invoked via document.write. This may be blocked by the browser if the
device has poor network connectivity. See
https://www.chromestatus.com/feature/5718547946799104 for more details.
Reference information:

http://stackoverflow.com/questions/39719334/javascript-google-visualization-selection-option

### DIFF
--- a/e107_handlers/chart_class.php
+++ b/e107_handlers/chart_class.php
@@ -2,7 +2,7 @@
  /*
  * e107 website system
  *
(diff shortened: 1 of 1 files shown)


CaMer0n:
Excellent! Thank you!

PR #2191, opened 2016-12-25
LANS in Media Class #6

MikeyGMT:
LANS in Media Class #6 - although there is plenty left to do but many
seem to be debug messages rather than user prompts.


MikeyGMT:
Interesting codeclimate comment "Avoid unused private methods such as 'imageSelect'."
Particularly as I could not find where this is used for testing.

Issue #2217, opened 2016-12-30
PM_COMPOSE shortcode not working


pricey1981:
On the PM Inbox page the 'Send New Message' button is not working (there is nothing in the href attribute of the link). I've tracked this through to the shortcode and the following line returns null...

`$urlCompose = e107::url('pm','index','', array('query'=>array('mode'=>'send')));`

The resulting output is `<th style='width:100px'><a class='btn btn-sm btn-primary btn-block-level' href=''><i class='fa fa-edit'></i> Send New Message</a></th>`

Any idea what the issue is?

CaMer0n:
Could you please share a list of the files located in your e107_plugins/pm folder ?

pricey1981:
It's the latest from github

images/
languages/
admin_config.php
e_cron.php
e_shortcode.php
e_url.php
index.html
plugin.xml
pm.css
pm.php
pm_class.php
pm_conf.php
pm_default.php
pm_func.php
pm_notes.txt
pm_setup.php
pm_shortcodes.php
pm_sql.php
pm_template.php
private_msg_menu.php

CaMer0n:
Thanks.. are you using SEF Urls or regular?

pricey1981:
@CaMer0n SEF urls

CaMer0n:
@pricey1981 Are you using the latest files from here? Also, is your .htaccess file up-to-date (ie. with code from e107.htaccess )

pricey1981:
I initially getting this issue with 2.1.2 so downloaded the github PM files. So I'm using 2.1.2, but with the github PM files.

Moc:
It may very well be that this issue is already resolved in the latest update. Please try updating from Github before we look into this issue more closely. 

There is an easy way to sync with Github:
- Enable developer mode in `Admin Area > Preferences > Advanced features`
- Go to `Admin Area > Tools > Database > Sync with Github`
- Press the `Overwrite files` button and wait until the process has finished. 

Alternatively you can download the ZIP file with all the latest files included by hitting the green 'Clone or download' button at the top of this page: https://github.com/e107inc/e107

Moc:
Closing this due to lack of feedback. Please let us know if the issue is still present, then I'll re-open

PR #2218, opened 2016-12-30
Update pm_shortcodes.php

pricey1981:
missing . from line 588 where $extra is being checked. Had the effect of presenting an incorrect link in the outbox.


CaMer0n:
Do I understand well that this fixes issue #2217 ?

pricey1981:
unfortunately not. This was a separate issue I spotted when tidying up my PMs

CaMer0n:
okay, thanks.

PR #2219, opened 2016-12-30
Text To LANS #6 infopanel

MikeyGMT:
Text To LANS #6 infopanel

### DIFF
--- a/e107_admin/includes/infopanel.php
+++ b/e107_admin/includes/infopanel.php
@@ -2,14 +2,12 @@
 /*
  * e107 website system
  *
- * Copyright (C) 2008-2013 e107 Inc (e107.org)
+ * Copyright (C) 2008-2016 e107 Inc (e107.org)
  * Released under the terms and conditions of the
  * GNU General Public License (http://www.gnu.org/licenses/gpl.txt)
  *
  * Info panel admin view
  *
- * $URL$
- * $Id$
  */
 
 if (!defined('e107_INIT'))
@@ -160,7 +158,6 @@ class adminstyle_infopanel
 
 		';
 	*/	
-		//TODO LANs throughout.
 
 		$user_pref = $this->getUserPref();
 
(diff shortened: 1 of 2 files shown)


MikeyGMT:
Testing 
image

Moc:
@CaMer0n Would it be an idea to have the months (January, February, etc.) defined in the English common LAN?

MikeyGMT:
I thought that too... but was reluctant to do so. I can always revisit.

MikeyGMT:
I'm actually thinking there's a way to call a formatted date month only and then no translation needed.

MikeyGMT:
OK ... so testing this (not committed yet)

```
			$data['labels'] 	= array(strftime("%B",mktime(0,0,0,01,01,16)), //"January",
							strftime("%B",mktime(0,0,0,02,01,16)), //"February",
							strftime("%B",mktime(0,0,0,03,01,16)), //"March",
							strftime("%B",mktime(0,0,0,04,01,16)), //"April",
							strftime("%B",mktime(0,0,0,05,01,16)), //"May",
							strftime("%B",mktime(0,0,0,06,01,16)), //"June",
							strftime("%B",mktime(0,0,0,07,01,16))  //"July"
							);
```
Returns the correct months in English. Switch e107 admin to Welsh ... still returns English Month Names. 

Hmmm downloading another lan and testing with that.

MikeyGMT:
Tested with NL, PT, same result ... 

OK so in the line above I set 
`setlocale(LC_ALL,"hu_HU.UTF8");`
The months have come out as the intended. 
So does this mean the site language is not observed in all scripts?

MikeyGMT:
Aha ok ... So I was switching Language with the multilan switcher. 

Setting it with the core language settings worked. 

Sadly php doesn't seem to understand Welsh though :(

MikeyGMT:
Sorted ... I had in the top of Welsh.php 

setlocale(LC_ALL, 'cym.UTF-8', 'cym.utf8', 'cym.utf8', 'cym');

This seems to work better. Although I am not sure what the different parameter are (or should be). 
setlocale(LC_ALL, 'cy_GB.utf8','cym.utf8', 'cym.utf8',  'cym');

MikeyGMT:
Retesting 
image

PR #2223, opened 2016-12-30
Text To LANS in FAQs #6 

MikeyGMT:
Text To LANS in FAQs #6 
Also split prefs into 3 tabs ... General, FAQs and Categories so there are less Lan Defines needed.
There is more LANS work TODO, but let's do that in another pull.

### DIFF
--- a/e107_plugins/faqs/faqs.php
+++ b/e107_plugins/faqs/faqs.php
@@ -470,7 +470,7 @@ class faq
 		{
 			$sc->setVars($rw);
 			$text .= $tp->parseTemplate($FAQ_LIST_LOOP, true);
-			$caption = "&nbsp;Category: <b>".$rw['faq_info_title']."</b>";
+			$caption = "&nbsp;".LAN_CATEGORY.": <b>".$rw['faq_info_title']."</b>";
 		}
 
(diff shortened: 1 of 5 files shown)


MikeyGMT:
Testing

image

Moc:
LAN_REMOVE may be added as a common term I think, or we can replace it with LAN_DELETE? Depending on the context. 

Great work with the LAN's Mike!

e107_plugins/faqs/faqs.php
```diff
@@ -333,7 +329,7 @@ function view_all_query($srch='')
 			$insert = " AND (f.faq_question LIKE '%".$srch."%' OR f.faq_answer LIKE '%".$srch."%' OR FIND_IN_SET ('".$srch."', f.faq_tags) ) ";
 
 
-			$message = "<span class='label label-lg label-info'>".$srch." <a class='e-tip' title='Remove' href='".$removeUrl."'>×</a></span>";
```
Moc: LAN_REMOVE may be added as a common term I think, or we can replace it with LAN_DELETE? Depending on the context.

MikeyGMT: Not sure where to access - cannot see a "remove" button in the FAQ anywhere - suspect may be legacy code ... can you give a clue?

PR #2256, opened 2017-01-07
Text to LANS #6 Welcome Messages

MikeyGMT:
Text to LANS #6 Welcome Messages


MikeyGMT:
Testing 
image

image

MikeyGMT:
Testing 
image

e107_languages/English/English.php
```diff
@@ -121,6 +121,7 @@
 define("LAN_IMAGE","Image");
 define("LAN_ABOUT", "About");
 define("LAN_TITLE", "Title");
+define("LAN_MESSAGE", "Message");
```
CaMer0n: Good move! I see a few plugins using "Message" in their lan files. ;-)

PR #2264, opened 2017-01-09
LANS Clean Up #6 - User Class

MikeyGMT:
LANS Clean Up #6 - User Class - more to do within conflicting defines
lan_user.php

### DIFF
--- a/e107_admin/userclass2.php
+++ b/e107_admin/userclass2.php
@@ -2,7 +2,7 @@
 /*
  * e107 website system
  *
- * Copyright (C) 2008-2015 e107 Inc (e107.org)
+ * Copyright (C) 2008-2017 e107 Inc (e107.org)
  * Released under the terms and conditions of the
  * GNU General Public License (http://www.gnu.org/licenses/gpl.txt)
  *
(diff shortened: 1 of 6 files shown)


MikeyGMT:
Testing 

image

image

image

e107_admin/userclass2.php
```diff
@@ -391,7 +391,7 @@ public function initialPage()
 				    <td>".$frm->select('init_class_stage', $initClassStages, $sel_stage)."<span class='field-help'>".UCSLAN_46."</span>
 
 				    </td></tr></table>
-				    <div class='buttons-bar'>".	$frm->admin_button('set_initial_classes','no-value','create',LAN_UPDATE)."</div>";
+				    <div class='buttons-bar' center>".	$frm->admin_button('set_initial_classes','no-value','create',LAN_UPDATE)."</div>";
```
CaMer0n: I believe the 'center' should be inside the single-quote?

e107_admin/userclass2.php
```diff
@@ -718,16 +718,16 @@ function checkAllowed($classID, $redirect = true, $fullEdit = FALSE)
 		//		unset($pref['frontpage'][$class_id]);		// (Should work with both 0.7 and 0.8 front page methods)
 		//		save_prefs();
 			}
-			$emessage->add(UCSLAN_3, E_MESSAGE_SUCCESS);
+			$emessage->add(UCSL/AN_3, E_MESSAGE_SUCCESS);
```
CaMer0n: What's happening here?

e107_admin/userclass2.php
MikeyGMT: Oops. I was thinking that when I added it. Silly me. Correct it tomorrow.

e107_admin/userclass2.php
MikeyGMT: Stops old lans in commented blocks from falsely appearing as missing must be renabled in deprecated LANs checker. My bad?

CaMer0n: Well UCSLAN_3 is "Class deleted." right? So, replace with LAN_DELETED should be enough I guess.

e107_admin/userclass2.php
MikeyGMT: It's this 
`<div class='buttons-bar center'>`
not this right?   
`<div class='buttons-bar' 'center'>`

Oops No 3

CaMer0n:
looks good @MikeyGMT !!

Issue #2278, opened 2017-01-14
Deprecated Lans - Minor Tweaks 


MikeyGMT:
Hi All, 
Does anyone object to me making a couple of minor tweaks to Deprecated Lans function? 

![image](https://cloud.githubusercontent.com/assets/2847651/21956483/c9bcb636-da79-11e6-9663-9c0083ec3c41.png)

Hence move the Info text to it's own row. Add tr's

```
		$text .= " 
<tr>
				<td><div class='alert-info alert alert-block'>".e107::getParser()->toHTML(LANG_LAN_140, true)."</div></td> 
</tr>
			   <tr>
				<td class='form-inline'>
					<select name='deprecatedLans[]' multiple style='height:200px'>
						<option value=''>".LANG_LAN_141."</option>";

```

Sort the file arrays, hence these new lines - which use asort. I'll indent appropriately on commit.

```
		$omit = array('languages','\.png','\.gif','handlers');
		
		$lans = $fl->get_files(e_ADMIN,'.php','standard',0);
asort($lans);
		
		$fl->setFileFilter(array("^e_"));
		
		$root = $fl->get_files(e_BASE,'.*?/?.*?\.php',$omit,0);
asort($root);
		
		//$handlers = $fl->get_files(e_CORE."handlers",'.*?/?.*?\.php',$omit,0);//mikey

		$templates = $fl->get_files(e_CORE."templates",'.*?/?.*?\.php',$omit,0);
asort($templates);		
		
		$shortcodes = $fl->get_files(e_CORE."shortcodes",'.*?/?.*?\.php',$omit,1);
asort($shortcode);

```		

Thirdly, I want to ask why all files are not searchable via the function? For example Lans are used in Plugins, Handlers etc, and as we need to check if a define is anywhere in the code, it would make sense to be able to search all the files.  

Cheers
Mikey

MikeyGMT:
And here 

```
		$selected = ($_POST['deprecatedLanFile'][0] == 'auto') ? "selected='selected'" :"";
		$text .= "<option value='auto' {$selected}>".LANG_LAN_142."</option><optgroup label='".LANG_LAN_143."'>\n";//Auto-Detect
asort($_SESSION['languageTools_lanFileList']);
		foreach($_SESSION['languageTools_lanFileList'] as $val)
```

MikeyGMT:
[...]

PR #2283, opened 2017-01-15
Deprecated Lans Sort Files in Order and Move Help To top Row #2278

MikeyGMT:
Deprecated Lans Sort Files in Order and Move Help To top Row #2278
Does not resolve missing handlers and for phrase in search all files.


CaMer0n:
You're probably the only person on the planet using this, so I'll go with the change, without testing it myself. :-)

PR #2267, opened 2017-01-10
Feature Request: Forum Link Icon/Image #1635

chory:
Lets show an image in the list of forums

PR #1635, opened 2016-05-09
 Feature Request: Forum Link Icon/Image

chory:
Show a image in forum title

PR #2303, opened 2017-01-20
Forum Link Icon/Image #1635 #2267

chory:
Lets show an image in the list of forums

Create new forumshorcodes, sc_forumimg($parm = '') and
sc_parentimg($parm = ''), with parameters

$parm:
w => width ( default=$tp->thumbWidth() )
h => height ( default=$tp->thumbHeight() )
crop => crop image ( default = 1)
fmt => format
   txt => Show Text (default)
   img => Show Image


chory:
Why fails? 😢 
else command in this code, i think is necesary. 😠

PR #2304, opened 2017-01-20
Use e107.behaviors handler to initialize click event on ".e-modal" el…

lonalore:
…ements.


CaMer0n:
Tested. It works! Thank you!

PR #2307, opened 2017-01-20
Patch for forum post cancel buttons

rica-carv:
Patch for forum post cancel buttons as reported by @Reconix


CaMer0n:
Far from ideal. I wonder, would we like to track user movement throughout the site? Each URL could be stored in a session and 'going back' would be rather easy if this was the case.

rica-carv:
@CaMer0n So how could you store the previous referer in a session, without using $_SERVER ????
I've searched high and low, and i didn't find a satisfatory replacement for using $_SERVER....
The only one that could really solve it would be use js and go mess with history...

CaMer0n:
Thanks..

Issue #125, opened 2013-02-27
Remove require_once() for known core handlers from all scripts. 


CaMer0n:
eg. one might replace:

```
 require_once(e_HANDLER."message_handler.php");
```

with: 

```
$mes = e107::getMessage();
```

(And test to make sure nothing breaks)

If we can get this done, @SecretR will be happy. :+1:

PR #2311, opened 2017-01-21
Replace require_once for userclass #125

MikeyGMT:
Replace require_once for userclass #125


CaMer0n:
I removed all the old code. Thanks.

PR #2317, opened 2017-01-22
Better render for download admin updates links

rica-carv:
Better render for download admin updates links
A download iconized text is far more easy to click than a single 11px x 12px icon....
untitled 1
1st line, proposal.
2nd line, as is...


CaMer0n:
This was rewritten and moved elsewhere.

PR #2326, opened 2017-01-26
Update lan_notify.php

yesszus:


yesszus:
@CaMer0n pls review this.

Moc:
Please give him a little more than 24 hours :)

yesszus:
Ok

Moc:
Idea, not sure: should we create a LAN_TEST in English.php for general use?

yesszus:
Ok, work this and changed.

CaMer0n:
lan_users.php has "Test [x]" .  It would be better to stick with that, and place it in lan_admin.php for use by notify.php. I don't see the need for "Test" on the front end.

yesszus:
Changed.

Issue #2302, opened 2017-01-20
High Memory Usage in Admin


MikeyGMT:
Hi Cameron, 
As our per discussion some days back, am still getting high memory issues on navigating to admin. This generates entries in the error log, like these.  

```
[Fri Jan 20 10:15:49 2017] [error] [client x.x.x.x] (12)Cannot allocate memory: couldn't create child process: /opt/suphp/sbin/suphp for /home/xxx/public_html/error.php, referer: http://xxxx.co.uk/e107_admin/admin.php
[Fri Jan 20 10:15:49 2017] [error] [client x.x.x.x] (12)Cannot allocate memory: couldn't create child process: /opt/suphp/sbin/suphp for /home/xxx/public_html/e107_admin/admin.php, referer: http://xxxx.co.uk/e107_admin/admin.php
```

It also looks related or a failure as a result of out of memory... the e107 Latest News Tabs for Core, Plugin and Theme, load inconsistently. Any one of three can fail to load, there does not seem to be a pattern. This seems to be a symptom rather than a real problem.  

I did quite a bit of research on it and my hunch is it could be related to the use of require_once / include_once throughout the system. #125 I gather that the new method is more efficient and unloads the module once used.

I have a memory trace here... https://docs.google.com/spreadsheets/d/1zHye-8Mfs_cCbnMkFQzm9d2eI3DBSXHbkMMkWCac0js/ 

I've tried various PHP settings, and spoken to the host. They recommended increasing the memory through cPanel > Choose PHP Version advanced settings (slightly improved). 

(I have been getting similar with thumb.php on the front end, but clearing the cache out & re-enabling and gzip site output and relinking the images to posts seemed to alleviate that somewhat).  

I've opened this so we can share information on it, and to get some guidance as to how to implement #125 in with specific handler examples. 

Thanks
Mikey

MikeyGMT:
Despite lots of debugging and analysis don't think it one particular thing in our code. It just seems to be that PHP 7 uses Memory differently (more) that earlier versions. I'll continue analysing and leave this open as it continues to be a problem, although I am not sure what can specifically be done... Search php 7 huge memory requirements ...  http://stackoverflow.com/questions/39740398/i-am-facing-more-memory-consumption-in-php-7-compare-to-php-5-6/39861791

MikeyGMT:
[...]

PR #2328, opened 2017-01-27
Cache Handler Performance Improvement #2302

MikeyGMT:
strpos is faster and takes less memory than substr with very large files
such as S_Theme_meta.cache.php which contains around 1600 lines, 48k
chars. I have replicated the same logic but using strpos. There is a
small gain in php 5.6 and slightly more in php7.0
Test results shown here.
https://docs.google.com/document/d/1o4X4ds377esQUSy8lE_RGN9BruKZ2aETxr6G1WTdiRc/

### DIFF
--- a/e107_handlers/cache_handler.php
+++ b/e107_handlers/cache_handler.php
@@ -150,7 +150,7 @@ class ecache {
 					}
 					elseif(strpos($ret,'<?php') === 0)
 					{
-						$ret = '<?php' ;			// Handle the history for now
+						$ret = substr($ret,5);			// Handle the history for now
 					}
 					return $ret;
 				}
--- a/e107_handlers/cache_handler.php
+++ b/e107_handlers/cache_handler.php
@@ -146,13 +146,13 @@ class ecache {
 				else
 				{
 					$ret = file_get_contents($cache_file);
-					if (substr($ret,0,strlen(self::CACHE_PREFIX)) == self::CACHE_PREFIX)
+					if (strpos($ret,self::CACHE_PREFIX) == 0 )  
 					{
-						$ret = substr($ret, strlen(self::CACHE_PREFIX));
+						$ret = self::CACHE_PREFIX;
 					}
-					elseif(substr($ret,0,5) == '<?php')
+					elseif(strpos($ret,'<?php') == 0)
 					{
-						$ret = substr($ret, 5);		// Handle the history for now
+						$ret ='<?php' ;			// Handle the history for now
 					}
 					return $ret;
 				}
--- a/e107_handlers/cache_handler.php
+++ b/e107_handlers/cache_handler.php
@@ -148,7 +148,7 @@ class ecache {
 					$ret = file_get_contents($cache_file);
 					if (strpos($ret,self::CACHE_PREFIX) == 0 )  
 					{
-						$ret = self::CACHE_PREFIX;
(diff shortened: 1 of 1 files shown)


MikeyGMT:
BEFORE   7.0.15
Render time: 0.517 cpu sec (33.00% load, 0.060 startup). Clock: 1.59 sec (24% of that for queries) DB queries: 211. Memory: 11.40MB/14.00MB

image
image
image





AFTER 7.0.15
Render time: 0.509 cpu sec (37.00% load, 0.064 startup). Clock: 1.37 sec (16% of that for queries) DB queries: 219. Memory: 11.36MB/14.00MB

image

image

CaMer0n:
Try with `strpos` using `=== 0` ?

MikeyGMT:
Hi @CaMer0n  
Yay that worked! 

Plus I checked all settings the same. Slighlty more text in welcome message on the BEFORE site. Other than that they are more or less identical - same plugins etc.  

PHP7 - Front - Before -> After
image
image

PHP7 - Admin - Before -> After
 
image
image

e107_handlers/cache_handler.php
```diff
@@ -146,13 +144,13 @@ public function retrieve($CacheTag, $MaximumAge = false, $ForcedCheck = false, $
 				else
 				{
 					$ret = file_get_contents($cache_file);
-					if (substr($ret,0,strlen(self::CACHE_PREFIX)) == self::CACHE_PREFIX)
+					if (strpos($ret,self::CACHE_PREFIX) === 0 )  
 					{
-						$ret = substr($ret, strlen(self::CACHE_PREFIX));
+						substr($ret, strlen(self::CACHE_PREFIX));
```
CaMer0n: We're missing `$ret =` here.

e107_handlers/cache_handler.php
```diff
@@ -146,13 +144,13 @@ public function retrieve($CacheTag, $MaximumAge = false, $ForcedCheck = false, $
...
-					if (substr($ret,0,strlen(self::CACHE_PREFIX)) == self::CACHE_PREFIX)
+					if (strpos($ret,self::CACHE_PREFIX) === 0 )  
 					{
 						$ret = substr($ret, strlen(self::CACHE_PREFIX));
 					}
-					elseif(substr($ret,0,5) == '<?php')
+					elseif(strpos($ret,'<?php') === 0)
 					{
-						$ret = substr($ret, 5);		// Handle the history for now
+						$ret = '<?php' ;			// Handle the history for now
```
CaMer0n: I didn't spot this before. This would create huge issues. What it is meant to do is return everything AFTER `<?php` 
ie. `substr($ret,5) `  (the value starting from the 5th character)

MikeyGMT: Corrected... 
Render time: 0.515 cpu sec (40.00% load, 0.058 startup). Clock: 1.30 sec (6% of that for queries) DB queries: 213. Memory: 11.69MB/14.00MB

image
image

CaMer0n: Maybe something went wrong. It doesn't look fixed here: 

					elseif(strpos($ret,'<?php') === 0)
					{
						$ret = '<?php' ;			// Handle the history for now
					}

MikeyGMT: Git didn't complete the Sync.

PR #2349, opened 2017-01-30
Update forum.css

mcpeace-maw:
I changed the colors to rgba to work with white and dark backgrounds themes. Also centered left column elements in viewtopic.


CaMer0n:
With any visual changes a screenshot would be helpful. Thanks! 👍

PR #2350, opened 2017-01-30
Update view_shortcodes.php

mcpeace-maw:
Removed bootstrap classes from 'usercombo' button group, in this way btn-group will have bootstrap original style, dropdown-toggle button will have a fixed size and will be more space for user mane. I hade a user that 'killed' me with this 'bug', his nick was from 18 digits.


CaMer0n:
Thanks. What's the before and after screenshots look like?

mcpeace-maw:
I'll send pics for forum changes on Gitter in a few minits.

PR #2358, opened 2017-01-31
Update admin_style.css

mcpeace-maw:
Fix for Menu Manage page: first menu will have a nice space on top all the time; the page will have scrollbar on right side if is needed. Please see images with before/after on gitter.


CaMer0n:
Thank you! 👍

PR #2375, opened 2017-02-03
Fix for Fatal Error during installation.

lonalore:


CaMer0n:
Sorry, didn't see this. It should be fixed now.

PR #2377, opened 2017-02-04
Replace require_once for userclass #125

MikeyGMT:
Replace require_once for userclass #125 - tested ok. added FIXME note
where it breaks


MikeyGMT:
Closed as agreed with @CaMer0n

Issue #2369, opened 2017-02-02
[Question] Menu admin UI design help


rica-carv:
Sorry for posting this here, but i think it's a better place than in e107help.org.

I'm currently designing some menus admin ui, and i'm having a hard time on figuring out how to correctly setup fields (ie, preferences?) and retrieve them on the xxx_menu.php file.

For instance, i have the $fields array defined inside e_menu.php, which is correctly displayed in the menu manager config area.
1st question: How can i define a default value for those $fields? In plugin.xml? How?
2nd: How can i access those $fields values from xxx_menu.php? I've looked at the various menus.php coded here, and simply can't figure it out...

Any help appreciatted....
Thanks...

CaMer0n:
This example (slightly modified from news_grid_menu.php) should answer both of your questions. 
    
    $gridSize  = vartrue($parm['layout'], 'col-md-4');

1) 'col-md-4' in this example
2) $parm in this example

rica-carv:
@CaMer0n Once more, thanks for pointing me on the right direction.
Anyway, i'm thinking of making some use of the plugin.xml defaults:
```
$plugin_pref = e107::pref('plugin');
$gridSize  = vartrue($parm['layout'], $plugin_pref['layout']);
```
Any thoughts about this approach?

rica-carv:
Err, might have called victory too soon...

I've got a small gremlin...
If the menu is rendered inside a regular menu area (ie: `{MENU=1}`), the `$parm` variable is correctly appended to the menu rendering, and therefore available for it's frontend code.
But if the menu is rendered outside of a regular menu area (ie: `{MENU=plugin\plugin_menu}`), there's no `$parm` available for frontend code....

rica-carv:
Forget it, i ended up writing my own fetch parm function. Thanks.

Moc:
Reopening to make sure this is documented somewhere

PR #2391, opened 2017-02-07
Possible patch for issue #2369

rica-carv:
Possible patch for issue #2369

### DIFF
--- a/e107_handlers/menu_class.php
+++ b/e107_handlers/menu_class.php
@@ -495,23 +495,6 @@ class e_menu
 			return $ret;
 		}
 	}
-
-private function menuarraysearch($array, $key, $value)
-{
-    $results = array();
-
-    if (is_array($array)) {
-        if (isset($array[$key]) && $array[$key] == $value) {
-            $results[] = $array;
-        }
-
-        foreach ($array as $subarray) {
-            $results = array_merge($results, $this->menuarraysearch($subarray, $key, $value));
-        }
-    }
-
-    return $results;
-}
 	
 	/**
 	 * Render menu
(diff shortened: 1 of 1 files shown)


CaMer0n:
Please try with my last commit. eg. 

    $parms = $tmp->getParams('news_months_menu', 2); // area 2

rica-carv:
@CaMer0n Your last commit works, but forces a plugin writer to know in which menuarea is the menu activated...
A bit odd, no?
Furthermore, even in a menuarea that only has a menu ([MENU=loginmenu}, it's needed to call the new getParams function, even inside the menu itself, so a menu coder needs to call it, in order to prevent if the menu could be activated in a single named menu area... Is this good practice?

My commit makes the params from the menu database directly available (if they're present, beware #2400...) to those menuareas during code run, without the need to call another function....

CaMer0n:
There's 2 different ways of rendering menus.. and you're mixing them, which isn't an intended feature.
You're choices are like this: 

1) Use Menu-Manager, assign your prefs in there (e_menu)  and assign the menu placement in there. 
OR
2) Hard-code your menu and assign your prefs in the shortcode. eg. {MENU: path=loginmenu&limit=3.etc.etc...}

It sounds like what you're trying to do is take the prefs from #1 and render them like #2 with the hardcoded prefs omitted. I don't encourage that.

rica-carv:
@CaMer0n So, since a hard-coded menu can't have admin ui managed parms, a theme designer will have to know which parms a menu needs and hardcode them in a single hardcoded area, right?
A admin will not have any control in those menu parms (unless it edits the theme itself)....
It's kind of odd, no?
A admin can control the parms of menus in menu areas via admin ui but cannot control the parms of a hardcoded menu via admin ui....
Isn't it possible for menu manager to handle also the hardcoded menus parms?

Somehow, i get your point, since hardcoded menus are theme related, but as i see it, it's parms are not, since they are somehow defined by default in menu config files inside the plugin folder...

rica-carv:

PR #2399, opened 2017-02-08
Update pm_shortcodes.php

pricey1981:
Where no parameter is passed in the shortcode $parm is received as an array. This resulted in 'inbox' being added to the delete url in the outbox where {PM_DELETE} is called.

### DIFF
--- a/e107_plugins/pm/pm_shortcodes.php
+++ b/e107_plugins/pm/pm_shortcodes.php
@@ -575,13 +575,13 @@ if(!class_exists('plugin_pm_pm_shortcodes'))
 			}
 
 
-			if($parm != '')
+			if(count($parm) == 0)
 			{
-				$extra = '.'.$parm;
+				$extra = '.'.($this->pmMode == 'outbox' ? 'outbox' : 'inbox');
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


e107_plugins/pm/pm_shortcodes.php
```diff
@@ -575,13 +575,13 @@ public function sc_pm_delete($parm = '')
 			}
 
 
-			if($parm != '')
+			if(count($parm) == 0)
```
CaMer0n: Thank you. Does `!empty($parm)` do the same job?

pricey1981: yes if $parm was an empty string count($parm) would equal zero. If it is an empty array, it equals zero. But $parm !='' doesn't work for an array, only a string

CaMer0n: Thanks. Would you be okay to update it? I'm trying to use 'empty' throughout the core where applicable.

rica-carv: @pricey1981 http://php.net/manual/en/function.empty.php

pricey1981:
sorry, misread your original response. Fixed it now

CaMer0n:
Thank you! 👍

PR #2401, opened 2017-02-09
Update request.php

kreativmind:
Removed line 150 alert-warning style=text-align:center'> because on the downloads page, it would show the code followed by the message No Records Found instead of just No Records Found


Moc:
@kreativmind Thanks for the help 👍 

The actual problem is not in the file you edited but it is here: e107_plugins/download/handlers/download_class.php (line 214). I'll push out a fix again. Closing this PR :)

PR #2403, opened 2017-02-09
Update e107_admin/plugin.php

yesszus:
Add missing lang codes.

### DIFF
--- a/e107_languages/English/admin/lan_admin.php
+++ b/e107_languages/English/admin/lan_admin.php
@@ -328,7 +328,7 @@ define("LAN_POSTED_BY_X","Posted by [x] [y]");
 define("LAN_VIEW_ALL","View All");
 
 define("LAN_BAN","Ban");
-
+define("LAN_PATH", "Path");
 
 define("LAN_UPLOAD", "Upload");
 define("LAN_UPLOADS","Uploads");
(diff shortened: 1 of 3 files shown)


e107_languages/English/admin/lan_plugin.php
```diff
@@ -306,7 +306,7 @@
 define("LAN_INSTALL_FAIL", "Installation failed!");
 define("LAN_UNINSTALL_FAIL", "Unable to uninstall!");
 define("LAN_PLUGIN_IS_USED", "[x] plugin is used by:");
-
+define("LAN_PATH", "Path");
```
CaMer0n: It should be in the non-universal format: EPL_ADLAN_XX 
The ONLY files which should contain the non-numeric format should be lan_admin.php and e107_languages/English/English.php . Please keep this in mind when adding new LANs. Thank you!

yesszus: Ok! Moved to lan_admin.php.

e107_languages/English/admin/lan_admin.php
```diff
@@ -328,7 +328,7 @@
 define("LAN_VIEW_ALL","View All");
 
 define("LAN_BAN","Ban");
-
+define("LAN_PATH", "Path");
```
CaMer0n: Unless 'path' is used often throughout the admin of e107, (I don't think it is) this should be in lan_plugin.php 
What do you think @MikeyGMT ?

MikeyGMT: It's defined more than once:- 
define("FMLAN_29", "Path");
define("IMALAN_110", "Path"); 

There's a few Path to [x] candidates too.
There may be some synonyms, e.g. directory, folder, location, url, which could be reviewed to use a consistent term.

CaMer0n: okay!! Thank you!! 👍

PR #2429, opened 2017-02-14
Fix for Menu Manager modal tooltips.

lonalore:


CaMer0n:
Thank you so much @lonalore !

PR #2450, opened 2017-02-20
Unneeded unset in form_handler.php

rica-carv:
Unneeded unset in form_handler.php, array is totally redefined line above


rica-carv:

Issue #2452, opened 2017-02-21
Wrong margin button on Verify Language-Packs


eduardocp:
The second button has a wrong margin-top

![image](https://cloud.githubusercontent.com/assets/1816859/23148527/19a44728-f7c5-11e6-8804-83ed2c61c030.png)

rica-carv:
They might have different top margins because firts button is a `<input>`, rendered under bootstrap btn class, and the second is a `<a>` tag also rendered as bootstrap btn class.
This is a bootstrap old related bug:
[Buttons and anchors with the btn class have different heights #1794](https://github.com/twbs/bootstrap/issues/1794)
[.btn on input vs button vs href elements result in different heights in FF #2985](https://github.com/twbs/bootstrap/issues/2985)

Issue #1794, opened 2016-08-06
Bootstrap3 - Page Content / Title Not floating down with header menu area as it changes in height.


MikeyGMT:
Hi Themers 

This looks great

![image](https://cloud.githubusercontent.com/assets/2847651/17457214/eeecb69a-5be9-11e6-95c6-62d8dad1c0f8.png)

But widen a touch ... 

![image](https://cloud.githubusercontent.com/assets/2847651/17457219/09d3f5e0-5bea-11e6-8d18-3193d9c57547.png)

Full width... 

![image](https://cloud.githubusercontent.com/assets/2847651/17457222/20100f42-5bea-11e6-967e-e8af206a342d.png)

I wonder how to ensure the menu/header zone does not overlay the content area?

Thanks

Mike

rica-carv:
As far i see it, it might be a bootstrap3 css coding issue, not a e107 directly related one....

MikeyGMT:
Does it on several sites - upgraded and fresh install - from today's git repo. 
Shortening the sitename pref seems to prevent it from happening. It would be nice if the content started below the menu/header. Looks like content starts at a fixed point, and the menu spills over.

rica-carv:
Tested, definetly bootstrap3 css issue, since this css rule on bootstrap3 navbar.less fixes navbar class, overlapping content:
```
.navbar-fixed-bottom, .navbar-fixed-top {
    position: fixed;
    right: 0;
    left: 0;
    z-index: 1030;
```

[Solutions here](https://www.google.pt/search?q=boostrap+fixed+navbar+overlapping+conten&ie=utf-8&oe=utf-8&client=firefox-b-ab&gfe_rd=cr&ei=9R2mV777IZLS8AfYnJ3oCg#q=bootstrap+fixed+navbar+overlapping+content)

CaMer0n:
@MikeyGMT You'll need to adjust your theme/css to accommodate the long site title.

CaMer0n:
Closing

MikeyGMT:
Quick update 
If the Jumbotron gets too big then overlays the content. 
This seems to be the cause 
`body { padding-top: 48px;  `

This seems to do the trick 
`body { padding-top: 4.5pc; `

It's not perfect, but at least the content does not get overlayed.

MikeyGMT:
It's line 3 of bootstrap3/style.css actually it was 50px not 48px  (viewing in chrome)

[...]

PR #2453, opened 2017-02-21
Update e107_admin/lancheck.php

eduardocp:
Correction of issue #2452


rica-carv:
This is a bootstrap related issue, as i comented on the issue itself.
I wonder if this couldn't be solved via css, instead of hardcoding css styles inside html....

eduardocp:
@rica-carv I do not see this as a bootstrap bug, just a lack of pattern.
We can change the a tag to input and so we standardize the buttons

rica-carv:
@eduardocp As stated by me in #2452:

> They might have different top margins because firts button is a `<input>`, rendered under bootstrap btn class, and the second is a `<a>` tag also rendered as bootstrap btn class.
This is a bootstrap old related bug:
[Buttons and anchors with the btn class have different heights #1794](https://github.com/twbs/bootstrap/issues/1794)
[.btn on input vs button vs href elements result in different heights in FF #2985](https://github.com/twbs/bootstrap/issues/2985)

Probably changing the `<a>` to `<input>` will make the problem persist, but that's a option.
The real go would be to render both buttons with the same HTML element...

PR #2455, opened 2017-02-21
Allow admin ui boolean disabled radio field

rica-carv:
Allow admin ui boolean disabled radio field.

Purpose - Show user why field is not enabled, since it depends on a uninstalled plugin:
untitled 1

writeParms-> disabled is used for enable or disable it (as it was already trying to do it on part of the code, but since this field was used for custom fields name, it wasn't really used, since it was unset before the code run on radio rendering...)
writeParms-> label is now used for the customization of both fields, as text (example: `"Enabled&Disabled"`)
For the label, alternative would be to use `optArray `[(like is currnetly used for radio buttons)](https://github.com/e107inc/e107/blob/79c5bff6741348dafccc168d5b2b162d771c1fc4/e107_plugins/newsfeed/admin_config.php#L92) instead of `label`...

_P.S.: Alternative for this is to render the "not installed" warning instead of the radio buttons, ~~but since there's no warning or alert support in form handler, it requires heavy customization...~~
Something like this:
untitled 1_

_P.S.S.: Another alternative is disable them using javascript/jquery, but it won't work if admin disables javascript...._


CaMer0n:
Sorry, I don't think we can use this code with the changes that were made for bootstrap-switch.

rica-carv:
No problem, i got a workaround...
Thanks anyway...

rica-carv:
Anyway, bootstrap-switch should handle itself disabled...
My changes were made before the code change to bootstrap-switch...
By the way, what if a site admin doesn't like the bootstrap-switch lookalike for boolean fields? It has to be forced to use it?

CaMer0n:
It is not forced on the frontend, only the backend.

rica-carv:
Indeed, but anyway, it's forced...
If i got a non bootstrap admin theme, or if i don't wan tto use bootstrap on the backend, i'll be always forced to have those switchs.....

Moc:
At this point you cannot change the admin theme I think. And I don't think there are plans to support different admin themes..

Trying to make all frontend themes work provides plenty of work, having the admin standardised is therefore better to avoid having to spend precious time on that as well.

rica-carv:
According to [this](https://github.com/Bttstrp/bootstrap-switch/issues/494), a simple add of disabled='disabled' to the original input would solve this question....
It's almost what i was trying to do on the first pull request...

CaMer0n:
Did you try `writeParms=>array('disabled'=>true)` ?

rica-carv:
@CaMer0n Yes, i did. `writeParms=>array('disabled'=>true)` messes the switch:
untitled 1


 My original pull request was addressing the "bug" on the disabled parm being sent to a boolean, since it will use it as definition lan: 
```
$lenabled = vartrue($parms['enabled'], 'LAN_ON');
$ldisabled = vartrue($parms['disabled'], 'LAN_OFF');
```

As far i see it, **DISABLED** should be reserved for disabling forms controls.....

rica-carv:
The issue is that when the field type is boolean, it will call radio-switch, and puf, no disabled option there.
But the radio does handle disabled, as it is used here: https://github.com/e107inc/e107/blob/e1932a9f1a20a1f69823f829d3b6a7106407ec02/e107_admin/prefs.php#L950

Why the existance of both? Don't they render the same thing?

CaMer0n:
@rica-carv Not perfect, but does the commit fix the issue if you use `writeParms=>array('disabled'=>true)` ?

rica-carv:
@CaMer0n No, it still has the issue. Tested and retested again.
The bug is inside the boolean render function itself, as i changed in my commit, so this will still render the switch as enabled, and with a "1" instead of "off", as pictured above....

PR #2468, opened 2017-03-01
changed path to default video background

Jimmi08:
original video was removed


CaMer0n:
I would prefer a CDN. Otherwise it will chew-up a lot of bandwidth on e107.org

Jimmi08:
OK, but original url is removed.

CaMer0n:
Can you find another URL for it?

Issue #2292, opened 2017-01-17
active class with NAVIGATION shortcode for homepage


Jimmi08:
Maybe I have something wrong set

http:://website.sk -  no active class on li tag 
http:://website.sk/index.php - active class is on li tag with index.php in url

It's default installation. Maybe URL configuration or something in sitelinks.

Thanks

Jimmi08:
Solved.

PR #2490, opened 2017-03-04
FIX for #2292 active class for homepage without index.php

Jimmi08:
This is workround for setting homepage url with index.php. Correct way should be just /
Now you have double content.


CaMer0n:
Thank you!

PR #2497, opened 2017-03-06
Ability to use bs-switch in menu-manager config modals.

lonalore:


CaMer0n:
Excellent! Thank you!

PR #2500, opened 2017-03-08
Update lancheck.php

yesszus:
Removed edit button  width style.


yesszus:
Work this plugins and themes tabs.

PR #2551, opened 2017-04-03
PM Plugin templates to V2.x locations

rica-carv:
PM Plugin templates to V2.x locations
Also changed some "hardcoded" HTML from menu.php to template.php...


rica-carv:
Templates fully updated to V2 standards, enabling use of core function gettemplate....

e107_plugins/pm/pm.php
```diff
@@ -373,17 +388,21 @@ function show_pm($pmid, $comeFrom = '')
...
 	
-		$PM_BLOCKED_HEADER = $this->updateTemplate($PM_BLOCKED_HEADER);
-		$PM_BLOCKED_TABLE  = $this->updateTemplate($PM_BLOCKED_TABLE);
-		$PM_BLOCKED_EMPTY  = $this->updateTemplate($PM_BLOCKED_EMPTY);
-		$PM_BLOCKED_FOOTER  = $this->updateTemplate($PM_BLOCKED_FOOTER);
+// Is updateTemplate really necessary for v2.x templates?
+		$PM_BLOCKED_HEADER = $this->updateTemplate($PM_BLOCKED['start']);
+		$PM_BLOCKED_TABLE  = $this->updateTemplate($PM_BLOCKED['item']);
+		$PM_BLOCKED_EMPTY  = $this->updateTemplate($PM_BLOCKED['empty']);
+		$PM_BLOCKED_FOOTER  = $this->updateTemplate($PM_BLOCKED['end']);
```
CaMer0n: What you've done is excellent for v2! Unfortunately, it looks like it will ignore a v1 template that is in use, but maybe that is not a big deal - best ask the theme developers how often they themed the PM. ;-)

rica-carv: I thought of that when doing the changes, but how can you deal with a V1 template when the plugin itself is upgraded to V2 standards?
V1.x theme templates were in the theme root, V2.x theme templates are on the templates/plugin folder, and therefore, if the plugin uses the V2.x gettemplate core function, it won't simply pick up the legacy V1.x template, and will pick up it's own plugin "default" template.
So, for me, as i see it, moving the old legacy template to the v2.x theme template location only will add confusion, as the template won't be a real V2.x template on a V2.x theme template location, so it won't work neither...
Theme designers should upgrade their templates to V2.x standards, as old V1.x custom plugins templates simply won't work on upgraded plugins to V2.x standards....
Do i see it wrong, or not?

We can always do a theme templates conversion on the fly function, to convert v1.x theme templates to V2.x standars, but it would be a waste of time, as i see it.
The use of V1.x templates should be discouraged, or at least, V1.x templates with custom plugins templates on it. It should be a visible warning for the admin, about V1.x plugins templates present in the current theme that the admin is trying to use, in order to prevent questions like: "why my v1.x theme custom template is not working?". The warning should also advert the admin that those theme legacy plugnin templates should be upgraded to V2.x standards...

P.S.: Sorry for the long text, i hope i explained my point of view...

CaMer0n: You use the same variables `$PM_WHATEVER` and make them equal to the new $template array when` (THEME_LEGACY !== true)`
I have done this for quite a lot of plugins in the core already.

rica-carv: I know, i just saw it on forum plugin.
But THEME_LEGACY is dependen of the existance of theme.xml.
Imagine you got a old theme template "injected" in the templates folder of a v2.x theme.
What happens? Caos....

Anyway, if the problems is the handling of legacy pm plugin templates, i can add the check, but it will lead to the replacement of gettemplate from the plugin, [since the gettemplate only checks for the templates in 2 places](https://github.com/e107inc/e107/blob/ca1fd4c40c8c34500f9655833f9dd7107a0c5d15/e107_handlers/e107_class.php#L2682):
Example usage: <code>e107::getTemplate('user', 'short_start');</code>
	 * - e107_themes/current_frontend_theme/templates/user_template.php (if $override is true)
	 * - e107_themes/templates/user_template.php (if override not found or $override is false)

So, what's the point on full upgrading to V2.x? If it's to deal with legacy templates, might better leave it this way...

As i stated before, if a plugin is V2, it should be FULLY V2, not just parts of the code.
If there's core code that can't handle legacy, that's another issue....

rica-carv:
Added legacy theme templates load, if present

CaMer0n:
Great stuff @rica-carv !

PR #2553, opened 2017-04-08
fix these fields have opposite meanings - hide vs view

Jimmi08:


CaMer0n:
Excellent! Thank you!

Issue #2539, opened 2017-03-30
Social plugin issues


yesszus:
Not work for me plugin settings and Facebook, Github and Twitter login / registration.

See this [video](https://www.youtube.com/watch?v=aW82ynEVdcc).

My e107 version: 2.1.6

Moc:
Try clearing all your caches (e107 and browser). Does this help?

Jimmi08:
I know that this is something different, just to have all related issues in one list.
https://github.com/e107inc/e107/issues/2127

yesszus:
@Moc Cleared all e107 and browser caches.
No changes were made.

yesszus:
@Moc @Jimmi08 pls test this commits.
Work for me.

yesszus:
Ok, now work for me plugin settings.
Not work for me Facebook, Github and Twitter login / registration.

CaMer0n:
@yesszus hybridAuth has just been upgraded. Please test again.

yesszus:
Facebook login / registration now work for me.

[5]Authentication failed! Twitter returned an error.
Github: [5]Error SQL Insert Error

Would anyone test this?

tetolk:
I will test the Twitter part today.

yesszus:
And work?

Jimmi08:
With LinkedIn I got error (but I have ID and secret key inserted):
 
![image](https://user-images.githubusercontent.com/5429548/30581586-f8ce3412-9d20-11e7-9348-3ca894d0232b.png)

With Facebook login and signup I got notification  Event Triggered: user_xup_login as main admin (for both, login and signup)
Event Triggered: user_xup_login
![image](https://user-images.githubusercontent.com/5429548/30581654-2a0d28bc-9d21-11e7-8476-df861eec66c8.png)

Nothing happend.

Jimmi08:
Ok, maybe reason for LinkedIn:

there is this check in Provider_Model_OAuth2.php:
```
    if (!$this->config["keys"]["id"] || !$this->config["keys"]["secret"]) {
      throw new Exception("Your application id and secret are required in order to connect to {$this->providerId}.", 4);
    }
```
but values for $this->config["keys"] are:
![image](https://user-images.githubusercontent.com/5429548/30588930-eb944d0a-9d38-11e7-92a6-8991159f9676.png)

$this->_config() in user_handler.php
[...]

PR #2594, opened 2017-04-22
Update  e107_plugins/social

yesszus:
Social config buttons changer works again.

### DIFF
--- a/e107_core/controllers/system/xup.php
+++ b/e107_core/controllers/system/xup.php
@@ -14,6 +14,7 @@
  *    $Author: e107coders $
 */
 
+e107::lan('social', false, true);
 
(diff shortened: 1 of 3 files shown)


yesszus:
@CaMer0n Still not work for me Facebook, Github and Twitter login / registration.
Issue: #2539

e107_core/controllers/system/xup.php
```diff
@@ -14,6 +14,7 @@
  *    $Author: e107coders $
 */
 
+e107::lan('social', false, true);
```
CaMer0n: You're loading a plugin language file for a system script. `e107_core/controllers/system/xup.php` 
**lan_user.php** would be more appropriate. (see line 120+)

PR #2595, opened 2017-04-23
Pm_menu template new bug

rica-carv:
New bug preventing loading of template. Corrected


e107_plugins/pm/pm_menu.php
```diff
@@ -102,7 +102,7 @@ function pm_show_popup($pm_inbox, $alertdelay = 0)
 $template = e107::getTemplate('pm', 'pm_menu');
 
 //if(!isset($pm_menu_template))
-if(!isset($PM_MENU_TEMPLATE))
+if(!isset($template))
```
CaMer0n: Looks good, but do you also include the template the old way for legacy themes?

rica-carv: pm_menu_template is new, before v2.1.5 theres was no such template, so no chance for legacy themes to use that template...

PR #2597, opened 2017-04-24
Update lan_user.php

yesszus:


e107_core/controllers/system/xup.php
```diff
@@ -14,6 +14,7 @@
  *    $Author: e107coders $
 */
 
+e107::coreLan('user');
```
CaMer0n: Does it function without this line?

yesszus: Not work without this line.

yesszus:
@CaMer0n  not good this?

PR #2602, opened 2017-04-26
Menu manager left col: Fix if theme have many layuts

veskoto:
Fix if theme have many layuts


CaMer0n:
Thank you @veskoto !! Nice fix! 👍

PR #2608, opened 2017-04-30
Update lan_admin.php

yesszus:


yesszus:
@CaMer0n  pls review this.

e107_admin/admin.php
```diff
@@ -535,7 +535,7 @@ private function checkPasswordEncryption()
 
 		if($us->passwordAPIExists() === true && $us->getDefaultHashType() !== PASSWORD_E107_PHP && e107::pref('core','password_CHAP')==0)
 		{
-			$message = "It is HIGHLY recommended that you [change your password encoding] to the PHP Default. (Password hashes will be automatically upgraded during user login.)";
+			$message = e107::getParser()->toHTML(LAN_UI_PASSWORD_WARNING, true);
```
CaMer0n: This is not part of the admin-ui, so LAN_UI is not an appropriate prefix. 
Also, it does not contain regular bbcodes such as `[b]`,  so running it through the parser is not needed.

yesszus: Ok @CaMer0n changed this line and removed regular bbcode.

CaMer0n:
Thank you! 👍

PR #2609, opened 2017-04-30
Fix for admin dashboard.

lonalore:
After upgrading to v2.1.6 my dashboard flexpanel settings has been gone. Furthermore the new dashboard style was quite weird:

**Before:**
screencapture-e107projects-e107_admin-admin-php-1493572145217 1

**After my fix:**
screencapture-e107projects-e107_admin-admin-php-1493572636300

### DIFF
--- a/e107_themes/bootstrap3/admin_style.css
+++ b/e107_themes/bootstrap3/admin_style.css
@@ -1565,11 +1565,9 @@ body#admin-menus .admin-left-panel,
 
 }
 
-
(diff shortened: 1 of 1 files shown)


lonalore:
And I think we should remove ".row-flex" class from the row if its empty, because now it has a height due to the margin-bottom. Any idea how to fix this?

kepernyokep_2017-04-30_19-49-55

CaMer0n:
Thank you! 👍

PR #2619, opened 2017-05-04
Update pm_menu_template.php

yesszus:
Replaced lang codes, messages count and add line break to send new privat message button.


rica-carv:
I just kept the old phrase.
Now it can be customized anyway, inside the theme itself...

PR #2625, opened 2017-05-08
New admin ui form type: header

rica-carv:
New admin ui form type: header

Code:
`          '_header'  => array('title'=> LAN_EUSER_ADMIN_515,'tab' => 1, 'type'=>'header')`

Renders like this:
untitled 1


rica-carv:
Nevermind, solved with a custom method....

CaMer0n:
And it's not a 'field', so it's good you found a work-around.

rica-carv:
Tabs are also not fields...
But anyway, ahead...

PR #2627, opened 2017-05-10
possibility to style with css classes login button

Jimmi08:
Yes, I know about $frm idea, this is just shortcut.

### DIFF
--- a/e107_core/shortcodes/batch/login_shortcodes.php
+++ b/e107_core/shortcodes/batch/login_shortcodes.php
@@ -151,15 +151,19 @@ class login_shortcodes extends e_shortcode
 
 	}
 	
+	/* example {LOGIN_TABLE_SUBMIT} */
+	/* example {LOGIN_TABLE_SUBMIT=large}  old way */  
+	/* example {LOGIN_TABLE_SUBMIT: class=btn btn-primary btn-large btn-lg pull-right} */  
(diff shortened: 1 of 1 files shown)


e107_core/shortcodes/batch/login_shortcodes.php
```diff
@@ -151,15 +151,19 @@ function sc_login_table_rememberme($parm=null)
...
-		if(empty($this->userReg))
+    $class = ($parm == 'large') ? "btn-large btn-lg" : "";
+    if(empty($this->pref['user_reg']))
 		{
 			return null;
 		}
-
-		$class = ($parm == 'large') ? "btn-large btn-lg" : "";
-		return "<input class='btn btn-primary ".$class." button' type='submit' name='userlogin' value=\"".LAN_LOGIN_9."\" />";
+
```
CaMer0n: Mixing `$parm` as a string and `$parm` as an array will get messy. (especially with PHP 7.1+)  It may be better to create a new shortcode which uses `$parm` only as an array. . ie. **sc_login_table_button($parm=null)**

Jimmi08: I would prefer to remove that old way totally, not new shorcode. So no large parm.

CaMer0n: Yes, but it would break backward-compatibility. A new shortcode avoids that issue.

PR #2629, opened 2017-05-11
Update bootstrsp switch CDN to 3.3.4

rica-carv:
Last stable is v. 3.3.4 (https://github.com/Bttstrp/bootstrap-switch), also is available via CDN (https://cdnjs.com/libraries/bootstrap-switch)

Also might be worth updating local files?

_NOTE: cdn.jsdelivr.net only has v3.3.3 & 4.0.0 alpha.1_

### DIFF
--- a/e107_handlers/library_manager.php
+++ b/e107_handlers/library_manager.php
@@ -498,7 +498,7 @@ class core_library
 			),
 			// Override library path to CDN.
 			'library_path'      => 'https://cdn.jsdelivr.net/bootstrap.switch',
-			'path'              => '3.3.2',
+			'path'              => '3.3.4',
 		);
 
 		// Bootstrap Switch (local).


jimaek:
You can use this URL to get auto-updates directly from npm https://cdn.jsdelivr.net/npm/bootstrap-switch@3

More info here https://github.com/jsdelivr/jsdelivr#usage

rica-carv:
@jimaek That would need to have some code rewriting, so i wonder if it's a option. @CaMer0n ?
Anyway, that url would only be valid for 3 versions, when 4 version pops up, code would need to be rewritten...
Ideially should be a "auto-update" url for last version, if possible...

jimaek:
If you dont specify a version you will get the latest https://cdn.jsdelivr.net/npm/bootstrap-switch
But its not recommended to use in production, unless you are 100% sure a newer version of bootstrap-switch wont break your code

rica-carv:
@jimaek Yes, i read that also, so there's no "reliable" auto-update solution...

jimaek:
Version aliasing is the most reliable auto-update. Other than that, manually changing the URL is the only 100% reliable system. Depends on what is your worry

MartinKolarik:
@rica-carv the "latest" links are, of course, as reliable as any jsDelivr links - you'll always get the correct file. The problem here is that an auto-update solution can never be reliable, because there are breaking changes between major versions that might require changes in your code.

rica-carv:
@MartinKolarik So auto-update will never be an option, since code has to be adapted to newer versions, if  it's somehow required...

PR #2637, opened 2017-05-15
Update Icon Selector

drasehat:
make icon selector not disturb other field
when i try to create chapter, first row is icon selector and second is dropdown for new or existed book.
i try click the second field (dropdown) it always open media manager for icon selector. this is why :
icon-selector

with this patch, this is the result for the icon selector field
icon-result


CaMer0n:
Thank you! 👍

drasehat:
you are welcome, but it look strange .. i never seen <div> inside <a> tag 😄

PR #2646, opened 2017-05-22
[html] - Fix end tag mix-up in forum_viewforum_template 'item' block.

arunshekher:
Dear Admins,
This mixup caused unclosed div element in the generated markup. Please review and merge.

**FF view-source:**

image


CaMer0n:
Thank you! 👍

PR #2647, opened 2017-05-22
[html] - Fix forum search button start tag using self closing syntax

arunshekher:
Should fix 'Self-closing syntax (/>) used on a non-void HTML element' html validation error on the search button.

**FF view-source:**
image


CaMer0n:
Great! 👍

PR #2648, opened 2017-05-23
Fix invalid html for 'Delete' and 'Move' list items in forum Admin Options dropdown

arunshekher:
Dear Admins,
This commit cleanup unclosed 'li' tag, unclosed 'a' tag, remove stray 'i' end tags in 'Delete' and 'Move' list items in the forum 'Admin Options' drop down menu. 

Please review and merge if appropriate.

**FF view-source:**
image


CaMer0n:
Looks good! Thank you!

PR #2652, opened 2017-05-25
New pm shortcode PM_BREADCRUMB

rica-carv:
Now pm breadcrumb is hardcoded and forced on every pm templates.

With this new pm shortcode {PM_BREADCRUMB}, it allows full customization of pm_template, with or without breadcrumb...

### DIFF
--- a/e107_plugins/pm/pm.php
+++ b/e107_plugins/pm/pm.php
@@ -405,7 +405,7 @@
 
 //		$PM_SHOW = $this->updateTemplate($PM_SHOW);
 
-			$txt = e107::getParser()->parseTemplate($PM_SHOW, true, $sc);
+//			$txt = e107::getParser()->parseTemplate($PM_SHOW, true, $sc);
 
 			if($comeFrom == 'outbox')
 			{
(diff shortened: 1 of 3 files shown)


CaMer0n:
Thanks. I'll come back to this after 2.1.6 is released.

e107_plugins/pm/pm.php
```diff
@@ -882,7 +892,10 @@ function pm_user_lookup()
 	switch($action)
 	{
 		case 'send' :
-			$ns->tablerender(LAN_PM, $pm->breadcrumb(LAN_PLUGIN_PM_NEW) . $mes->render() . $pm->show_send($pm_proc_id));
+//--			$ns->tablerender(LAN_PM, $pm->breadcrumb(LAN_PLUGIN_PM_NEW) . $mes->render() . $pm->show_send($pm_proc_id));
+ 			$sc->bc_type = LAN_PM_55;
+   		$sc->bc_other = NULL;
+			$ns->tablerender(LAN_PM, $mes->render() . $pm->show_send($pm_proc_id));
 			break;
```
CaMer0n: Shouldn't $sc->bc_type = LAN_PLUGIN_PM_NEW here?

CaMer0n: @rica-carv I'd still like to get this merged. What are your thoughts on the comment above?

Moc: @rica-carv any thoughts on Cameron's comment?

rica-carv:

e107_plugins/pm/pm.php
rica-carv: @CaMer0n @Moc Sorry, i only see now your questions... @CaMer0n You're right, it should be PM_NEW...

PR #2675, opened 2017-06-03
oldpolls.php comments container html fix

arunshekher:
Dear Admins,
Please review and merge if appropriate.

Fixes 'Element li not allowed as child of element div in this context.' html validation error in oldpolls.php

### DIFF
--- a/e107_plugins/poll/oldpolls.php
+++ b/e107_plugins/poll/oldpolls.php
@@ -88,13 +88,13 @@ if(e_QUERY)
 
 		if ($comment_total = $sql->gen($query))
 		{
-			$text .= "<hr /><div>";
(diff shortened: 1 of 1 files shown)


CaMer0n:
I believe the whole code, include the while loop should be replaceable by another comment-class method.

arunshekher:
@CaMer0n 
Is it so:

```php
$ns->tablerender(LAN_PLUGIN_POLL_NAME." #".$row['poll_id'], $text);
echo "<hr />";
e107::getSingleton('comment')->compose_comment('poll', 'comment', intval( $row['poll_id'] ), null, 'Poll Comment');              
```

CaMer0n:
Exactly! Thank you! 👍

PR #2679, opened 2017-06-05
Update lan_image.php

yesszus:
Add phrases.

### DIFF
--- a/e107_admin/admin.php
+++ b/e107_admin/admin.php
@@ -229,7 +229,8 @@ class admin_start
 				}
 				else
 				{
-					$mes->addWarning("Unable to create <b>".$dr."</b>. Please check your folder permissions.");
+					$message = e107::getParser()->lanVars(e107::getParser()->toHTML(ADLAN_187, true), array('x'=>$dr));
+					$mes->addWarning($message);
 				}
 			}
 		}
@@ -245,7 +246,8 @@ class admin_start
 
 		if(e107::getDate()->isValidTimezone($timezone) == false)
 		{
-			$mes->addWarning("Your timezone setting (".$timezone.") is invalid. It has been reset to UTC. To Modify, please go to Admin -> Preferences -> Date Display Options.", 'default', true);
+			$message = e107::getParser()->lanVars(ADLAN_188, array('x'=>$timezone));
+			$mes->addWarning($message, 'default', true);
 			e107::getConfig()->set('timezone','UTC')->save(false,true,false);
 			$this->refresh = true;
 		}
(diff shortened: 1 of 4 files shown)


e107_admin/admin.php
```diff
@@ -229,7 +229,8 @@ private function checkPaths()
 				}
 				else
 				{
-					$mes->addWarning("Unable to create <b>".$dr."</b>. Please check your folder permissions.");
+					$message = e107::getParser()->lanVars(e107::getParser()->toHTML(ADLAN_187, true), array('x'=>$dr));
```
CaMer0n: Instead of using `toHtml()` use `lanVars(LAN,$dr,true);` to make the variable bold. There is no need to use `array('x'=>$dr)` when only one variable of [x] is in the LAN.

e107_admin/admin.php
```diff
@@ -245,7 +246,8 @@ private function checkTimezone()
 
 		if(e107::getDate()->isValidTimezone($timezone) == false)
 		{
-			$mes->addWarning("Your timezone setting (".$timezone.") is invalid. It has been reset to UTC. To Modify, please go to Admin -> Preferences -> Date Display Options.", 'default', true);
+			$message = e107::getParser()->lanVars(ADLAN_188, array('x'=>$timezone));
```
CaMer0n: As mentioned above: `$message = e107::getParser()->lanVars(ADLAN_188, $timezone);` should do the same thing.

e107_admin/admin.php
```diff
@@ -522,7 +524,8 @@ private function checkIncompatiblePlugins()
 		if($inCompatText)
 		{
 			$text = "<ul>".$inCompatText."</ul>";
-			$mes->addWarning("The following plugins are not compatible with this version of e107 and should be uninstalled: ".$text."<a class='btn btn-default' href='".e_ADMIN."plugin.php'>uninstall</a>");
+			$message = e107::getParser()->lanVars(ADLAN_189, array('x'=>$text, 'y'=>"<a class='btn btn-default' href='".e_ADMIN."plugin.php'>uninstall</a>"));
```
CaMer0n: No need for using [x] or [y] . 
Just need 2 lans..  `LAN1 . $text . "<a href..... >".LAN2."</a>";`

e107_admin/image.php
```diff
@@ -2545,7 +2545,8 @@ function batchImportForm()
 		{
 			if(empty($f))
 			{
-				e107::getMessage()->addWarning("0 byte file found in: ".e_IMPORT."<br />Please remove before proceeding.");
+				$message = e107::getParser()->lanVars(IMALAN_180, array('x'=>e_IMPORT));
+				e107::getMessage()->addWarning($tp->toHtml($message, true));
```
CaMer0n: Use 2 LANs.  eg. `LAN1."<br />".LAN2`

yesszus:
@CaMer0n updated all.

CaMer0n:
Much better. Thank you! 👍

Issue #2659, opened 2017-05-28
Why is there many missing language defines in Poll plugin?


arunshekher:
In old polls page and poll for forum etc. 

Is there any particular reason for this? I think it has been like this for about 2-3 months now.

Screenshots:
![image](https://cloud.githubusercontent.com/assets/315195/26532255/fe3eaa7a-440c-11e7-9d1c-cbb2a98309fd.png)

![image](https://cloud.githubusercontent.com/assets/315195/26532269/44988b08-440d-11e7-81e9-d526c37ca5d8.png)

tgtje:
very quick lookup :   e107::includeLan(e_LANGUAGEDIR.e_LANGUAGE.'/admin/lan_admin.php');  should be added to poll_class php  (please check if so) else commit by... ?   (some defines used are located in admin/lan_admin.. not in poll or forum)

arunshekher:
@tgtje Thank you for your reply so if I understand you correctly those defines are waiting to be moved to poll plugin language files? 

Had gone through some of the discussions here about the LAN in the past and call to remove duplicated LAN defines. The best guess I could make about it is that these missing defines where commented out in such effort? right?

'In poll plugin' was used in the sense of 'effector' thought that was obvious.

tgtje:
a) no, they are **not** to be moved to poll.... they are in use (some at least) sidewide. The method is adding THAT line to poll class; so whenever poll class is running it 'loads' the defines from within the system languages.
b) **no** likely they are not commented out or deleted, but have gotten a new (more general use throughout the system) LAN. The goal is to rename LAN's that are the same use 1 (decrease lan numbers) all over, (as long as the 'meaning/words' are similar for understanding example....LAN_SAVE (meaning : Save) is pretty obvious ? so wherever in the system the WORD Save is hardcoded, you can/could (if possible) replace that by using LAN_SAVE...   (it may happen that some older lan's stay intact...legacy reason when upgrading). **_So i do believe that the original commit has  'missed' that piece and guessing was tested but not at forum related location It happens, we are human.._**.That is all in a nutshell.
Note, _commit should be easy, but do not know if it is enough_, hencee, quick repl; i (when commiting)  try to test it as much as possible).

tgtje:
**Posted a commit** ( some research showed the new LAN .. multpl.  was moved to admin/lan_admin so became more general (was previous in poll/English as POLLAN_9 (poll use) and LAN_FORUM_3033 (forum use).They are commented out (old ones).

[...]

PR #2683, opened 2017-06-06
Update English.php

tgtje:
Added used coded LAN's :  fixes (see #6 #2659) These two defines are added but not implemented into a lot of other files. 
These can be used (implemented) whenever a text to lan defines change takes place. The words (singular) 'from' and 'to' are represented through these two LAN's. Replacing all in 1 go will deliver too much stressload on multiple files... 
Do NOT merge if not ok using these two.


CaMer0n:
Thanks @tgtje . I think it's better to use "From [x] to [y]" instead of separate LANs which could end up in the wrong order with RTL languages.

tgtje:
I know such would be the case.. but it is more in the context of a time span... date....from... date...to
 (specially since in poll (oldpoll) use; committed  a few months ago, never implemented) . 
In word use it exist multiple times hence the question (in issue 6 last post) if it was not better te recode to old standard using pollans ?

PR #2685, opened 2017-06-11
Fix 'Illegal character in query: space is not allowed.' html validation error

arunshekher:
Dear Admins,
Please take a look and merge if applicable.

e107-news php-tags-link-fix


CaMer0n:
Thank you! 👍

PR #2688, opened 2017-06-15
[LAN] - Optimize theme admin LANs

arunshekher:
Changes:

	- Move all theme admin config LANs from e107_themes/bootstrap3/languages/English_admin.php to e107_languages/English/admin/lan_theme.php
	- Update e107_themes/bootstrap3/theme_config.php reflect the LAN terms change
	- Update hardcoded terms in e107_themes/voux/theme_config.php to use corresponding LANs from e107_languages/English/admin/lan_theme.php
	- Update e107_themes/landingzero/theme_config.php
	- Comment out duplicated LANs that have been moved to e107_languages/English/admin/lan_theme.php in e107_themes/landingzero/languages/English_admin.php and e107_themes/bootstrap3/languages/English_admin.php

Hi @CaMer0n,
If these changes comply with the LAN optimization changes you've proposed in issue #6 please review and merge.

If not please advise so that I can modify and resubmit.

### DIFF
--- a/e107_languages/English/admin/lan_theme.php
+++ b/e107_languages/English/admin/lan_theme.php
@@ -130,3 +130,15 @@ define("TPVLAN_92", "New Theme Folder");
 define("TPVLAN_93", "Selection");
 define("TPVLAN_94", "Site theme changed to [x].");
 
+// theme admin config
+define("TPVLAN_95", "Branding:");
+define("TPVLAN_96", "Navbar Alignment:");
+define("TPVLAN_97", "Signup/Login Placement:");
+define("TPVLAN_98", "Bootswatch Styles:");
+define("TPVLAN_99", "Site Name");
+define("TPVLAN_100", "Logo");
+define("TPVLAN_101", "Logo &amp; Site Name");
+define("TPVLAN_102", "left");
+define("TPVLAN_103", "right");
+define("TPVLAN_104", "top");
+define("TPVLAN_105", "bottom");
--- a/e107_themes/bootstrap3/languages/English_admin.php
+++ b/e107_themes/bootstrap3/languages/English_admin.php
@@ -7,16 +7,16 @@
 +----------------------------------------------------------------------------+
 */
 
-define("LAN_THEMEPREF_00", "Branding:");
-define("LAN_THEMEPREF_01", "Navbar Alignment:");
-define("LAN_THEMEPREF_02", "Signup/Login Placement:");
-define("LAN_THEMEPREF_03", "Bootswatch Styles:");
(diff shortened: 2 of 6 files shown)


Jimmi08:
@arunshekher why this change?  What LAN optimization do you mean? (I saw that issue, but I didn't find this exact reason there)

These strings are specific for this theme, they are not used in other themes.
@tgtje @rica-carv

tgtje:
@arunshekher @Jimmi08  : i can see the concept, looks good (general LAN use for all themes IFF they are core bundled). 
However Jimmi also has a point there.Theme specific is (imo) better in this case (landingzero), because any change in either system LAN's OR theme ( since it becomes a mix of..) might cause issues. 
As for BS3/Voux.. it is more a _complete package_, so there it would work fine.

Moc:
I agree with @Jimmi08. These LANs are theme specific (although they are similar/ the same currently across the different themes. 

It is best to keep these outside the regular LAN files so we can keep the flexibility make adjustments per theme. 

Thanks for the PR though! :)

rica-carv:
Indeed, i agrre that those LANS are more related to theme than anything else, so despite there's a "rule" so themes shouldn't have LANS, i agree that those LANS should be kept in the theme....

arunshekher:
@Jimmi08, @tgtje, @Moc, @rica-carv
Optimization proposed by @CaMer0n in the issue #6 if I understand it correctly. Bottom-line, its up to the maintainer and the community to collectively to decide upon it. 

I just went ahead with my contribution based on the following rationale:

- These are core themes which will be controlled and maintained within the parent repository.
- At-least three of the terms are repeated across all three core themes bs3, voux and landing zero
- They are used within the admin configuration section of the theme

tgtje:
Yes, the proposed optimization is definitly there and followed nicely 👍. Work itself is great, and contribution certainly appreciated.

This is a good proposal/commit  although all are just commenting on (imo) 'overall idea regarding theme specific'things. (these are the pros and cons on Commenting areas :-) ) And do not forget as mentioned : the 'maintainer' (that is the key everything evolves around).

CaMer0n:
Yes, as Moc suggested, it is theme-specific and married to the theme_config.php and associated prefs. But I appreciate the work put into it. If you have any doubts with code and wish to discuss changes such as this before losing your time, please feel free to message me in the gitter chatroom.

arunshekher:
No problem @CaMer0n, barely took 10 minutes of my time. Thought the discussion can be brought in right here, as I'm a bit shy of real time communication platforms. Think I understand the collaborative workflow reasonably and in no way steadfastly think that all the PRs I submit are going to be accepted. Glad to contribute.

PR #2692, opened 2017-06-19
correct sql query

Jimmi08:


Jimmi08:
It started to work, I really don't know where problem was.

rica-carv:
Maybe cache?

PR #2706, opened 2017-06-24
Fix html errors - Unclosed element span, Element ul not allowed as child of element ul.

arunshekher:
Fix html errors - Unclosed element span, Element ul not allowed as child of element ul.

screen shot 2017-06-25 at 12 23 59 am


CaMer0n:
Thank you! 👍

PR #2730, opened 2017-07-14
Refactoring for Cron Scheduler.

lonalore:
Tested on e107projects.com. It works fine.

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -1,220 +1,29 @@
-#!/usr/bin/php -q
 <?php
-/*
-+ ----------------------------------------------------------------------------+
-||     e107 website system
-|
-|     Copyright (C) 2008-2014 e107 Inc 
-|     http://e107.org
-|
-|
-|     Released under the terms and conditions of the
-|     GNU General Public License (http://gnu.org).
-|
-*/
 
-// Usage: [full path to this script]cron.php --u=admin --p=password // use your admin login. 
-// test
+/**
+ * e107 website system
+ *
+ * Copyright (C) 2008-2017 e107 Inc (e107.org)
+ * Released under the terms and conditions of the
+ * GNU General Public License (http://www.gnu.org/licenses/gpl.txt)
+ *
+ * @file
+ * Handles incoming requests to fire off regularly-scheduled tasks (cron jobs).
+ *
+ * @example
+ * https://your-e107-website.com/cron.php?token=83ab94d77c2ea6c3422146f1ea4ab2e83c43de03
+ */
 
 $_E107['cli'] = true;
 $_E107['debug'] = false;
 $_E107['no_online'] = true;
 $_E107['no_forceuserupdate'] = true;
 $_E107['no_menus'] = true;
-$_E107['allow_guest'] = true; // allow crons to run while in members-only mode. 
+$_E107['allow_guest'] = true; // allow crons to run while in members-only mode.
 $_E107['no_maintenance'] = true;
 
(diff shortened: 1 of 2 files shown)


CaMer0n:
Thank you @lonalore !!

CaMer0n:
I upgraded e107.org and the cron jobs stopped functioning directly after that. Reverted this commit and they began again. Please re-submit after more testing. Thanks!

lonalore:
How do you call cron.php? Do you use wget, curl, lynx or you run PHP as a CGI?

lonalore:
I did some tests on my debian based development environment:

**wget** (Tested, it works)
`/usr/bin/wget -O - -q http://e107.dev/cron.php?token=TOKEN > /dev/null 2>&1`

**curl** (Tested, it works)
`/usr/bin/curl --silent --compressed http://e107.dev/cron.php?token=TOKEN > /dev/null 2>&1`

**lynx** (Tested, it works)
`/usr/bin/lynx -source http://e107.dev/cron.php?token=TOKEN > /dev/null 2>&1`

**PHP** (Tested, it works)
`/usr/bin/php /var/www/drupalvm/e107.dev/web/cron.php token=TOKEN`

And it also works on my VPS (crontab is managed by IspConfig) and on a hungarian webhosting (crontab is managed by cPanel).

CaMer0n:
Thank you @lonalore 

`token=` is a more recent addition, the older method used arguments. 
Called via cPanel's Cron job, it looks similar to this: 

`/public_html/cron.php 289e8c3ae39aa5ee11535fca26d24294bc018cd0`

lonalore:
You can use it without `token=` too. It uses `str_replace('token=', '', $pwd);` I didn't change that code:

[Old code](https://github.com/e107inc/e107/blob/master/cron.php#L33-L42) vs [New code](https://github.com/lonalore/e107/blob/cron/e107_handlers/cron_class.php#L1141-L1150)... no differences. :)

So, `/public_html/cron.php 289e8c3ae39aa5ee11535fca26d24294bc018cd0` should work too. Please check your file permission.

lonalore:
No, it doesn't. The permission is 644 on my cron_class.php, and it works. I have no idea. :/ 
I will try to do more tests on this week.
Thank you!

lonalore:
I might know what the problem is. As you mentioned, you run the cron using the following command:
```
/public_html/cron.php 289e8c3ae39aa5ee11535fca26d24294bc018cd0
```
But in this case, you attempt to run `cron.php` as a **shebang** script. It worked before my changes because of [this line](https://github.com/e107inc/e107/blob/master/cron.php#L1).

But I removed it, because it's path specificand it's only supported by UNIX based systems, and it does not work for example on Windows servers.

Please change your command to the command below, then try to run your cron again:
```
/usr/bin/php /public_html/cron.php 289e8c3ae39aa5ee11535fca26d24294bc018cd0
```

If we really want to support running cron as a UNIX shell script, then we should create a **cron.sh** file instead of mixing the PHP and shell script. For example:

```bash
#!/usr/bin/env php
<?php
include 'cron.php';
```

CaMer0n:
For sure it is due to the shebang removal. I don't think it is a good idea to remove it, and then have to ask everyone who has crons configured on all their sites to update all their cron configurations.  Windows is not officially supported, while the majority are using Unix based systems, so rather than modifying what works for Unix already, perhaps a plugin can be made to support Windows systems if that's what you need?.

lonalore:
I put the shebang line back to the code. I'll create a new pull-request.

PR #2757, opened 2017-08-04
fix double caption when it failed

drasehat:
tablerender make additional caption and breadcrumb above the error message when the process failed. (email not recognized)
no problem if the caption, is simple .. but if we have styled caption it look weird.
maybe there are better solutions ?

thank you


CaMer0n:
Looks good! Thanks.

PR #2766, opened 2017-08-11
Remove margin from empty draggable-panels.

lonalore:
So this margin will be removed:
kepernyokep_2017-08-11_10-26-36


CaMer0n:
Great! Thank you!

Issue #2778, opened 2017-09-02
FAQ plugin - parse error because concatenation LAN strings


Jimmi08:
Latest github

After installing FAQ plugin:
`Parse error: syntax error, unexpected '.', expecting ')' in C:\xampp\htdocs\e107v2.1.6v\e107_plugins\faqs\admin_config.php on line 232`
Reason:  concatenation strings in admin_ui is not allowed:
`
		protected $preftabs				= array(LAN_GENERAL, LAN_FAQS_ASK_A_QUESTION, LAN_MANAGE.": ".LANA_FAQ_QUESTIONS, LAN_MANAGE.": ".LAN_CATEGORIES);`

`Parse error: syntax error, unexpected '.', expecting ')' in C:\xampp\htdocs\e107v2.1.6v\e107_plugins\faqs\admin_config.php on line 250`

`			'orderby'					=> array('title'=> LAN_ORDER,        'tab'=>0, 'type'=>'dropdown', 'writeParms'=>array('faq_order-ASC'=>LANA_FAQ_PREF_16, 'faq_id-ASC'=>LAN_ID." ".LAN_ASC, 'faq_id-DESC'=>LAN_ID." ".LAN_DESC, 'faq_datestamp-ASC'=>LAN_DATE." ".LAN_ASC, 'faq_datestamp-DESC'=>LAN_DATE."-".LAN_DESC)),`

After left only clean LAN strings, error is solved.

PHP 5.6
Only English lang file

Jimmi08:
https://e107help.org/3705/error-on-faqs-plugin

Jimmi08:
https://github.com/e107inc/e107/commit/0fa1246b6173919e3ba57c7f3aa28204d54885cc

tgtje:
Idea but no time now... what happens when **"** become **'**  (double to single) I can not check as i have no real error reported.. By editor (pspad) it displays it accepts it....

Jimmi08:
I wasn't never able to use more than one LAN string as value in admin_ui().  I always need to concate them in lang file, not in admin_ui().

rica-carv:
Err, @Jimmi08, i can, althought in a "perverted" way....
`define("LAN_EUSER_ADMIN_COLORH",  LAN_EUSER_ADMIN_CLICKHERESTART.LAN_EUSER_ADMIN_PICKCOLOR);
`
I define it outside of the class (but inside admin_config.php), and then use LAN_EUSER_ADMIN_COLORH inside the class....

[...]

PR #2781, opened 2017-09-06
Update FAQ plugin

yesszus:
Bug solved: #2778


tetolk:
This works now, very nice work guys!

Issue #2858, opened 2017-11-11
not able to remove datastamp field value from field


Jimmi08:
If I start news start (or any datestamp field)  and then I want to remove it (left it empty as default) , it's not possible, old value is still inserted

![image](https://user-images.githubusercontent.com/5429548/32688674-d57c40a6-c6d5-11e7-84e5-46d84c6f9c39.png)
![image](https://user-images.githubusercontent.com/5429548/32688678-e428a2e8-c6d5-11e7-9fd8-f72af5b31d22.png)

if I manually rewrite value (without clicking it in calendar  - which is really annoying with little change)
![image](https://user-images.githubusercontent.com/5429548/32688712-735f173a-c6d6-11e7-86b7-0543ae53c2af.png)

saved value is not changed
![image](https://user-images.githubusercontent.com/5429548/32688703-3353d090-c6d6-11e7-9313-0bc2cbc2eb06.png)

this is admin_ui issue not news, news was just example

CaMer0n:
Perhaps the javascript is ignoring empty values. @lonalore  ?

CaMer0n:
@Jimmi08 Can this be closed now?

Jimmi08:
Tested. It works. Thank you very much

PR #2864, opened 2017-11-14
Fix for Issue #2858.

lonalore:
#2858


CaMer0n:
@lonalore Thanks. I tried some of these in the past if I remember well, and it made it slower. Anyway.. will merge and we can test.

PR #2902, opened 2017-12-09
og:image setting if there is no already set one

Serios:
I think the logic for setting og:image in social plugin is to be used site wide when no other is defined, but currently it's set too early in header so for example if we looking at news item which have some images, the social plugin og:image is grabed first by fb sharer since it came first in page output. My modification check if there is og:image defined and if there is already, skips adding the social plugin one.

### DIFF
--- a/e107_plugins/social/e_event.php
+++ b/e107_plugins/social/e_event.php
@@ -26,8 +26,43 @@ class social_event
 
 	function config()
 	{
+		$event = array();
+		
+		$event[] = array(
+			'name'     => "system_meta_pre",
+			'function' => "ogg_image_add",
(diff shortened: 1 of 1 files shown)


CaMer0n:
Excellent! Thank you @Serios !! 👍

CaMer0n:
@Serios I made some small changes. Would you mind re-testing after syncing with the github repo?

Serios:
@CaMer0n, I do some testing and it seems to work perfect, thank you.

CaMer0n:
Thank you @Serios for re-testing and for the pull-request. Nice coding! 👍

Issue #2894, opened 2017-11-28
Q/I: Find Plugins


LaocheXe:
Checking out Plugin Manager > Find Plugins -I've notice something odd, i don't know if this is a bug, so I will ask this: How does the Find Plugins sort function work? I notice the fist page shows featured v2 plugins, and some none-featured v2 plugins. the second page shows v1 plugins. and I've checked most, if not all the pages for any more v2 plugins (because I don't see some that I used to see there).

This capture shows the first page - which displays 10 Made for v2 plugins
![capture-p1](https://user-images.githubusercontent.com/5818022/33335377-8889b25a-d432-11e7-85d5-3e086e2264d0.JPG)

The next page only shows v1 plugins, where is the rest of v2?
![capture-p2](https://user-images.githubusercontent.com/5818022/33335468-c23337d8-d432-11e7-8620-a4de0e0c5714.JPG)

I have one plugin for v2 on e107.org - can't find it looking all 20 pages - so I thought maybe it's because I have the plugin already, however, as you see in the first capture (image) it list v2 plugins even if I have them or not.

So I used the search function, and found it.
![capture-ps](https://user-images.githubusercontent.com/5818022/33335556-095c31fa-d433-11e7-86f0-9a29d629ee56.JPG)

So how does this work? I remember seeing NodeJS plugin, or something similar before, did the plugin get removed? Or is the Find Plugin working by showing new uploaded plugins, and then showing most downloaded?

LaocheXe:
I would think that this would show all version 2 plugins before any version 1 since most version1 plugins are outdated.

lonalore:
The problem was that there were 10 items in the table, but the list was paginated by 30 items. So the second and third pages were missing.

PR #2925, opened 2017-12-18
Issue #2894

lonalore:
#2894


rica-carv:
Excuse me for this, but wouldn't it be more straight forward and cleaner to have those variables (`$perPage, $amount, $size`) removed and use the already defined constant `PLUGIN_ITEMS_PER_PAGE`?

I just can't see the benefit of reusing a constant as multiple variables, for me it's ankward....

Sorry, just my 2 cents......

lonalore:
Good point! :+1: This was just a quick fix. Feel free to reroll it.

LaocheXe:
Would it also be possible to add a time error for people who try to download to many plugins at once? Currently still getting the 
[#] Bad Link
Unable to continue

CaMer0n:
Great! Thank you @lonalore

PR #2950, opened 2018-01-04
Master

dimante:
needed change to prevent \' showing on screen


e107_themes/landingzero/languages/English.php
```diff
@@ -21,7 +21,7 @@
 define("LAN_LZ_THEME_11", "Phone");
 define("LAN_LZ_THEME_12", "Your message here...");
 define("LAN_LZ_THEME_13", "Get in Touch");
-define("LAN_LZ_THEME_14", "We love feedback. Fill out the form below and we\'ll get back to you as soon as possible.");
+define("LAN_LZ_THEME_14", "We love feedback. Fill out the form below and we&apos;ll get back to you as soon as possible.");
```
CaMer0n: Thanks. Did you try replacing `we\'ll ` with `we'll`

dimante:
Yes that did not work on the display either the code in there with \' is
wrong though period...


Best Regards,

John Gates, CISSP

*Let’s Connect!*

<https://twitter.com/johngatesIII>   <http://www.linkedin.com/in/JohnGates>

e107_themes/landingzero/languages/English.php
dimante: I did and it did not display properly... What I can tell you is the '\' shows in the rendered text and is wrong ;-)

dimante: that was supposed to be the backslash char in the ''

Jimmi08:
@yesszus Could you please look at this?  You language could have the same problem.  There shouldn't be `&apos;`   correct fix should be `we'll`
Thanks

yesszus:
@Jimmi08 tested and work this `&apos;` and `we'll` without display error.

Issue #2898, opened 2017-12-04
bootstrap4  - report


Jimmi08:
Defined version 4 this way:

`define("BOOTSTRAP", 	4);`
loaded files this way:
```
e107::css('url', 'https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/css/bootstrap.min.css');
e107::css('url', 'https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css');
e107::js("footer", 	    'https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.11.0/umd/popper.min.js', 'jquery' );
e107::js("footer", 	    'https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/js/bootstrap.min.js', 'jquery' );
```
I can't rid the "plus" at the beginning of site:
![image](https://user-images.githubusercontent.com/5429548/33577363-06c37740-d943-11e7-973f-5c119d942a9b.png)
![image](https://user-images.githubusercontent.com/5429548/33577375-109122a4-d943-11e7-8764-f0a4c568f9c3.png)

similar problem was if bootstrap 3 js wasn't fully version, modal script was missing or something like this.

Thanks

Jimmi08:
btn-default - replaced by btn-secondary
But correct replace is use btn-primary for all buttons in e107, only second button should be btn-default

Jimmi08:
this needs to be fixed for bootstrap 4 too:
```
function breadcrumb($array)
{

$sep = (deftrue('BOOTSTRAP') === 3) ? "" : "<span class='divider'>/</span>";

}
```

Jimmi08:
With bootstrap 4 there is problem with using fade class in alert debug message.  Quick fix in theme style.css
```
.s-message.fade {
    opacity: 1;
    visibility: inherit;
}
```
Without it there is just empty place, no message

Jimmi08:
**Comment buttons**
`{COMMENT_REPLY} {COMMENT_EDIT} {COMMENT_MODERATE}`
have different classes
`e-comment-reply btn btn-default btn-secondary btn-sm btn-mini btn-xs`
`btn btn-default btn-secondary btn-mini btn-xs e-comment-edit`
`e-comment-delete btn btn-default btn-secondary btn-sm btn-mini btn-xs`

there is missing btn-sm from COMMENT_EDIT and button has different size in bootstrap4
[...]

PR #2959, opened 2018-01-08
download plugin with bootstrap 4 #2898

Jimmi08:


e107_plugins/download/download_shortcodes.php
```diff
@@ -691,13 +695,18 @@ function sc_download_view_link($parm)
...
+		switch (deftrue('BOOTSTRAP')) {
+				case 4:
+						$img = e107::getParser()->toGlyph('fa-download',false);
+						break;
+				case 3:
+						$img = e107::getParser()->toGlyph('icon-download.glyph',false);
+						break;
+				default:
+						$img = "<img src='".IMAGE_DOWNLOAD."' alt='".LAN_DOWNLOAD."' title='".LAN_DOWNLOAD."' />";
+						break;
```
CaMer0n: Perhaps using `toIcon()` would be more efficient. When it detects the .glyph extension it automatically uses` toGlyph()` otherwise, it returns an ``. ie. you could send it `fa-download.glyph`, `icon-download.glyph` or `{e_PATHXX}whatever.png`

Jimmi08: It was written this way before, I just added switch. I need an example where this is used correct way and I can repeat it, but I am not able to write it myself.

CaMer0n: You should be able to find toIcon() used throughout the core in various ways. Please let me know if you can't, and I'll do a search for you.

Jimmi08: yes, but I wrote you somewhere that that code looks reductant because it's already in shortcodes.

e107_plugins/download/templates/download_template.php
```diff
@@ -512,17 +512,13 @@
...
-    </li>
-    <li class="next">
-    	{DOWNLOAD_VIEW_NEXT}
-    </li>
-    </ul>
+<div class="d-flex">
+<ul class="nav mx-auto   pager download-view-nextprev">
+    <li class="nav-item pager-prev pager-next">{DOWNLOAD_VIEW_PREV}</li>
+    <li class="nav-item">{DOWNLOAD_BACK_TO_LIST}</li>
+    <li class="nav-list ext">{DOWNLOAD_VIEW_NEXT}</li>
```
CaMer0n: I think that 'ext' is meant to be 'next' ?
Also, what is d-flex? Is that part of BS4?

Jimmi08: I really can't remember now

d-flex: 
https://getbootstrap.com/docs/4.0/utilities/flex/

hm, and ext really looks like mistypo.  sorry

PR #2966, opened 2018-01-09
Captcha fix

dimante:
Fix for captcha


e107_handlers/secure_img_handler.php
```diff
@@ -402,7 +402,7 @@ function render($qcode, $color='')
 
 		if(isset($secureimg['font']) && is_readable($fontpath.$secureimg['font']))
```
rica-carv: Just my two cents: Shouldn't the is_readable check also search with realpath, ie:
`is_readable(realpath($fontpath.$secureimg['font']))`

CaMer0n:
Thank you! I made some additional changes.

Issue #3003, opened 2018-01-24
Multiple form validation logic flaws in Site Preferences


Deltik:
## Problem

If you save a username and password in the browser's autofill, the browser will try to populate "SMTP Username" and "SMTP Password" in Site Preferences (`/e107_admin/prefs.php`).  The autofill needs to be stopped in the first place.

Even if the autofill is not stopped, the form will not submit on some passwords.

The current required pattern is as follows: `[\S]{4,}`
This means that whitespace is not accepted, and there must be at least four characters.

To make matters worse, if "smtp" is not selected and visible as the "Emailing method", the user will not be taken to see the form validation error.

## Solution

Three fixes are needed:
* Disable autofilling of the SMTP credentials fields
* Allow whitespace in the SMTP password
* Make the SMTP password optional

## Screenshot

[![screenshot_20180124_041836](https://user-images.githubusercontent.com/1364268/35326930-c544a12a-00bd-11e8-9707-1fa826e2b220.png)](https://user-images.githubusercontent.com/1364268/35326930-c544a12a-00bd-11e8-9707-1fa826e2b220.png)

Jimmi08:
Maybe you will be able to fix this too:
https://github.com/e107inc/e107/issues/2882

PR #3004, opened 2018-01-24
Improvements to Preferences form validation for SMTP credentials

Deltik:
SMTP server max string length increased from 50 to 128
SMTP username max string length increased from 50 to 128
SMTP password max string length increased from 50 to 128

SMTP password is actually optional now, as intended

SMTP password now accepts spaces

SMTP password needs to be at least 4 characters long

Autocomplete disabled for Firefox 38+, Chrome 34+, and Internet Explorer 11+

Fixes: #3003


CaMer0n:
Thank you @Deltik

Issue #2882, opened 2017-11-23
edit user in admin area - problem with prefilled fields


Jimmi08:
I am editing user via this option:
![image](https://user-images.githubusercontent.com/5429548/33187243-c451482a-d090-11e7-978f-130b21893ba5.png)

It fills these fields with my creditials  (probably browser does this, I have problems with this in Chrome)
![image](https://user-images.githubusercontent.com/5429548/33187258-e2e1c95e-d090-11e7-87ec-5dda2bc4c27a.png)
There is my password, not original from user.

If I don't notice this, it replace password for edited user

PR #3005, opened 2018-01-24
Disable new password autocomplete for admins

Deltik:
Admins can now set user passwords longer than 20 characters.  New character limit is 128.

Autocomplete is disabled for new user passwords set by admins so that they don't autofill their own passwords into the users'.

Fixes: #2882 

---

_My editor messed with the DOS-style line endings, so go to https://github.com/e107inc/e107/pull/3005/files?w=1 to see the real changes._


CaMer0n:
Thank you! @Deltik  For the sake of the changelog, I'm going to add this code manually.

Issue #3006, opened 2018-01-24
Allow spaces in passwords system-wide by default


Deltik:
## Problem

Some e107 users and admins may use whitespace in their passwords, but the client-side HTML validation currently prohibits them by default:

From `./e107_handlers/form_handler.php`:

```php
$options['pattern'] = vartrue($options['pattern'],'[\S]{4,}');
```

The server accepts passwords with whitespace if this client-side validation is disabled, so there does not appear to be a technical reason for this restriction.

## Solution

Allow whitespace characters in passwords:

```php
$options['pattern'] = vartrue($options['pattern'],'.{4,}');
```

PR #3007, opened 2018-01-24
Allow spaces in passwords system-wide by default

Deltik:
Some e107 users and admins may use whitespace in their passwords, but the
client-side HTML validation currently prohibited them by default.

The server accepts passwords with whitespace if this client-side validation is
disabled, so there does not appear to be a technical reason for this
restriction.

This pull request allows whitespaces in passwords by default.

Fixes: #3006


CaMer0n:
I believe we should prevent spaces at the beginning/end to prevent copy/paste issues.

CaMer0n:
Thank you ! 👍

Issue #3011, opened 2018-01-28
Add DECIMAL type support to permissive field validation in db_verify


Deltik:
# Problem

`db_verify` is too strict in validating `DECIMAL` table field types.

Related: #2998

## Examples

**Expected:** `DECIMAL(10,2) NOT NULL DEFAULT '0.00'`
**Actual:** `DECIMAL(10,2) NOT NULL DEFAULT 0.00`

# Screenshot

![Screenshot](https://user-images.githubusercontent.com/1364268/35481675-95452a48-03ed-11e8-951c-798cbdb81d75.png)

Issue #2998, opened 2018-01-23
db_verify verifies default values too strictly


Deltik:
# Problem
Even if it's a new installation of e107, `db_verify` will be too strict in validating database field defaults that are equivalent to each other.

This can happen on some MySQL or MariaDB servers that correct the field definitions, which were not strictly correct in the table structures built into e107 in the first place.

## Examples

**Expected:** `INT(10) UNSIGNED NOT NULL DEFAULT '0'`
**Actual:** `INT(10) UNSIGNED NOT NULL DEFAULT 0`

**Expected:** `MEDIUMTEXT`
**Actual:** `MEDIUMTEXT DEFAULT NULL`

## Screenshot

[![dqmf3ejezo8lhvsqqwnsoogdsu6rmz6znupqyxqirqvevdk](https://user-images.githubusercontent.com/1364268/35279176-eadd47e0-0011-11e8-82e3-a7957eb150d6.png)](https://user-images.githubusercontent.com/1364268/35279176-eadd47e0-0011-11e8-82e3-a7957eb150d6.png)

PR #3012, opened 2018-01-28
Permissive numeric field validation in db_verify

Deltik:
`db_verify::diffStructurePermissive()` now checks for all MySQL/MariaDB numeric types and permissively allows the expected and actual default values to be strings rather than numerical.

Fixes: #3011
Related: #2998


CaMer0n:
Awesome! Thank you 👍

Issue #3013, opened 2018-01-28
File Inspector broken in PHP 7 due to language changes


Deltik:
# Problem

If you try to run the File Inspector (`/e107_admin/fileinspector.php`) on PHP 7 or newer, an error like this appears:

> **Fatal error:** Uncaught ArgumentCountError: Too few arguments to function file_inspector::inspect(), 4 passed in /home/deltik/public_html/e107_admin/fileinspector.php on line 963 and exactly 6 expected in /home/deltik/public_html/e107_admin/fileinspector.php:572 Stack trace: #0 /home/deltik/public_html/e107_admin/fileinspector.php(963): file_inspector->inspect(Array, Array, 0, '/home/deltik/pu...') #1 /home/deltik/public_html/e107_admin/fileinspector.php(87): file_inspector->scan_results() #2 {main} thrown in **/home/deltik/public_html/e107_admin/fileinspector.php** on line **572**

This behavior does not occur in PHP 5.4 through PHP 5.6.

# Screenshot

![7b0a8fb1-c57c-4466-8903-74234935dba1](https://user-images.githubusercontent.com/1364268/35481960-8876ac3c-03f3-11e8-8cd6-0220892b683c.png)

PR #3014, opened 2018-01-28
Proper file_inspector::inspect() optional arguments

Deltik:
`file_inspector::inspect()` now works in PHP 7 because the pass-by-reference variables are now optional as they were intended.

Fixes: #3013


CaMer0n:
Thank you! 👍

Issue #3021, opened 2018-02-02
New e107 install without admintheme set in SitePrefs causes infinite redirect loop


Deltik:
## Problem

New e107 installations can possibly set `'admintheme' => NULL` in e107_core's SitePrefs config.  Current logic attempts a migration from e107 v1.x to e107 v2.x in `e107_admin/auth.php`:

```php
// Required for a clean v1.x -> v2 upgrade. 
$core = e107::getConfig('core');

if($core->get('admintheme') != 'bootstrap3')
{
        $core->update('admintheme','bootstrap3');
        $core->update('adminstyle','infopanel');
        $core->update('admincss','admin_dark.css');
        $core->set('e_jslib_core',array('prototype' => 'none', 'jquery'=> 'auto'));
        $core->save();
        e107::getRedirect()->redirect(e_SELF);
}
```

`$core->update('admintheme','bootstrap3');` doesn't have any effect, so the redirect always happens.

## Screenshot
![screenshot_20180202_080654](https://user-images.githubusercontent.com/1364268/35737184-4f151888-07f0-11e8-8b05-b1885e04789e.png)

Issue #2933, opened 2017-12-26
Cannot pass step 6 of installation on PHP 7.2


fwhcat:
11 days ago, with the commit https://github.com/e107inc/e107/commit/ff2b79145ec6030406070e9cd8b0ef09fc366260
you introduced a bug.

Calling $output .= $this->thumbnailSelector('sitetheme', $opts, DEFAULT_INSTALL_THEME); is a great idea, but in $opts array, preview are the templates folder and not their preview image,
the thumbnailSelector checks if the preview field is a file and in that case, it fails.

OS: ArchLinux
PHP: 7.2
MariaDB: 10.1.29

fwhcat:
Here is an attached screenshot:

![image](https://user-images.githubusercontent.com/20411249/34355465-c49ed270-ea35-11e7-9573-d6bef73359ba.png)

Jimmi08:
Installation on php 5.6 without problems.

tgtje:
Although local (php 7.1.1) no issues (1,5 days ago)

CaMer0n:
@fwhcat Please test again with install.php?debug as your URL so that you can see the errors. 

    'preview'   =>  e_THEME.$val."/".$themeInfo['thumbnail'],

My guess is that the XML is not being parsed correctly.

fwhcat:
No errors displays. The following is the array displayed via ?debug
```
Array
(
    [language] => English
    [mysql] => Array
        (
            [createdb] => 1
            [server] => localhost
            [user] => root
            [password] => test
            [db] => e107
            [prefix] => e107_
            [overwritedb] => 1
        )

    [paths] => Array
        (
            [hash] => 416f4602e3
        )

    [admin] => Array
        (
            [user] => root
            [display] => root
            [email] => root@localhost
            [password] => test
        )

    [prefs] => Array
        (
            [admincss] => css/bootstrap-dark.min.css
            [sitename] => test
            [sitetheme] => 
        )

    [generate_content] => 1
    [install_plugins] => 1
)
```

My web server outputs a notice, might be related:
```
[Tue Dec 26 19:12:53 2017] PHP Notice:  Undefined index: @value in /home/fabien/Dev/e107/e107_handlers/theme_handler.php on line 626
[Tue Dec 26 19:12:53 2017] PHP Stack trace:
[Tue Dec 26 19:12:53 2017] PHP   1. {main}() /home/fabien/Dev/e107/install.php:0
[...]

PR #3022, opened 2018-02-02
e_pref::update() accepts existing keys with NULL

Deltik:
`e_pref::update()` will now update keys' values where the keys have a value of `NULL`.

Fixes: #3021
Fixes: #2933


tgtje:
See post php 7.2 issue in regards to what database does... (loop 3021 is handled nice)..

CaMer0n:
Thank you! 👍

Issue #3015, opened 2018-01-28
Remove dependency on MySQL/MariaDB CREATE PROCEDURE permission


Deltik:
# Problem

On Wednesday, 24 January 2018, @CaMer0n asked me:

> Good morning! I was wondering if you would mind taking a look at this issue?
>
> In `e107_handlers/admin_ui.php` is a method: `getParentChildQry()`. It uses custom MySQL functions to sort by parent/child fields of the table taking into account the 'order' field of the table.
>
> I talked to one of the users about trying to do this is PHP alone to remove this need for mysql-function permissions. Here is his pull-request: #2929
>
> Unfortunately, it didn't work correctly. The best test case is to use the vstore plugin and add some multi-level categories in the admin area.
>
> If you don't have time, I completely understand - please feel no obligation. Just thought you might have the skills to solve this one.

The task is to port the custom MySQL/MariaDB functions `getDepth()` and `getTreeSort()` to pure PHP so that e107 will no longer be dependent on the `CREATE PROCEDURE` permission in MySQL/MariaDB.

# Screenshot of Expected Behavior

![Screenshot of expected behavior](https://user-images.githubusercontent.com/1364268/35485559-07e97728-0427-11e8-9b6a-3e132fffecb2.png)

PR #3025, opened 2018-02-03
Reimplement e_tree_model sorting in pure PHP (with pagination)

Deltik:
Eliminated e107 dependency on the MySQL/MariaDB `CREATE PROCEDURE` permission

As a bonus, any attempt to use e107's native custom sorting (e.g. `?field=cat_name&asc=desc`) will preserve the correct tree hierarchy.  The past behavior was to sort as requested but keep showing depth, even though the apparent tree hierarchy would be wrong.

This is the same implementation as #3016, except it **also supports pagination**.  The performance and memory usage overhead is expected to be negligible for the vast majority of users.

Fixes: #3015
Supersedes: #2929
Supersedes: #3016


CaMer0n:
Excellent!! 👍

Issue #3018, opened 2018-02-02
PHP 7.2: Installation blocked by theme thumbnails not loading


Deltik:
## Problem
It is not possible to install e107 v2.1.8 preview on PHP 7.2 because no thumbnails show up on step 6 of 8 for selecting a theme.

## Screenshot
![fireshot capture 81 - e107 installation __ step 6 - website pr_ - https___dev deltik org_install php](https://user-images.githubusercontent.com/1364268/35731933-2b54820c-07dd-11e8-9ff1-80771cadae24.png)

CaMer0n:
With this change in the pull-request, it actually caused the problem referenced. (PHP 5.4 and 7.1) Removing the change brought back the screenshots. Please retest, running the installation in debug mode. Theme caching is now disabled during debug. Perhaps this was related to the issue. 

`/install.php?debug`

Deltik:
If the expected output in PHP 5.4 through PHP 7.1 was
```php
    [screenshots] => Array
        (
            [image] => Array
                (
                    [0] => preview_frontend.png
                )

        )
```
… PHP 7.2 would have produced this:
```php
    [screenshots] => Array
        (
            [image] => Array
                (
                    [0] => Array
                        (
                            [@attributes] => 
                            [@value] => preview_frontend.png
                        )

                )

        )
```
… because `$tags = get_object_vars($xml);` in `xmlClass::xml2array()` would have contained the key `0` instead of being empty, and `0` would be interpreted further down in `if($tags)` as a tag.

`$tag` would be equal to `0` in `switch($tag)`, which causes the first case, `case '@attributes'`, to be triggered, and this leads to the undesired blank `@attributes` key and the undesired `@value`.

tgtje:
I use this issue post to keep you update, else it is allover : (issue php 7.2 and 7.2.2) all in one go:
4 screen captures (mp4) where is described the following (all is fresh; wamp and e107 latest)
nu1 = wamp64 php 722  WITHOUT xml commit (not accepted)  > fails at 6
nu2 = xampp php 7.2      WITHOUT xml commit (not accepted)  > fails at 6
nu3 = xampp php 711  (sorry typo movie)  WITHOUT xml commit (not accepted) > success...
nu4 = wamp 64 php 7.22  WITH xml commit > install ok, but sitelinks issue remains

https://tgtje.nl/video/nu1.mp4
https://tgtje.nl/video/nu2.mp4
https://tgtje.nl/video/nu3.mp4
[...]

PR #3019, opened 2018-02-02
Theme thumbnail now uses correct preview value

Deltik:
Fixes: #3018 

## Screenshot
fireshot capture 82 - e107 installation __ step 6 - website pr_ - https___dev deltik org_install php

PR #3026, opened 2018-02-04
Sanity check in xmlClass::xml2array() for PHP 7.2

Deltik:
Since PHP 7.2, running `get_object_vars()` on a `SimpleXMLElement` object may return sequential arrays rather than nothing if there were no children.

This pull request introduces a sanity check in `xmlClass::xml2array()` that ensures child tags are processed rather than sequential array keys.

This sanity check does not affect past major versions of PHP.

Fixes: #3018
Supersedes: #3019

### DIFF
--- a/e107_handlers/xml_class.php
+++ b/e107_handlers/xml_class.php
@@ -553,7 +553,7 @@ class xmlClass
 		}
 
 		//Recursive calls start here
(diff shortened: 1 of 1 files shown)


CaMer0n:
Excellent! Thank you! 👍

Issue #3027, opened 2018-02-04
PHP 7.2: XML imports contain no values


Deltik:
## Problem

`xmlClass::e107Import()` fails to import any values when running on PHP 7.2.  @tgtje took some videos showing the [expected output](https://github.com/e107inc/e107/files/1693327/nu3.mp4.gz) and the [actual output](https://github.com/e107inc/e107/files/1693328/nu4.mp4.gz).

<sup><sub>[@tgtje's original comment](https://github.com/e107inc/e107/issues/3018#issuecomment-362912918)</sup></sub>

Expected array excerpt:

```php
    [database] => Array
        (
            [dbTable] => Array
                (
                    [0] => Array
                        (
                            [@attributes] => Array
                                (
                                    [name] => core_media_cat
                                )

                            [item] => Array
                                (
                                    [0] => Array
                                        (
                                            [field] => Array
                                                (
                                                    [0] => Array
                                                        (
                                                            [@attributes] => Array
                                                                (
                                                                    [name] => media_cat_id
                                                                )

                                                            [@value] => 1
                                                        )
```

Actual array excerpt:

```php
    [database] => Array
        (
            [dbTable] => Array
                (
                    [0] => Array
                        (
                            [@attributes] => Array
                                (
                                    [name] => core_media_cat
                                )

                            [item] => Array
                                (
                                    [0] => Array
                                        (
                                            [field] => Array
                                                (
                                                    [0] => Array
                                                        (
                                                            [@attributes] => Array
                                                                (
                                                                    [name] => media_cat_id
                                                                )

                                                        )
```

The issue can be reproduced by installing e107 from branch `master` on PHP 7.2 and choosing "Include Content/Configuration" on step 6.  You'll see that the demo content will not be populated.

PR #3028, opened 2018-02-05
PHP 7.2 consistency hack in xmlClass::xml2array()

Deltik:
This is a workaround to make the output of `xmlClass::xml2array()` in PHP 7.2 equal to that of prior PHP versions.

Consider this fix to be a part 2 of #3026.

PHP 7.2 changed the behavior of how we extract tags from SimpleXMLElement resources.  This hack adds a toleration for either the PHP 7.2 behavior and the past behavior, which allows consistent parsing
of XML files.

One side effect of this change is the fixing of an innocuous bug that provided blank values in some "`@value`" keys that should not have existed in the first place.  This should hopefully have no practical effect on current uses of the modified method.

@tgtje: The bug you reported should be fixed as a result of this pull request.

Fixes: #3027

### DIFF
--- a/e107_handlers/xml_class.php
+++ b/e107_handlers/xml_class.php
@@ -561,16 +561,17 @@ class xmlClass
 			//loop through tags
 			foreach ($tags as $tag)
 			{
+				if(is_int($tag)) continue;
 				switch($tag)
 				{
 					case '@attributes':
 						$tmp = (array) $xml->attributes();
 						$ret['@attributes'] = $tmp['@attributes'];
 
-						if($count_tags == 1) //only attributes & possible value
+						if($count_tags == 1 || ['@attributes', 0] === $tags) //only attributes & possible value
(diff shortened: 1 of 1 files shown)


tgtje:
@Deltik will test asap (little time due to circum)

tgtje:
Aquick small test on wamp64 php 7.2.2. ONLY (both on MySQl server and Mariadb > 👍 
I had this commit pre-installed.. (using the very latest git version).
So first test is 100% on **INSTALL**, everything looks present, no erros to be seen.

tgtje:
Just note to complete : tested : **INSTALL ONLY**  php 5.6.32  and php 7.1.1.  (both mariadb)  no noticable sideeffect ; both install fine... (mind :local)

CaMer0n:
Thank you @Deltik and @tgtje !! 👍

Issue #3029, opened 2018-02-06
Admin UI Does not order the results


chory:
my URL = /e107_admin/users.php?from=0&field=user_id&asc=asc
result:
![error107](https://user-images.githubusercontent.com/4314070/35877423-899df7da-0b75-11e8-8a47-722265919b43.jpg)

Deltik:
Can you visit the same URL again, except add `&[debug=sql+]` to the end?

We would need to see the "SQL Analysis" farther down the page to get a better idea of why this issue is happening.

Deltik:
Caused by #3025

chory:
Now, it shows nothing
URL =/e107_admin/users.php?mode=main&action=list
![imagen](https://user-images.githubusercontent.com/4314070/36046472-5285d98c-0dd9-11e8-9ece-eed97d8231f7.png)

CaMer0n:
@chory Please try this URL:

`e107_admin/users.php?[debug=showsql]&field=user_id&asc=asc&from=0`

..then paste the listQry value here in the comments here.

chory:
> 
> searchQuery:
> 
> Array
> (
> )
> 
> Using Custom listQry
> QRY=SELECT SQL_CALC_FOUND_ROWS u.*,ue.* from e107_user AS u LEFT JOIN e107_user_extended AS ue ON u.user_id = ue.user_extended_id ORDER BY u.user_id ASC LIMIT 0, 10
>

Deltik:
I messed up.  I'll have a commit to fix this in a few minutes.

Deltik:
Fixed in #3032

chory:
Ok, works fine for me. Show members in correct order.
Great work

PR #3031, opened 2018-02-09
Implemented custom sorting for e_tree_model

Deltik:
In #3025, e_tree_model sorting was reimplemented in pure PHP, but this broke custom sorting (like `?field=cat_name&asc=desc`).

This pull request introduces a hack that simulates a subset of MySQL/MariaDB `ORDER BY` clauses, which should be sufficient for all known custom sorting that can be requested.

Tree formatting is always preserved, but custom sorting will apply for all items at a certain depth under the same parent.

This pull request also contains some minor formatting fixes and makes a minor change to some regex to make use of non-capturing groups.

Fixes: #3029


CaMer0n:
👍 Thank you!!

Issue #3033, opened 2018-02-09
Headless (CLI) e107 tries to use uninitialized SQL if siteurl preference is not cached


Deltik:
# Problem

If `e107_system/*/cache/content/S_Config_core.cache.php` exists and contains `siteurl`, headless e107 can skip the database for getting the website URL and continue initializing until the database actually loads.

Without the cache, this happens:

```php
deltik@host1 [~/public_html]$ /opt/cpanel/ea-php72/root/usr/bin/php -d allow_url_fopen=On -d display_errors=On cron.php
deltik@host1 [~/public_html]$ rm -fv e107_system/6de9d0d30a/cache/content/S_Config_core.cache.php
removed ‘e107_system/6de9d0d30a/cache/content/S_Config_core.cache.php’
deltik@host1 [~/public_html]$ /opt/cpanel/ea-php72/root/usr/bin/php -d allow_url_fopen=On -d display_errors=On cron.php

Fatal error: Uncaught Error: Call to a member function query() on null in /home/deltik/public_html/e107_handlers/mysql_class.php:535
Stack trace:
#0 /home/deltik/public_html/e107_handlers/mysql_class.php(803): e_db_mysql->db_Query('SELECT e107_val...', NULL, 'db_Select', false, '', '')
#1 /home/deltik/public_html/e107_handlers/pref_class.php(479): e_db_mysql->select('core', 'e107_value', 'e107_name='Site...')
#2 /home/deltik/public_html/e107_handlers/pref_class.php(450): e_pref->_load(false)
#3 /home/deltik/public_html/e107_handlers/pref_class.php(848): e_pref->load()
#4 /home/deltik/public_html/e107_handlers/e107_class.php(1009): e_core_pref->__construct('core', true)
#5 /home/deltik/public_html/e107_handlers/e107_class.php(1032): e107::getConfig()
#6 /home/deltik/public_html/e107_handlers/e107_class.php(4479): e107::getPref('siteurl')
#7 /home/deltik/public_html/e107_handlers/e107_class.php(413): e107->set_urls()
#8 /home/deltik/public_html/e107_handlers/e107_class.php(317): e107->_init(Array, '/home/deltik/pu...', Arra in /home/deltik/public_html/e107_handlers/mysql_class.php on line 535
```

Headless e107 cannot be used in this case.

Deltik:
Caused by f7eadfd588

PR #3037, opened 2018-02-12
Defers setting the SITEURL constant and dependencies until MySQL is connected

Deltik:
This is the least intrusive solution I could come up with.

Anything else I tried to reduce technical debt would break the very tangled dependency web, so I settled for this.

I also removed some unnecessary layers of indentation by replacing them with [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html), so the diff in GitHub might be harder to read.  Try viewing the diff [here](https://github.com/e107inc/e107/pull/3037/commits/d1a69b0c16ad06c45078b428aa12913df9959966?w=1).

Fixes: #3033

### DIFF
--- a/class2.php
+++ b/class2.php
@@ -464,6 +464,10 @@ elseif ($merror === 'e2')
 /* PHP Compatabilty should *always* be on. */
 e107_require_once(e_HANDLER.'php_compatibility_handler.php');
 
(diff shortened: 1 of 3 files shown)


CaMer0n:
Thank you! 👍

Issue #3044, opened 2018-02-23
Error viewing categories with admin-ui


kreossino:
Fatal error: Maximum function nesting level of '256' reached, aborting! in class2.php on line 2409

CaMer0n:
Confirmed. Vstore and Download categories affected. 
@Deltik An endless loop in the model tree?

PR #3045, opened 2018-02-23
e_tree_model::multiFieldCmp() string sort fields

Deltik:
An untested oversight in `e_tree_model::multiFieldCmp()` where `$sort_field` could be a string has now been corrected.

`$sort_field` now accepts a string to prevent infinite recursion.

Fixes: #3044


CaMer0n:
Thanks :+1:

Issue #3062, opened 2018-03-06
Loading user list results in 502 Bad Gateway with large user database


BigBoiJefe:
I have a user database of roughly 360,000.  When I try and access the user list i get a 502 Bad Gateway.  I've tried increasing php memory to 5000M, however, I now get 504 Gateway Time-out

Deltik:
Caused by #3025

PR #3066, opened 2018-03-07
e_tree_model performance improvements

Deltik:
e_tree_model is apparently used for flat lists as well as parent-child
relationships (trees). Trees are expected to be far smaller than possible flat
lists. Very large flat lists (10,000+ rows or greater) are rendered very slowly
because of the tree computation overhead.

This change figures out whether a flat list or a tree is requested and chooses
the appropriate code to run based on what is requested. Trees run the more
expensive code while flat lists are returned as-is.

In addition, the tree rendering code has been optimized. Optimizations:

* Unchanging tree node ID is set once instead of inside a foreach() loop
* The parent-child query is now sorted by the sort parent ID so that each move
  rows to tree nodes iteration doesn't have to run through every remaining row

Fixes: #3062


CaMer0n:
Thank you! :+1:

PR #3071, opened 2018-03-11
Update Featurebox English_global.php

yesszus:
Tested and work this!

### DIFF
--- a/e107_plugins/featurebox/featurebox_setup.php
+++ b/e107_plugins/featurebox/featurebox_setup.php
@@ -29,6 +29,7 @@ class featurebox_setup
 	function install_post($var)
 	{
 		e107::includeLan(e_PLUGIN.'featurebox/languages/'.e_LANGUAGE.'_admin_featurebox.php');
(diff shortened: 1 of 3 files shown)


tgtje:
Or adding  eg  e107::lan(feauturebox') ...  to **setup** file ???
Above Commit works > what would be prefered way?

Jimmi08:
All 03,04,05 lans needs to move to admin lan file, they are used during installation process too (names of categories)

I am curious about prefered way - like @tgtje asked .

CaMer0n:
Installation terms should NOT be in a `xxxx_global` language file. Global files are for information which may be necessary throughout multiple pages and area of e107.  eg. menu captions, admin navigation etc.  Installation terms are specific to a single page. 
Using `e107::lan() ` inside the setup file is the preferred way.

yesszus:
e107::lan() not work for me!
Added new line and work.

CaMer0n:
All those installation terms should be in the _admin language file (not _global).  e107::lan() won't work on a plugin which isn't installed yet. This is why includeLan() is better in this case.

PR #3100, opened 2018-04-20
Truly disable a button on FormSubmit

Serios:
We need to add attribute disabled to the button to truly disable it, otherwise the click event is fired up again and again causing form submissions

### DIFF
--- a/e107_web/js/core/all.jquery.js
+++ b/e107_web/js/core/all.jquery.js
@@ -373,6 +373,7 @@ var e107 = e107 || {'settings': {}, 'behaviors': {}};
 							if($button.attr('data-disable') == 'true')
 							{
 								$button.addClass('disabled');
+								$button.prop("disabled", true);
 							}
 						});
 


Serios:
@Cameron I don't think so, as far as I can see, this is code is fired up after form submission, missing required field prevents form from submitting. So the flow goes like this Button click -> Check if all required fields are filled-> If not prevent submission, display error / If yes submit the form -> Add class and attribute to the button so i't can be clicked again.

CaMer0n:
Thank you! I'll merge it.  :+1:

CaMer0n:
@Jimmi08 @tgtje Are you able to reproduce as @Alex-e107nl mentioned?

Jimmi08:
I tried it on php 7.0 and no problem with delete ban record.

Serios:
@Jimmi08 Save options and Remove expired bans doesn't work. When I tested deletion of ban record worked. However as I comment above in revert pull request. Already came up with solution, just waiting for CaMer0n to revert this commit, to submit it.

Jimmi08:
@Serios I was able to test only deleting, not using this.  I wrote there it.  Thank you for your work.

PR #3105, opened 2018-04-27
Revert "Truly disable a button on FormSubmit"

Serios:
Reverts e107inc/e107#3100
@CaMer0n  I can confirm what @Alex-e107nl mentioned. Just reproduced it. The issue is that if attribute disabled is set, the DOM element doesn't get included in the POST, therefore if in PHP we are checking for this element name to be posted it fails and the logic within too.


Serios:
I became with a different logic to stop form from multiply submitting. Will add it when this revert is accepted.

CaMer0n:
Thank you!

Issue #3075, opened 2018-03-14
Files stored on a different server - Protect with ngx_http_secure_link_module


BigBoiJefe:
I currently have a bunch of files stored on a file server and I have them stored in e107 as a direct download with referral blocking.  The problem with that is that the referral can be easily spoofed, I'm looking for more security for the downloads.

Thanks!

Deltik:
@BigBoiJefe: I have a proposed fix for this issue.

## Using the patch file

Here is a patch that you can try: [fix-3075.patch.txt](https://github.com/e107inc/e107/files/1960370/fix-3075.patch.txt)

To apply the patch, run the following commands:

```sh
cd /path/to/e107 # where /path/to/e107 is where your installation of e107 is
wget https://github.com/e107inc/e107/files/1960370/fix-3075.patch.txt -O fix-3075.patch
patch -p 1 < fix-3075.patch
```

To roll back the patch, run this command:

```sh
patch -p 1 -R < fix-3075.patch
```

To remove the patch file, run this command:

```
rm -fv fix-3075.patch
```

## Using the proposed fix

### Configure your NGINX download server

Set up your NGINX download location config with `secure_link_md5` to protect your downloads.  For an example of how to do this, see the **Using Secured URLs that Expire** section of [this blog post from NGINX](https://www.nginx.com/blog/securing-urls-secure-link-module-nginx-plus/).

This is a sample `secure_link_md5` expression:

```nginx
$secure_link_expires$uri$remote_addr secret
```

- `$secure_link_expires` enables the expiration of links generated by the e107 Downloads plugin.
- `$uri` matches the download path intended to be protected.
- `$remote_addr` ensures that only the client's IP address may be used to download the file.
- `secret` is a string known only to e107 and your NGINX config so that your users/downloaders cannot generate their own secure links.

### Configure e107

1. Navigate to `/e107_plugins/download/admin_download.php?mode=main&action=settings` in your web browser.
2. Click on the **Protection** tab.
3. Select **NGINX secure_link_md5** as the **URL protection mode**
4. Type in your `secure_link_md5` expression into the **NGINX secure_link_md5 expression** field.
5. In the **Duration of validity in seconds** field, enter a whole number for how long in seconds you want each generated download link to live after the link is generated.
6. Click the green **Update Options** button.

Now, all downloads will have `md5` appended as a query parameter, generated from the expression you put in.  If the expression contains `$secure_link_expires`, `expires` will be appended as a query parameter as well.

[...]

PR #3111, opened 2018-05-01
Add NGINX secure_link_md5 URL protection feature to e107 Downloads plugin

Deltik:
Fixes: #3075

## Rationale

When e107 Downloads redirects the downloader to the download URL, the downloader can trivially hotlink to the download in the future, even with referer protection, and the e107 site admin currently cannot prevent this.

By establishing a secret between e107 and the download server, undesirable hotlinking can be prevented.

## Implementation

This pull request adds a **Protection** tab to Downloads » Preferences to allow the configuration of secure links.

Only the **NGINX secure_link_md5** protection mode has been implemented.  As this is currently implemented, all downloads are affected by this mode.

A new class `NginxSecureLinkMd5Decorator` decorates mirror and direct URLs for each download to be in the expected secure link format for NGINX server blocks configured with `secure_link_md5`.

Three NGINX variables are supported in this pull request:

- `$secure_link_expires`
- `$uri`
- `$remote_addr`

### Configure e107

Ideally, both e107 and NGINX should be configured at the exact same time, but e107 may be configured first because doing so would minimize the chance of disruption to downloaders, since the **NGINX secure_link_md5** URL protection mode merely appends query string arguments to the download URL.

1. Navigate to `/e107_plugins/download/admin_download.php?mode=main&action=settings` in your web browser.
2. Click on the **Protection** tab.
3. Select **NGINX secure_link_md5** as the **URL protection mode**
4. Type in your `secure_link_md5` expression into the **NGINX secure_link_md5 expression** field.
5. In the **Duration of validity in seconds** field, enter a whole number for how long in seconds you want each generated download link to live after the link is generated.
6. Click the green **Update Options** button.

Now, all downloads will have `md5` appended as a query parameter, generated from the expression you put in.  If the expression contains `$secure_link_expires`, `expires` will be appended as a query parameter as well.

### Configure your NGINX download server

Set up your NGINX download location config with `secure_link_md5` to protect your downloads.  For an example of how to do this, see the **Using Secured URLs that Expire** section of [this blog post from NGINX](https://www.nginx.com/blog/securing-urls-secure-link-module-nginx-plus/).

This is a sample `secure_link_md5` expression:

```nginx
$secure_link_expires$uri$remote_addr secret
```

- `$secure_link_expires` enables the expiration of links generated by the e107 Downloads plugin.
- `$uri` matches the download path intended to be protected.
- `$remote_addr` ensures that only the client's IP address may be used to download the file.
- `secret` is a string known only to e107 and your NGINX config so that your users/downloaders cannot generate their own secure links.

If your `secure_link_md5` expression contains `$secure_link_expires`, the `secure_link` directive should read `secure_link $arg_md5,$arg_expires;` like in this example server block:

```nginx
server {
    listen 80;
        server_name secure-link-demo;

        location /files {
            root /var/www;
            secure_link $arg_md5,$arg_expires;
            secure_link_md5 "$secure_link_expires$uri$remote_addr secret";

            if ($secure_link = "") { return 403; }
            if ($secure_link = "0") { return 410; }
        }
}
```

If your `secure_link_md5` expression _does not_ contain `$secure_link_expires`, the `secure_link` directive should read `secure_link $arg_md5;` like in this example server block:

```nginx
server {
    listen 80;
        server_name secure-link-demo;

        location /files {
            root /var/www;
            secure_link $arg_md5;
            secure_link_md5 "$uri$remote_addr secret";

            if ($secure_link = "") { return 403; }
        }
}
```


CaMer0n:
Excellent! Thanks :+1:

Issue #3113, opened 2018-05-01
Bug: Issue with number format while saving to db (1,1 becomes 11)


SimSync:
This looks like a bug in e_model->toNumber() and e_db_mysql->_getFieldValue() ...

When the sitelanguage is set to something different than english (e.g. dutch or italian), where the decimals and thousands separator are different than in english speaking countries, those functions fail to translate a float nr. correctly.

e.g. 10.000,45 becomes 1000045. Even 4.4 and 4,4 are parsed to 44
The whole parsing goes wrong here...

PR #3114, opened 2018-05-01
closes e107inc/e107#3113 parses a string into a db ready number format

SimSync:
Fix for issue #3113 

new functions will convert float number like strings to numbers.
e.g.
1,1 > 1.1
1.1 > 1.1
1,001.12 > 1001.12
1.001,12 > 1001.12

Any other characters than - and 0 to 9 are removed from the string.
Solution adapted from https://secure.php.net/manual/en/function.floatval.php#114486

### DIFF
--- a/e107_handlers/model_class.php
+++ b/e107_handlers/model_class.php
@@ -1792,20 +1792,37 @@ class e_model extends e_object
 	 * @param string $value
 	 * @return integer|float
 	 */
-	public function toNumber($value)
-	{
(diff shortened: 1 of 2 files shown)


SimSync:
@CaMer0n Just refactored the toNumber() function from various places (mysql_class.php, model_class.php, validator_class.php) into a single function in e_parse_class.php. Hopefully, codeclimate validates my changes without (new) issue...

PR #3115, opened 2018-05-03
Prevent form from multiply submission

Serios:
This change will set var value to true when form is submitted, then if we click submit button again will prevent it from another submission. Adding attribute disabled to form button prevents its name to get trough POST call, thus breaking the logic if we checking for form button submission trough PHP.


Moc:
Related #3100

Moc:
And #3105

CaMer0n:
Thank you @Serios :+1:

PR #3121, opened 2018-05-08
Update userposts.php

yesszus:
Some corrections.


e107_languages/English/lan_error.php
```diff
@@ -64,4 +64,8 @@
 
 define("LAN_ERROR_45", "What can you do now?");
 define("LAN_ERROR_46", "Check log for details.");
+define("LAN_ERROR_47", "Validation error: News title can\'t be empty!");
```
Moc: There should be no need to escape the single quotes (`\'`) when using a double quotes define (`define(" ", " ")`). Please re-check.

CaMer0n:
Thanks @Moc ! :+1:

PR #3130, opened 2018-05-15
Fixed smaller issues in date_handler.php

SimSync:
### convert_date()
- removed an orphaned . (dot)

### strptime()
- fixed issue with wrong monthname on windows. e.g. it returned "April" where it should be "May"

### e_form::number()
Implemented additional options to be able to define the number of decimals allowed, the step (of the spinners), min and max value and a custom regex.
Unfortunatly, some browser do not respect the min/max boundaries and you can enter (manually) values that are bigger or smaller than allowed. My regex skills are not good enough to dynamically create a regex to limit the input to a certain range.

If you don't set any of the new options, the behaviour is the same as before (only positive integers).
Setting the "decimals" option to 0 dynamically creates a regex to allow only integers and sets the "step" (if not defined) to 1.
Setting the "decimals" option to 1 (or higher) dynamically creates a regex to allow float values with the given number of decimals and sets the "step" (if not defined) to "0.1", "0.01", etc. (depending on the number of decimals).
Setting the "step" option works only if the "decimals" option is **not** set.
Setting the "min" option to a value smaller than "0", dynamically defines the regex to accept negative values. 
Setting the "max" opton, defines the max value which can be entered. Unfortunatly, some browser don't respect the min/max setting and you can enter values that are larger or bigger than the given boundaries.
The "pattern" option allows the usage of a custom regex. When set, the decimals and step options will be ignored.

### DIFF
--- a/e107_handlers/date_handler.php
+++ b/e107_handlers/date_handler.php
@@ -173,7 +173,7 @@ class convert
 			break;
 			
 			case 'inputtime': 
-				$mask .= e107::getPref('inputtime', '%H:%M');
+				$mask = e107::getPref('inputtime', '%H:%M');
 			break;
 
 			case 'forum': // DEPRECATED - temporary here from BC reasons only
--- a/e107_handlers/date_handler.php
+++ b/e107_handlers/date_handler.php
@@ -791,7 +791,7 @@ class convert
 			
 			$unxTimestamp = mktime($vals['tm_hour'], $vals['tm_min'], $vals['tm_sec'], ($vals['tm_mon'] + 1), $vals['tm_mday'], ($vals['tm_year'] + 1900));
 			
-			$vals['tm_fmon'] = strftime('%B', mktime($vals['tm_hour'], $vals['tm_min'], $vals['tm_sec'], $vals['tm_mon']));
+			$vals['tm_fmon'] = strftime('%B', mktime($vals['tm_hour'], $vals['tm_min'], $vals['tm_sec'], $vals['tm_mon'] + 1));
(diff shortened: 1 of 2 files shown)


CaMer0n:
@SimSync Thank you. fyi. There's more than just date changes in this pull-request. 
Btw. Did you already test your date changes in linux too?

CaMer0n:
Thanks @SimSync !

CaMer0n:
We should probably look into upgrading this method eventually to use date_parse_from_format. 
See here: 

https://secure.php.net/manual/en/function.strptime.php

_Note:   Internally, this function calls the strptime() function provided by the system's C library. This 
function can exhibit noticeably different behaviour across different operating systems. The use of 
**date_parse_from_format()**, which does not suffer from these issues, is recommended on PHP 5.3.0 and later._

SimSync:
Yeah, I saw it and thought the same. Unfortunately, we have to translate the date/time formats from what is used now to what the new function requires. But this should be OK.
Where I worry about is, that the old version delivers an array where the month number is zero based and the new afaik not.
The old function was public, so we can't be sure, that nobody has used it in a plug in.
we can manipulate the result to look like a strptime result, but is this what we want?
Or do I think to much? :laughing:

rica-carv:
Shouldn't the [native PHP functions](http://php.net/manual/en/ref.datetime.php) already handle those issues?

CaMer0n:
@SimSync Yeah, I was thinking also to rename the keys so the result is like strptime() is now.  (ie. full Backward compatibility)  My main concern is translating the format consistently.

SimSync:
Than we need to write good tests... :laughing: :laughing: :laughing:

PR #3153, opened 2018-05-30
Closes #2265 User Extended Field for Date does not allow entry

### LINKED ISSUE #2265: User Extended Field for Date does not allow entry
Field was working, but since a recent GitHub sync (not sure when) the field stopped allowing users to either change this in their profile or allow users to specify at signup. As we have it compulsory for signup no users can sign up. Previously users had to enter YYYY-MM-DD for it to accept the date. However, you can edit this field from User Admin, where it shows a calendar in addition to typing the date in
extended user fields - manage - admin area three percenters mcc 2017-01-10 13-28-54
 (attachment 2)
users - manage - admin area three percenters mcc 2017-01-10 13-39-58

SimSync:
This fix closes issue e107inc/e107#2265 which pops up only when the JS files are cached (compressed).
There was a ";" missing at the end of the file after (jQuery).

### DIFF
--- a/e107_web/js/password/jquery.pwdMeter.js
+++ b/e107_web/js/password/jquery.pwdMeter.js
@@ -201,5 +201,9 @@ jQuery.fn.pwdMeter = function(options){
 	});
 
 }
-
-})(jQuery)
+/**
+ * ALLWAYS add a semicolon at the end, otherwise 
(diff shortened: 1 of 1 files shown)


CaMer0n:
:+1:  Thank you!

Issue #3151, opened 2018-05-29
Create an e_end for plugins?


oudoken:
Hi there,
There's a way to capture the OB at the very end of page from within a plugin?
I'm trying to integrate phpfastcache with e107 for creating static html pages (see also https://www.e107help.org/4108/capture-buffering-from-plugin).
For ob_start() e_module.php works fine but how to get the end page? e_footer.php it's called to early.
Maybe there's another method or it will be a good addition making an "e_end.php" to expose this functionality to plugins?
Thanks
Rob

PR #3183, opened 2018-06-12
Issue #3151 add an e_end to plugin class

oudoken:
Hi There,
Based on #3151 i've modified these files and seems to work ok, it will be an usefull addition to e107 for all the things related in output buffering (start and end) or maybe i haven't find a better option?

### DIFF
--- a/e107_core/templates/footer_default.php
+++ b/e107_core/templates/footer_default.php
@@ -374,13 +374,13 @@ $show = deftrue('e_POWEREDBY_DISABLE') ? "none" : "block"; // Let search engines
 unset($show);
 echo "\n</body>\n</html>";
 
-//hook into the end of page buffering
-//Load e_end.php files.
-if (!empty($pref['e_end_list']) && is_array($pref['e_end_list']))
+//hook into the end of page (usefull for example for capturing output buffering)
+//Load e_output.php files.
+if (!empty($pref['e_output_list']) && is_array($pref['e_output_list']))
 {
-	foreach($pref['e_end_list'] as $val)
+	foreach($pref['e_output_list'] as $val)
 	{
-		$fname = e_PLUGIN.$val."/e_end.php"; // Do not place inside a function - BC $pref required. . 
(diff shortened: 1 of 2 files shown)


CaMer0n:
@oudoken Nice work! How about calling it, e_buffer or possibly e_output ? Some of the code you copied from is a little outdated, I will update it so you can work from that.

oudoken:
@CaMer0n Hi Cameron, sure well e_end was really the mean of it... so i think it can be e_buffer or e_ouput (btw i prefer the last one).
In fact calling it e_end can motivate to rename e_module in e_start... But i leave the naming convention To You guys! 👍 
I'm already working on a V2.0 plugin that use phpFastCache for producing static html's cache files and it's working ok.

CaMer0n:
Thanks! Yeah, having e_end would be strange without an e_begin or e_start. e_module isn't quite the start either.. , so, let's go with **e_output.php**. If you can make the change and test it, I'll commit your pull-request. Thank you.

oudoken:
@CaMer0n Ok perfect lemme make the change Thanks a Lot Cameron

oudoken:
@CaMer0n Hi Cameron ok modified e_end in e_output also written some more in comments to explain a little more Thanks Rob 👍

CaMer0n:
Great! Thank you @oudoken ! :+1:

oudoken:
Hi @CaMer0n Thanks 👍 
If You have time check 
https://github.com/oudoken/static_cache
It's an example implementing this pull request
Thanks Kind Regards
Rob

CaMer0n:
Nice work Rob ! :+1:

CaMer0n:
Yes, JS manager.  
```
/**
	 * Minify JS/CSS for output
	 * @param string $minify
	 * @param string $type (js|css)
	 * @return string
	 */
	private function compress($minify, $type = 'js' )
    {
```
ie.  `$compressed = e107::getJs()->compress($code, 'css');`

PR #3197, opened 2018-06-19
#6 ToDo Lans bounceLAN in handlers/login.

MikeyGMT:
Tested login from login.php, from the login link in voux theme without issue. No bounced emails on test instance.

Sorry about the merge request, which is to update my copy! It does not seem to include anything.


CaMer0n:
Great! Thank you @MikeyGMT !

PR #3208, opened 2018-06-21
Fixes #3198 Legacy prefs, Missing forums; Fixes #3204 Filter/search forums

### LINKED ISSUE #3198: problems  with forum
1. I still get this message (it's new installation)
image

2. I can't find forum 
I see on Frontend that they are existing, but they are not displayed in admin area. I exported forum xml data and missing forum is there, it's not just displayed in admin area
 
Forums are there:
image

**But they are not in admin area (only parent forum)**
- there is forum with similar structure and subforums are displayed

No idea what to do.  Sef-urls checked, no missing, no doubles. 
I fixed this by set forum_parent and forum_sub to 0 in PhpMyAdmin, then moved all topics a deleted problematic forums.

### LINKED ISSUE #3204: forum - searching doesn't work
Latest Github, php 7.0.29

Isuppose that it should search at least title:

image

image

SimSync:
**Fixes #3198**
- Removed the orphaned debug message concerning "legacy forum menu prefs detected. Upgrading ..." as it actually doesn't detect or upgrade anything.
- Added a workaround to the e_tree_model::moveRowsToTreeNodes() method to make sure all data is displayed, even if the row order isn't as expected (e.g. if 1 sort field isn't enough to presort the data)
- Added a fix to the admin_ui::joinAlias() method for the case that 2 or more aliases point to the same table and to make sure the first found alias isn't overwritten by the following.

**Fixes #3204**
- Added a workaround to the forum_admin.php to filter the list by a search word, because the tree model doesn't support this at the moment.

@Deltik Could you please check my changes, especially those to the e_tree_model::moveRowsToTreeNodes() method? It is only a small one and once you're into it... could you think about a method to filter the list without requiring the parent nodes in the result? See my workaround in the forum_admin.php ...

### DIFF
--- a/e107_handlers/admin_ui.php
+++ b/e107_handlers/admin_ui.php
@@ -3934,7 +3934,7 @@ class e_admin_controller_ui extends e_admin_controller
 			$keys = array();
 			foreach($matches[1] AS $k=>$v)
 			{
-				if(varset($matches[3][$k]))
+				if(varset($matches[3][$k]) && !array_key_exists($v, $this->joinAlias))
 				{
 					$this->joinAlias[$v] = $matches[3][$k]; // array. eg $this->joinAlias['core_media'] = 'm';
 				}
--- a/e107_handlers/model_class.php
+++ b/e107_handlers/model_class.php
@@ -3458,7 +3458,10 @@ class e_tree_model extends e_front_model
 			// Note: This optimization only works if the SQL query executed was ordered by the sort parent.
 			if($nodeID !== $rowParentID)
 			{
-				break;
+				// workaround in case the rows are not in the expected order
+				// for some reason e.g. if one sort column is not enough.
+				continue;
+				//break;
 			}
 
 			$node['_children'][] = &$row;
--- a/e107_plugins/forum/forum_admin.php
+++ b/e107_plugins/forum/forum_admin.php
@@ -137,7 +137,7 @@ if(!deftrue('OLD_FORUMADMIN'))
 		protected $fields 		= array (
 			'checkboxes'                =>   array ( 'title' => '', 'type' => null, 'data' => null, 'width' => '5%', 'thclass' => 'center', 'forced' => '1', 'class' => 'center', 'toggle' => 'e-multiselect',  ),
 		    'forum_id'                  =>   array ( 'title' => LAN_ID, 'data' => 'int', 'width' => '5%', 'help' => '', 'readParms' => '', 'writeParms' => '', 'class' => 'left', 'thclass' => 'left',  ),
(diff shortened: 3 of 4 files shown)


Deltik:
(╯°□°）╯︵ ┻━┻

I now realize that the problem is deeper than what your workarounds seemingly fix.

Because of how trees are constructed from relational databases, every record must be put into a tree structure before any sorting/ordering and pagination can be done.

This is why there is `e_tree_model::prepareSimulatedPagination()` and `e_tree_model::prepareSimulatedCustomOrdering()`.  Since code all over e107 may interact directly with SQL, those methods parse the query to extract the desired pagination and ordering.

What I completely neglected to factor in was searching/filtering, which is why all `e_tree_model` searches are broken now.

I guess I'll work on a `e_tree_model::prepareSimulatedFilter()` method.  This tree implementation has been quite a handful!

Until this is fixed at `e_tree_model`, #3198 and #3204 should remain open.

----

As for the `e_tree_model::moveRowsToTreeNodes()` workaround, it should be reverted because the following code in `e_tree_model::prepareSimulatedCustomOrdering()` guarantees the correct sort before constructing the tree:

```php
			// Optimization goes with e_tree_model::moveRowsToTreeNodes()
			. " ORDER BY " . $this->getParam('sort_parent');
```

Deltik:
@SimSync: I've made two pull requests that supersede this one:

#3209 fixes #3204
#3210 fixes #3198

I cherry-picked two of your commits to put into #3210 so you'll have credit for fixing #3198.

SimSync:
I can send you data from @jimako which shows that sorting on the sort_parent is probably not enough. 
Maybe adding the sort_field as second sort parameter would help to move in the right order.
The sort_parent field is not unique, so the records sorted by this field, within the same parent ID, are sorted somehow as the records have been added to the db. This is not necessarily the expected order, which than results in records not being displayed.
The forums I've created in sequence were displayed fine, but when I used the data from @jimako, I faced the same issues: 3 forums that didn't display. It even told me that 27 entries where there, but displayed only 24. Records per page is 30...
When I stepped thru the code step by step I saw that it ignored the remaining data because that test on the parent ID failed. That's why I switched to continue.

Deltik:
@SimSync: I was able to reproduce the issue from the information you provided.  I made younger IDs the parent of older IDs, and suddenly, the tree lost its data.

I'll account for this in (TreeModelTest.php](https://github.com/e107inc/e107-test/blob/master/tests/unit/TreeModelTest.php), too.

Deltik:
@SimSync: I was able to reproduce the issue from the information you provided.  I made younger IDs the parent of older IDs, and suddenly, the tree lost its data.

I'll account for this in [TreeModelTest.php](https://github.com/e107inc/e107-test/blob/master/tests/unit/TreeModelTest.php), too.

Deltik:
@SimSync: Should be fixed in b471626fdb8d95252b33ffe8b4037c75eb7012ca

Deltik:
@SimSync or @CaMer0n: If all looks well with #3198 and #3204, this issue can be closed.

PR #3209, opened 2018-06-22
Disable tree format if model has filters

Deltik:
Parent-child relationship tables (trees) can only be constructed if the entire table is loaded. When WHERE filters are applied to the table, the tree cannot be made, which causes #3204.

When a filter is active, a tree representation is no longer appropriate because the filter can filter out parents of a matched row, which breaks the concept of a tree.

To resolve this and restore search functionality, this commit disables the `sort_parent` of the tree model in order to activate the `e_tree_model::getRowsList()` method and avoid the tree logic entirely.

Fixes: #3204
Supersedes: #3208


CaMer0n:
Thank you @Deltik !!

Issue #2557, opened 2017-04-10
Block Certain Email Domains from Registering


BigBoiJefe:
There's some email domains that I would like to block from registering.  Thanks!

LaocheXe:
This is a good idea for an enhancement

BigBoiJefe:
Looks like there's already a huge list of disposable email domains:

`https://github.com/martenson/disposable-email-domains/blob/master/disposable_email_blacklist.conf`

LaocheXe:
If I recall, the stop forum spam site had a lis5 as well, and the gas plug8n uses such a list which also blocked ema8ls, but not email domains

Moc:
Yeah, I don't think this is core area territory. Plugins such as the SFS plugin are very effective in stopping spambots from registering on your website (other examples are VisualCaptcha or reCaptcha).

CaMer0n:
I was sure we had something built-in for this at some point. Can't find it now though.

BigBoiJefe:
Maybe I wasn't clear what I was looking for?  What I'm talking about is blocking disposable domains from registering on your site.  Disposable domains aren't necessarily used my spambots.  I dunno ;)

Moc:
In v1, this was done through the ban list... Not sure where it is located in v2 now.

CaMer0n:
@Moc Yeah, I found it in the meantime. I've just upgraded the banlist admin area so it saves the email addresses/hosts etc. No idea if the ipHandler still manages the ban correctly though, will need some testing.

BigBoiJefe:
So can I add @domain.com?

CaMer0n:
`*@domain.com`

CaMer0n:
@BigBoiJefe Did this work for you?

Moc:
@BigBoiJefe can you please verify that this feature works now?

BigBoiJefe:
I am sooo confused.  Where is the option to list *@domain.com

Moc:
In `Banlist > Add to blacklist` - `/e107_admin/banlist.php?mode=main&action=create`
First row: `Enter IP, email address, or host`

Can you confirm if it works?

BigBoiJefe:
If I add *@domain.com to the ban list and try to register a user account with an @domain.com email address, after I click the register button the page reloads and is blank.

SimSync:
@BigBoiJefe Could you please check if this issue is fixed?

SimSync:
[...]

PR #3223, opened 2018-06-27
Fixed an issue in iphandler::checkBan() 

SimSync:
Fixed an issue in iphandler::checkBan() which caused the script to stop without any message instead of returning a value (true or false) in case the submitted email adress is blocked.
That resulted in an empty page (e.g. during signup), leaving the user with no message of what happend.

### DIFF
--- a/e107_handlers/iphandler_class.php
+++ b/e107_handlers/iphandler_class.php
@@ -991,6 +991,11 @@ class eIPHandler
 				echo "\nBanned</pre>";
 			}
 
+			// added missing if clause
+			if ($do_return)
+			{
+				return false;
+			}
 			exit();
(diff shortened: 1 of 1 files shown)


Moc:
Does this relate to #2557?

SimSync:
@Moc Not directly. The ban (e.g. *@mail.com) worked already before. The issue here was, when somebody tried to signup using a banned email adress or email domain, the page after submitting was just white. Instead of giving an explanation that the email was banned (or other issues popped up that stopped the signup process).

SimSync:
@Moc Just see, that @BigBoiJefe posted "my" issue into the other issue thread in the meanwhile.

PR #3252, opened 2018-07-04
fixes #3162 double encoding broke url

### LINKED ISSUE #3162: problem with news tags -  not existings urls in WMT
I have tag with space:  "custom theme"
in tag menu (correct):  /news/tag/custom%20theme
in news detail is wrong path :  
`<a class="news-tag" href="/news/tag/custom%2520theme">custom theme</a>`
{NEWSTAGS} is used

in tag_menu:
`$url = e107::getUrl()->create('news/list/tag',array('tag'=>$word)); // SITEURL."news.php?tag=".$word;	`

in news_shortcodes:
`$url = e107::getUrl()->create('news/list/tag',array('tag'=>rawurlencode($val))); // e_BASE."news.php?tag=".$val`

SimSync:
The extra encoding of the tag value broke the resulting url, because the tag value will be encoded automatically during url creation.

### DIFF
--- a/e107_core/shortcodes/batch/news_shortcodes.php
+++ b/e107_core/shortcodes/batch/news_shortcodes.php
@@ -1153,7 +1153,9 @@ class news_shortcodes extends e_shortcode
 		{
 			if(trim($val))
 			{
-				$url = e107::getUrl()->create('news/list/tag',array('tag'=>rawurlencode($val))); // e_BASE."news.php?tag=".$val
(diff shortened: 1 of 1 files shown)


CaMer0n:
Thans, I think this already got merged in with a different pull-request?

SimSync:
I don't think so, but I will check it. I think there was a similar PR title regarding the shortcodes parameters, but this one is not related to it.

SimSync:
Close PR because it was already fixed ...

PR #3253, opened 2018-07-04
fixes #3059 Language list was empty, even if languages were available

### LINKED ISSUE #3059: Language list error
Not loading available language list.
Blank page.

lang

SimSync:
**Language list was empty, even if languages were available**
Looks like the array structure changed over the time without adapting this part of the sourcecode.

**Download and info link returned from languagepacks.xml contain wrong links**
For some reason, the urls in the "languagepacks.xml" contains for some reason malformed urls. The urls start with `https://github.comhttps://github.com` instead of just `https://github.com`.
@CaMer0n Implemented a workaround to fix this urls, but they need to get fixed in the "languagepacks.xml" as well.

### DIFF
--- a/e107_admin/lancheck.php
+++ b/e107_admin/lancheck.php
@@ -913,11 +913,22 @@ class lancheck
 				return false;
 			}
 
-			foreach($rawData['language'] as $val)
+			foreach($rawData['language'] as $key => $att)
 			{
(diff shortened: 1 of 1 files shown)


CaMer0n:
Thanks! Maybe github changed some things.. will look into it.

PR #3255, opened 2018-07-04
fixes #3049 Realname was always shown and marked required,

### LINKED ISSUE #3049: Real name is required problem
Hi!

In the admin area / settings / register settings, the real name is set hide, but on the profile settings page is always show and it's required. This is a bug or not?

Thx. (e107 v2.1.7, php 5.6.33)

SimSync:
Realname was always shown and marked required, even if it was configured differently.
The template needed to get modified to make this work and the {REALNAME} shortcode had to be rewritten as it didn't took the settings into account.
The customtitle field had a similar issue.

### DIFF
--- a/e107_core/templates/usersettings_template.php
+++ b/e107_core/templates/usersettings_template.php
@@ -46,7 +46,7 @@ $sc_style['PASSWORD_LEN']['pre'] 		= "<br /><span class='smalltext'>  (".LAN_USE
 $sc_style['PASSWORD_LEN']['post'] 		= " ".LAN_USER_79.")</span>";
 
 $sc_style['USERCLASSES']['pre'] 		= "<tr>
-											<td style='width:40%;vertical-align:top' class='forumheader3'>".LAN_USER_76.":".req(varset(e107::getPref('signup_option_class')))."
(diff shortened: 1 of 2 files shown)


e107_core/templates/usersettings_template.php
```diff
@@ -77,7 +77,7 @@
 $sc_style['LOGINNAME']['post'] 			= "</td></tr>\n";	
 	
 
-$sc_style['SIGNATURE']['pre']			= "<tr><td style='width:40%;vertical-align:top' class='forumheader3'>".LAN_USER_71.req(varset($pref['signup_option_signature']))."</td>
+$sc_style['SIGNATURE']['pre']			= "<tr><td style='width:40%;vertical-align:top' class='forumheader3'>".LAN_USER_71.req(varset(e107::getPref('signup_option_signature')))."</td>
 											<td style='width:60%' class='forumheader2'>";
```
CaMer0n: Not sure we need varset() if using getPref() ?

SimSync: The more I think about it, varset() can't even work, because the argument isn't a variable. I'll check it later.

SimSync:
@CaMer0n Removed varset(), which was useless together with getPref().

CaMer0n:
Thanks! :+1:

PR #3257, opened 2018-07-06
fixes #3065 Added a method to translate wildcards to mysql counterparts

### LINKED ISSUE #3065: Admin area IP address search yields no results
Hi, when I search for an IP in the admin user area, it displays "No results" however, I know they exist.  I'm tying to do a search for a single ip or a portion of an IP and find out how many users are duplicates

Thanks.

SimSync:
The new method replaces the common wildcards "*" and "?" with the mysql counterparts "%" and "_" right before the search word is used in the query.


CaMer0n:
Great ! :+1:

PR #3263, opened 2018-07-11
Limited allowed userclasses for 'post_script' to a smaller group:

SimSync:
**Limited allowed userclasses for 'post_script' to a smaller group:**
- No One, (Main-) Admins, Mods and custom userclasses
- Added a check on opening of the prefs page to make sure, that the set value is only one of the allowed classes. Otherwise, class "No One" will be set and saved.

**Replaced deprecated function calls with their successor...**


CaMer0n:
:+1:

PR #3268, opened 2018-07-12
Fixes #3264 Switched url generation to v2 method

### LINKED ISSUE #3264: {STATLINK} shortcode in forum is not updated
it doesn't work with forum with sef-url set ON  - missing e_PLUGIN_ABS
It doesn't return SEF-URL for stats link (e107::url('forum','stats'))

```
	function sc_statlink()
	{
		 return "<a href='".e_PLUGIN."forum/forum_stats.php'>".LAN_FORUM_0017."</a>\n";
	}
```

SimSync:
Shortcode still used the v1 method of url creattion (fixed string).
Switched to v2 url() method.


CaMer0n:
:+1:

PR #3274, opened 2018-07-12
Fixes #2913 Applied the same visual appearance as in admin/comment.php

### LINKED ISSUE #2913: confusing setting (visual) of comment setting
Comment manager:

 
image

Site preferencies:
image

I was looking for blue yes.

SimSync:
In comment.php the "Yes" setting was "highlighted", but in the prefs.php, the same setting wasn't highlighted and not to easy to read and understand.


CaMer0n:
:+1:

PR #3293, opened 2018-07-19
fixes #3285 No rank images displayed in backend and on profile page

### LINKED ISSUE #3285: Rank images problem
Something does not appear in the user's profile for rank images
adminrtank
rank_user

SimSync:
- Disabled inline editing of rank image
- Fixed display of rank image in user profile


CaMer0n:
:+1:

PR #3296, opened 2018-07-20
Reduce reliance on global var $db_ConnectionID

Deltik:
When one of 11 methods in `e_db_mysql` look for a PDO object in `e_db_mysql::$mySQLaccess`, `global $db_ConnectionID` can be `NULL` in some cases, so it's more reliable to get the singleton `e_db_mysql` and use the `e_db_mysql::$mySQLaccess` from that singleton.

This change still allows `global $db_ConnectionID` to be used, but if `$db_ConnectionID` is `NULL`, the singleton fallback is used.

Fixes: e107inc/e107-test#6

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,7 +3,7 @@ version: "2"
 plugins:
   csslint:
     enabled: false
-    duplication:
+  duplication:
     enabled: true
     checks:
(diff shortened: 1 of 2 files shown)


CaMer0n:
Excellent! Thanks :+1:

PR #3301, opened 2018-07-21
Removed problematic cache from e107plugin::getinfo()

Deltik:
Every known usage of `e107plugin::getinfo()` requires up-to-date knowledge of the involved plugin's record in the database.

For this reason, I removed the object scope global state caching of results from `e107plugin::getinfo()`.

Also changed the method name to a more descriptive `getPluginRecord()`

Also made `e107plugin::getPluginRecord()` static

### DIFF
--- a/e107_admin/plugin.php
+++ b/e107_admin/plugin.php
@@ -631,7 +631,7 @@ class plugin_ui extends e_admin_ui
 
 				$eplug_addons = $plugin -> getAddons($eplug_folder);
 
-				$info = $plugin->getinfo($this->id);
+				$info = e107plugin::getPluginRecord($this->id);
 
 				$name = deftrue($info['plugin_name'],$info['plugin_name']). " v".$eplug_version. "({e_PLUGIN}".$info['plugin_path'].")";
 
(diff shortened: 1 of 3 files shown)


Deltik:
@CaMer0n: I can't reproduce.  How can I reproduce this, or at least, what is the error message?

CaMer0n:
@Deltik I just committed a change to thumb.php and added an acceptance test for thumbnail. 

I was getting a fatal error in thumb.php : null on prepare() statement in mysql. I thought it was the earlier mysql change, but it appears not.

CaMer0n:
Tested with: /thumb.php?src=e_PLUGIN%2Fgallery%2Fimages%2Fbutterfly.jpg&w=220&h=190

Deltik:
This is what I see when I reenable `error_reporting` for `thumb.php` at 8c2bf18679d081315d7e44c0d105b1dfb41ff7a3 using your test case:

screenshot_20180722_101255

Something else is going on, but I still don't have enough information to troubleshoot.

Deltik:
@CaMer0n: I can reproduce now that I've set up a barebones Ubuntu 18.04 virtual machine with a new e107 installation in it:
screenshot_20180722_110100

CaMer0n:
That's the one! Thank you! :+1:

PR #3306, opened 2018-07-22
Removed problematic cache from e107plugin::getinfo() and fixed regression in e_core_pref::__construct()

Deltik:
Reimplements #3301 and turns off "force" loading in the `e_core_pref` constructor, which caused a regression in `thumb.php`, which somehow depends on core preferences but mustn't connect to the database.  "Force" loading makes a query to the database.

### DIFF
--- a/e107_admin/plugin.php
+++ b/e107_admin/plugin.php
@@ -631,7 +631,7 @@ class plugin_ui extends e_admin_ui
 
 				$eplug_addons = $plugin -> getAddons($eplug_folder);
 
-				$info = $plugin->getinfo($this->id);
+				$info = e107plugin::getPluginRecord($this->id);
 
 				$name = deftrue($info['plugin_name'],$info['plugin_name']). " v".$eplug_version. "({e_PLUGIN}".$info['plugin_path'].")";
 
--- a/e107_handlers/plugin_class.php
+++ b/e107_handlers/plugin_class.php
@@ -1631,32 +1631,21 @@ class e107plugin
 	 * @param int $id
 	 * @return array plugin info
 	 */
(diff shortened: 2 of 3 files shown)


CaMer0n:
@Deltik oh, I see now. Thanks.

CaMer0n:
@Deltik The thumb.php issue is back. :/

image

And what's more surprising is that the thumbCest doesn't detect it now.

CaMer0n:
@Deltik Something must have become confused with the reverts. I removed that e_core_pref code, and it's working again. If you can think of more ways to detect a thumb.php error that codeception would pick up, please let me know. Thank you!

Deltik:
@CaMer0n: Why was #3306 reverted in 56d5bcafd762aa968eda133dbfe835040a018ba5?  [Line 848 of `e107_handlers/pref_class.php`](https://github.com/e107inc/e107/commit/56d5bcafd762aa968eda133dbfe835040a018ba5#diff-70ca995f0858b40132fe63a02e1b31eeL848) should read as follows:

```php
			$this->load();
```

The changes to `e107_admin/plugin.php` and `e107_handlers/plugin_class.php` should not be affecting the behavior of `thumb.php` and make the [tests in `e107inc/e107-test:tests/unit/pluginsTest.php`](https://github.com/e107inc/e107-test/blob/4320399ea7322fd8fa471a2765b0ddb57d5c5045/tests/unit/pluginsTest.php) pass.

As far as detecting errors in `thumb.php`, there seem to be an unknown number of different things that can go wrong to cause an error.  Tests are supposed to be reproducible, so coding in initial sets of conditions and testing for those are the only way we can catch regressions in `thumb.php`.

Issue #3259, opened 2018-07-07
system information on not boostrap theme


Jimmi08:
You can't click and close system messages on not boostrap theme

![image](https://user-images.githubusercontent.com/5429548/42412203-f8c0fa82-8207-11e8-9211-54a4819ef3aa.png)

Moc:
Does this happen on every non-bootstrap theme? If not, there may be a JS incompatibility

Jimmi08:
I did this:
-  update to latest github
- remove loading bootstrap from theme.xml in bootstrap3 theme
- but loading bootstrap css  (otherwise you don't see it)
- remove {ALERT} to get some message 
RESULT: you can't close this message because it depends on boostrap script

![image](https://user-images.githubusercontent.com/5429548/42586568-b588ec50-8538-11e8-8306-959747278890.png)

without bootstrap css:
![image](https://user-images.githubusercontent.com/5429548/42586657-e28a48b6-8538-11e8-968e-1c38d7fba493.png)

And I confirm next problem with tooltip: (yes, I removed tooltip from theme.php in bootstrap3  in footer-inline - otherwise this error is there 2x)
![image](https://user-images.githubusercontent.com/5429548/42586739-118c8b06-8539-11e8-86f3-11575feff7d0.png)

Jimmi08:
I found it:   issue about tooltip, it's related to need of bootstrap script
https://github.com/e107inc/e107/issues/3187

CaMer0n:
@Jimmi08 I believe a test for 'tooltip' needs to be added to the javascript, before the 'tooltip' function is called. @SimSync Does this one interest you?

SimSync:
I can look into it. But...
1. There are 2 issues: js and css as also the whole formatting is gone and you can not see a difference between different message types or even see that it is a message at all. To disable the function call to tooltip is the smallest issue.
2. Which result is desired? Just fix the js error or the styling issue or both.

@CaMer0n All those workarounds for non bootstrap themes are hurting the e107 source code...

CaMer0n:
@SimSync I agree about the source code. I think fixing the js to check for tooltip should be enough.

Jimmi08:
@SimSync tooltip error in console is away, but original issue is still there.

SimSync:
You have to include the bootstrap js to have the close functionality on messages.
Another question: Has the styling of the message changed? Please check if the backcompat.css file is loaded. This should be loaded automatically on legacy and non bootstrap themes.

[...]

PR #3310, opened 2018-07-22
possible fix for #3259: Added test if tooltip is loaded

SimSync:


CaMer0n:
:+1:

rica-carv:
Shouldn't this be a jquery issue rather than a e107 issue??

SimSync:
No, I don't think so. This is a call to a 3rd party function, which may not be available. To test if the function exists before calling it, is not the responsibility of jquery I think.

PR #3318, opened 2018-07-25
choose editor in forum and quick reply

SimSync:
New prefs for forum: editor & quickreply
editor makes it possible to choose between bbcode or tinymce (if installed and html enabled)
quickreply gives the option to choose between the standard textarea or the choosen editor (from pref editor)

### DIFF
--- a/e107_handlers/e107_class.php
+++ b/e107_handlers/e107_class.php
@@ -3627,16 +3627,37 @@ class e107
 
 	/**
 	 * Set or Retrieve WYSIWYG active status. (replaces constant  e_WYSIWYG)
-	 * @param null $val
+	 * @param bool $val if null, return current value, otherwise set value to registry
 	 * @return bool|mixed|void
 	 */
 	public static function wysiwyg($val=null)
 	{
-		
+		// Check the general wysiwyg setting
 		if (self::getPref('wysiwyg',false) != true)
 		{
 			return false; 	
 		}
+
+
+		if (defined('e_CURRENT_PLUGIN') && e_CURRENT_PLUGIN != '')
+		{
+			$editor = e107::getPlugPref(e_CURRENT_PLUGIN, 'editor', 'default');
+			if ($editor != 'default' && $editor != 'bbcode' && !e107::isInstalled($editor))
+			{
+				$editor = 'default';
+			}
+			switch ($editor)
+			{
+				case 'bbcode':
+					return false;
+				case 'tinymce4':
+					return true;
+				default:
+					break;
+			}
+
+		}
+
 
 		if(is_null($val))
 		{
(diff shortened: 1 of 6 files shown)


e107_handlers/e107_class.php
```diff
@@ -3627,17 +3627,38 @@ public static function minify($js,$options=array())
...
 	public static function wysiwyg($val=null)
 	{
-		
+		// Check the general wysiwyg setting
 		if (self::getPref('wysiwyg',false) != true)
 		{
 			return false; 	
 		}
 
+
```
CaMer0n: Thanks for this PR. I'm not a fan of this code here though.  Using e107::wysiwyg(false); in the forum plugin should be enough to disable it.

e107_plugins/forum/forum_admin.php
```diff
@@ -183,6 +183,9 @@ class forum_ui extends e_admin_ui
 			'popular'	            => array('title' => FORLAN_55, 'type'=>'number', 'data' => 'int','help'=>FORLAN_56),
 			'postspage'	            => array('title' => FORLAN_57, 'type'=>'number', 'data' => 'int','help'=>FORLAN_58),
 			'threadspage'	        => array('title' => FORLAN_186, 'type'=>'number', 'data' => 'int','help'=>FORLAN_187),
+
+			'editor'                => array('title' => 'Post editor', 'type' => 'dropdown', 'data' => 'str', 'readParms' => array(), 'writeParms' => array('optArray' => array(), 'defaultValue' => 'default'), 'help' => 'Which editor should be used to create/edit posts?'),
+			'quickreply'            => array('title' => 'Quick replay editor', 'type' => 'dropdown', 'data' => 'str', 'readParms' => array(), 'writeParms' => array('optArray' => array(), 'defaultValue' => 'default'), 'help' => '')
 		);
```
CaMer0n: I don't think there should be a pref for 'quick reply' . The whole idea behind quick-reply was that there is no formatting. (that's what makes it quick).  When formatting is required, then people should use the 'reply' button.  I'm guessing removing this will also remove the need to change line 405 of the template.

e107_handlers/e107_class.php
SimSync: That doesn't work. I tried it already. Look at the bbarea() function, where wysiwyg() is called and you see that any previous change will be overwritten.
Changing the wysiwyg(true) in bbarea to wysiwyg() is also a bad idea, because wysiwyg() is saving the last setting in the registry, which means that the bbarea will not be replaced by tinymce system wide when wysiwyg(false) is called anywhere in the code.
As a systemwide change is not what I want to achieve, I had to add some code that checks for the current plugin and to return a value without touching the global settings in the registry.

e107_plugins/forum/forum_admin.php
SimSync: I personally always modded the v1 forum to have bbcode in the quick reply field... And I could imagine, I was not alone. With this setting, the admin can decide if he/she wants a simple textarea or the chosen editor. Default is the textarea.

CaMer0n: okay :+1:

PR #3335, opened 2018-07-30
Update tinymce4/langs/en_GB.js

yesszus:
Include all e107 phrases.


Moc:
This PR also appears to change the spelling from British English (GB) to American English (US). 
For example: colour into color, and centre into center. I don't think that's intended, nor should it be. If you want American spelling, the en_US file should be used instead.

Moc:
@yesszus: don't get me wrong. Great changes, I'm just a bit picky on those minor other changes from GB to US :) 
Not sure why these were changed, e.g..by tinmyce themselves or by you, but just noticed it ..

PR #3341, opened 2018-08-03
fixes #2440 allows cyrillic (and other unicode characters) in loginname

### LINKED ISSUE #2440: Register in Cyrillic
It does not work if you enter the login registration in Cyrillic.

SimSync:


CaMer0n:
Thank you!! :+1:

PR #3346, opened 2018-08-04
tagcloud addon for gsitemap plugin

Jimmi08:
Added tagcloud addon for gsitemap plugin


CaMer0n:
Thanks @Jimmi08 !! :+1:  It's good, but it should be part of the 'news' plugin.

Issue #3344, opened 2018-08-04
discussion - possibility to "override" core addon


Jimmi08:
I try to explain what I need to solve with gsitemap addon:

- gsitemap plugin uses e_gsitemap addon  (example forum plugin)
- it's easy to add addon for news plugin or others to be able import them too.  

BUT. How to change them with custom plugin?  I don't want to change core, it's enough in most of cases, but with extended functionality I hit limits. 

Import array is creating this way:
```
		$addons = e107::getAddonConfig('e_gsitemap', null, 'import');

		foreach($addons as $plug => $config)
		{

			foreach($config as $row)
			{
				if(!in_array($row['name'], $existing))
				{
					$importArray[] = $row;
				}
			}

		}
```
$plug is plugin name - f.e. news. 

But I want with my plugin create new import array for news.  
Now I get double items, because I can't set something like that:
```
$importArray['news'] = array();
$importArray['myplugin'] = ... 
```
the better solution would be table name not plugin name for something like this.  

But then there is problem in what order are addons loaded.  I think it's in alphabetical order.  

It would be great if there is something like priority.  

Or this is too complicated and I should create separate import under custom plugin?

CaMer0n:
Why override it? The admin can simply choose which links to import to gsitemap from your canonical plugin instead of using the 'news' one.

Jimmi08:
Not practical, if you have a lot of links, add news, forum, download, everything doubled.  

Not mention if your download has the same name as news post - they will not be showed. 

Could at least have gsitemap table field active?  to not display links that shouldn't be in sitemap always in import list?

CaMer0n:
@Jimmi08 I believe we just need a 'check all' box at the top of the import table. (where it says "Import?"

![image](https://user-images.githubusercontent.com/2825679/43974658-170509fa-9c90-11e8-9316-d4a63d27b13d.png)

.. and then we need each plugin entry (sitelink, page, forum etc.),placed in its own tab,  so that each plugin has its own list with a "check all" box. 

Then you could simply check all your canonical links.

Jimmi08:
@CaMer0n too much work because just me.  Nobody else have any interest.  
There are more important things now. 
Thank you for your time.

Jimmi08:
This is fixed in core now, version 2.3.1 is needed

PR #3351, opened 2018-08-05
gsitemap addon for news plugin

Jimmi08:


CaMer0n:
Excellent work @Jimmi08 ! :+1:

Jimmi08:
Thanks. But now I need a solution how to get this out from core and replace by custom plugin.  There is no plugin name in that array :)

CaMer0n:
@Jimmi08 What would the custom plugin do differently?

Jimmi08:
I put example here: 

https://github.com/e107inc/e107/issues/3344

Issue #3336, opened 2018-07-30
threadMarkAsRead() - creates duplicates 


Jimmi08:
I think that there is mistypo here:

```
	function threadMarkAsRead($threadId)
	{
		global $currentUser;
		$_tmp = preg_split('#\,+#', $currentUser['user_plugin_forum_viewed']);
		if(!is_array($threadId)) { $threadId = array($threadId); }
		foreach($threadId as $tid)
		{
			$_tmp[] = (int)$tid;
		}
		$tmp = array_unique($_tmp);
		$viewed = trim(implode(',', $_tmp), ',');
		return e107::getDb()->update('user_extended', "user_plugin_forum_viewed = '{$viewed}' WHERE user_extended_id = ".USERID);
	}
```

![image](https://user-images.githubusercontent.com/5429548/43416345-4b09d1b4-9438-11e8-9553-785736af2ef3.png)

PR #3354, opened 2018-08-07
fixes #3338 fixed typo and e_SELF calling itself (infinite loop)

### LINKED ISSUE #3338: threadMarkAsRead() changes only last forum ID
I do this in plugin - marking more than one forum as viewed

```
if (e_QUERY == 'mark_all_as_read')
{
	unset($u_new);
	$qry = 'WHERE thread_lastpost > ' . $lvisit;
	if ($rows = $sql->retrieve('forum_thread', 'thread_id', $qry, true))
	{   
		foreach($rows as $forumrow)
		{               
			$forum->threadMarkAsRead($forumrow['thread_id']);
		}                                     
    e107::redirect(e_REQUEST_SELF);
		exit;
	}
}
```

Because $currentUser['user_plugin_forum_viewed'] is not refreshed after adding one thread (I don't know how to call this), it's still the same and so only last thread is changed

```
	function threadMarkAsRead($threadId)
	{
		global $currentUser;                 
		$_tmp = preg_split('#\,+#', $currentUser['user_plugin_forum_viewed']);
		if(!is_array($threadId)) { $threadId = array($threadId); }
		foreach($threadId as $tid)
		{
			$_tmp[] = (int)$tid;
		}                            
		$tmp = array_unique($_tmp);              
		$viewed = trim(implode(',', $tmp), ',');   
   
		return e107::getDb()->update('user_extended', "user_plugin_forum_viewed = '{$viewed}' WHERE user_extended_id = ".USERID);
	}
```

image

SimSync:
- Fixed a typo that caused the function threadMarkAsRead() not to work as expected.
- Replaced the redirect with e_SELF, with e107::url(), because e_SELF caused an infinite loop.

### DIFF
--- a/e107_plugins/forum/forum_class.php
+++ b/e107_plugins/forum/forum_class.php
@@ -1438,7 +1438,17 @@ class e107forum
 				$this->threadMarkAsRead($newIdList);
 			}
 		}
-		header('location:'.e_SELF);
+		//header('location:'.e_SELF);
+		// issue #3338: using e_SELF runs caused an infinite loop
+		if (empty($forum_id))
+		{
+			header('location: '.e107::url('forum', 'index'));
+		}
+		else
+		{
+			$forum_sef = e107::getDb()->retrieve('forum', 'forum_sef', 'WHERE forum_id='.$forum_id);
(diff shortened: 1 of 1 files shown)


Moc:
I think this fixes #3336 too right?

SimSync:
Yes

CaMer0n:
:+1:

PR #3357, opened 2018-08-07
fixes #3179 "Redirect to System not found page" didn't work

### LINKED ISSUE #3179: How 404 page in e107 works?
I thought this:

If I set Redirect to System not found page ON, it will go to file error.php and display content of this page.

But it doesn't work this way. Any 404 displays browser HTTP ERROR 404. 

But if I  set **Associate Root namespace to Page** - it will go to page.php and method processViewPage() will display page with this content.

```
		if(!$sql->gen($query))
		{
		 	header("HTTP/1.0 404 Not Found");
 
			// ---------- New (to replace values above) ----
			
			$this->page['page_title'] = LAN_PAGE_12;			// ***** CHANGED
			$this->page['sub_title'] = '';
			$this->page['page_text'] = LAN_PAGE_3;
			$this->page['comments'] = '';
			$this->page['rating'] = '';
			$this->page['np'] = '';
			$this->page['err'] = TRUE;
			$this->page['cachecontrol'] = false;
              }
```
`define("LAN_PAGE_12", "Invalid page");`
`define("LAN_PAGE_3", "Requested page does not exist");`

I use default e107 htaccess. There is this: 
```
# Custom  pages for php scripts only
<FilesMatch \.php$>
	ErrorDocument 400 /error.php?400
	ErrorDocument 401 /error.php?401
	ErrorDocument 403 /error.php?403
	ErrorDocument 404 /error.php?404
	ErrorDocument 500 /error.php?500
</FilesMatch>
```
I tried to use absolute path there and nothing. 

How this should work?

SimSync:
"Redirect to System not found page" didn't work because the response code of 404 made the redirect fail. It looks like, that response codes >= 400 are not supported by the header() function together with 'location:'.

### DIFF
--- a/e107_handlers/redirection_class.php
+++ b/e107_handlers/redirection_class.php
@@ -391,7 +391,8 @@ class redirection
 			header('Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0', true);
 			header('Expires: Sat, 26 Jul 1997 05:00:00 GMT', true); 
 		}
-		if(null === $http_response_code)
+		// issue #3179 redirect with response code >= 400 doesn't work. Only response codes below 400.
+		if(null === $http_response_code || $http_response_code >= 400)
 		{
 			header('Location: '.$url, $replace);
 		}


CaMer0n:
:+1:

PR #3365, opened 2018-08-09
fixes #3317 added bbcode template "forum"

### LINKED ISSUE #3317: bbcodes template for forum plugin
Forum post shortcode:

`{FORUM_POST_TEXTAREA}`

use this definition for rendering bbarea:

`return e107::getForm()->bbarea('post',$text,'forum');`

'forum' is template used for bbcodes.   This template doesn't exist. 

Only available templates are:
```
	  $temp['news'] 		= $BBCODE_TEMPLATE_NEWSPOST;
		$temp['submitnews']	= $BBCODE_TEMPLATE_SUBMITNEWS;
		$temp['extended']	= $BBCODE_TEMPLATE_NEWSPOST;
		$temp['admin']		= $BBCODE_TEMPLATE_ADMIN;
		$temp['mailout']	= $BBCODE_TEMPLATE_MAILOUT;
		$temp['page']		= $BBCODE_TEMPLATE_CPAGE;
		$temp['maintenance']= $BBCODE_TEMPLATE_ADMIN;
		$temp['comment'] 	= $BBCODE_TEMPLATE_COMMENT;
		$temp['signature'] 	= $BBCODE_TEMPLATE_SIGNATURE;
```

So for frontend default `$BBCODE_TEMPLATE` is used (the most restricted functionality).    

For admins `$BBCODE_TEMPLATE_ADMIN` should be used, because you can't post forum topic in admin area and your options are the same as quest then. So check for ADMIN_AREA is useless for forum plugin.

For bbcodes there should be the same rules as for tinymce - separate template for mainadmin, admin, member, quest. 

Because this is template, it's easy to customize this later if there is this need. 

No idea how this worked in version 1, but now at least 'submitnews' template should be used for forum topic. OR add $BBCODE_TEMPLATE_FORUM to templates.

SimSync:
- Added bbcode template "forum"
- Fixed issue with bbcode youtube to display the mediamanager only if the current user is an admin. Otherwise it will just add the plain bbcode string "[youtube]dkhkj[/youtube]" to the textarea.

### DIFF
--- a/e107_handlers/bbcode_handler.php
+++ b/e107_handlers/bbcode_handler.php
@@ -586,7 +586,7 @@ class e_bbcode
 			$BBCODE_TEMPLATE = $template;	
 			$template = 'comment';	
 		}
-		elseif(ADMIN_AREA)
+		elseif(deftrue('ADMIN_AREA'))
 		{
 			$BBCODE_TEMPLATE = $BBCODE_TEMPLATE_ADMIN;	
 		}
--- a/e107_handlers/bbcode_handler.php
+++ b/e107_handlers/bbcode_handler.php
@@ -547,7 +547,6 @@ class e_bbcode
 		$BBCODE_TEMPLATE_ADMIN = '';
 		$BBCODE_TEMPLATE_COMMENT = '';
 		$BBCODE_TEMPLATE_SIGNATURE = '';
-		$BBCODE_TEMPLATE_FORUM = '';
 
 
 		require(e107::coreTemplatePath('bbcode')); //correct way to load a core template.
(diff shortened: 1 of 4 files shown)


e107_handlers/bbcode_handler.php
```diff
@@ -574,7 +575,8 @@ function renderButtons($template='', $id='', $options=array())
 		$temp['maintenance']= $BBCODE_TEMPLATE_ADMIN;
 		$temp['comment'] 	= $BBCODE_TEMPLATE_COMMENT;
 		$temp['signature'] 	= $BBCODE_TEMPLATE_SIGNATURE;
-	
+		$temp['forum'] 	    = $BBCODE_TEMPLATE_FORUM;
+
```
CaMer0n: Here we need a v2.x template format and to load the forum template from the Forum plugin. (not the core). 
Let's discuss further on Gitter.

SimSync:
@CaMer0n I've updated my PR with a proposal of how to move the template to the plugin (not only for forum plugin).

e107_handlers/bbcode_handler.php
```diff
@@ -575,7 +574,20 @@ function renderButtons($template='', $id='', $options=array())
...
+			// and load the specific bbcode template from the plugin
+			// see forum plugin "templates/bbcode_template.php" for an example of the definition
+			$tpl = e107::getTemplate($template, 'bbcode', $template);
+			if (!empty($tpl))
+			{
+				// If the plugin has a template defined for bbcode, add it to the list
+				$temp[$template] = $tpl;
+			}
+			unset($tpl);
+		}
```
CaMer0n: Exactly!!!  Nice work! :+1:

CaMer0n:
:+1:

e107_handlers/bbcode_handler.php
Jimmi08: You removed e_bb.php functionality with this
https://github.com/e107inc/e107/issues/3394

PR #3367, opened 2018-08-09
fixes #3171 moved query param p to the options/query parameter of url()

### LINKED ISSUE #3171: forum_viewtopic.php - problem with e_QUERY and pagination
With SEF-URL forum.

First page (no parameter page)
**forum_viewtopic.php**
example: 
URL:  /forum/subforum/35/forum-topic
- e_PAGE is not set (so no possible test like with news if e_PAGE = forum_viewtopic.php (minor issue)
- e_LEGACY_URL is available so test can be with string:  
   e107_plugins/forum/forum_viewtopic.php?id=38&
- without pagination e_QUERY id=35&  
But with pagination:
` /forum/subforum/35/forum-topic/&p=2`
`e_QUERY id=35&&p=2`

So you can't use list($id,$p) = explode("&&", e_QUERY); - because it doesn't work for url without paging, and you can't use  list($id,$p) = explode("&", e_QUERY); because it doesn't work for paginated urls.

I think that problem is in e_QUERY, it shouldn't have first &  
Thanks

SimSync:
- Moved query param p to the options/query parameter of url()
- Implemented a fiy for accidentally double ampersand in the query part of the url

### DIFF
--- a/e107_handlers/application.php
+++ b/e107_handlers/application.php
@@ -160,11 +160,13 @@ class e_url
 				if($newLocation != $this->_request)
 				{
 					$redirect = e107::getParser()->replaceConstants($newLocation);
-					list($file,$query) = explode("?",$redirect,2);
+					list($file,$query) = explode("?", $redirect,2);
 
 					$get = array();
 					if(!empty($query))
 					{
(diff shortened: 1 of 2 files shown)


CaMer0n:
:+1:

PR #3370, opened 2018-08-10
fixes #3245 eHelper::title2sef() now strips any html or bbcode

### LINKED ISSUE #3245: bbcodes in forum titles and breacrumb issue
If you use bbcodes in forum title, it works  (images and emoticons):
image

But breadrumb title is not able to display emoticons:

image

And in links  it looks like this (the same in version 1)
image

Shoudln't be titles in breadcrumbs displayed just text (remove bbcodes?)

Hm, what then left there if there is no text, just bbcode?  Any idea how to get titles this way?

SimSync:
- eHelper::title2sef() now strips any html or bbcode
- forum_admin.php: make sure to strip bbcode and html from forum_name
- forum_admin.php: make sure forum_sef doesn't contain bbcode or html and is properly formatted
- English_admin.php: Added FORLAN_223 = forum_name help text

### DIFF
--- a/e107_handlers/application.php
+++ b/e107_handlers/application.php
@@ -4690,6 +4690,9 @@ class eHelper
 
 		$tp = e107::getParser();
 
+		// issue #3245: strip all html and bbcode before processing
+		$title = $tp->toText($title);
+
 		$title = $tp->toASCII($title);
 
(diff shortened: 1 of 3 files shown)


CaMer0n:
:+1:

PR #3381, opened 2018-08-15
fixes #3380 fixed the urls to the media manager used by the bbarea

### LINKED ISSUE #3380: url to media manager is wrong in bbarea
The url used by the bbarea (in admin area) to open the media manager points to the wrong folder, in case e107 is installed in a subfolder.

SimSync:


CaMer0n:
:+1:

PR #3383, opened 2018-08-16
prevent media manager to insert an empty img tag

SimSync:
- fixed an issue which prevents the media manager to insert an empty img tag.
- probably fixed an issue, which caused the selected image to be saved directly to the target textarea without clicking the "Save" button


CaMer0n:
:+1:

PR #3401, opened 2018-08-26
fixes #3394 Moved loading of e_bb after the loading of the templates

### LINKED ISSUE #3394: bbcode template on forum - removed custom bbcodes
Latest changes in forum template vs bbcodes - not all, just this one:

> fixes #3317 moved the forum template from core to plugin

https://github.com/e107inc/e107/pull/3365/commits/c1bec2b6d75445354028a858f98f0c51f5c69a41

caused that e_bb.php doesn't work and custom bbcodes are not displayed.

The reason is using $tpl = e107::getTemplate($template, 'bbcode', $template);
instead $BBCODE_TEMPLATE_FORUM;

SimSync:
e_bb was loaded before the final template was assigned, which eventually overwrote the e_bb contents.
Now this works again, but the only template that can be manipulated is `$BBCODE_TEMPLATE`, because this is the name of the final template var that get's assigned. All others will be ignored.

### DIFF
--- a/e107_handlers/bbcode_handler.php
+++ b/e107_handlers/bbcode_handler.php
@@ -551,18 +551,18 @@ class e_bbcode
 
 		require(e107::coreTemplatePath('bbcode')); //correct way to load a core template.
 
-		$pref = e107::getPref('e_bb_list');
-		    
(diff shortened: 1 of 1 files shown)


CaMer0n:
:+1:

PR #3403, opened 2018-08-27
The e_parse_class::toForm() didn't convert &039; back to '

SimSync:
see also https://github.com/e107inc/e107-test/pull/12

### DIFF
--- a/e107_handlers/e_parse_class.php
+++ b/e107_handlers/e_parse_class.php
@@ -814,8 +814,8 @@ class e_parse extends e_parser
 		}
 	//	return htmlentities($text);
 
-		$search = array('&#036;', '&quot;', '<', '>', '+');
-		$replace = array('$', '"', '&lt;', '&gt;', '%2B');
+		$search = array('&#036;', '&quot;', '<', '>', '+', '&#039;');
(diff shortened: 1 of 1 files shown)


CaMer0n:
Are you sure this is not deliberate? I imagine HTML inside single quotes could break?

ie. `$text = '<textarea>'.$tp->toForm($db_html).'</textarea>' `

Also, I'm not aware of any 'bug symptoms' associated with this very old method. Are you?

SimSync:
I don't have issues with it, except:
This came up due to your `testToForm()` method, because the `toDB()` method used there, converts `'` to `&039;`, but `toForm()` doesn't convert the converted string back to it's original (if it's not surrounded by `[html][/html]`).

CaMer0n:
Thanks. Let's hold off for now.

SimSync:
I thought once again about it and my expectation was wrong.
I expected to get the same string back that was given as input. But the `toForm()` method prepares the string to be used in form controls, not to return the original "raw" string that was entered by the user.
To get the raw input, i think i would have to use `html_entity_decode()` or `htmlspecialchars_decode()`.
So, that means my change on `toForm()` is wrong and the expected string should be with `&039;` instead of `'`.
I close this PR for now.

CaMer0n:
:+1:

PR #3404, opened 2018-08-27
Fixes #3375 Added GDPR Agreement checkbox to contact form (incl. Lans)

### LINKED ISSUE #3375: add field on form contact
in the new GDPR 
Is very important to add a field "checkbox" in form "contact.php"  because the user must select it to explicitly authorize privacy rules.
could you add it?
thank's

SimSync:
This PR includes:
- Added update routine from 2.1.9 to 2.2.0 (because of a new pref).
- Added new pref for a privacy policy url in contact forms (incl. Lans) 
- Added GDPR Agreement checkbox to contact form (incl. Lans)

### DIFF
--- a/e107_admin/update_routines.php
+++ b/e107_admin/update_routines.php
@@ -122,8 +122,6 @@ if (!$dont_check_update)
 	$LAN_UPDATE_4 = deftrue('LAN_UPDATE_4',"Update from [x] to [y]"); // in case language-pack hasn't been upgraded.
 	$LAN_UPDATE_5 = deftrue('LAN_UPDATE_5', "Core database structure");
 
-	$dbupdate['219_to_220'] = array('master'=>false, 'title'=> e107::getParser()->lanVars($LAN_UPDATE_4, array('2.1.9','2.2.0')), 'message'=> null, 'hide_when_complete'=>true);
-
 	$dbupdate['218_to_219'] = array('master'=>false, 'title'=> e107::getParser()->lanVars($LAN_UPDATE_4, array('2.1.8','2.1.9')), 'message'=> null, 'hide_when_complete'=>true);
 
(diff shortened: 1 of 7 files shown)


e107_admin/update_routines.php
```diff
@@ -564,6 +566,21 @@ function update_core_database($type = '')
...
+		// add new pref contact_privacypolicy
+		if (!e107::getPref('contact_privacypolicy', ''))
+		{
+			e107::getConfig()->setPref('contact_privacypolicy', '')->save(false, true);
+		}
+
+		return $just_check;
+	}
+
+
```
CaMer0n: We don't actually need this, we just add the pref to default.xml and e107 will do the rest.

CaMer0n:
You're reading my mind with this change. :+1:  I believe we need a pref for 'terms and conditions' and another for 'privacy policy'. 

https://www.econsultancy.com/blog/69253-gdpr-10-examples-of-best-practice-ux-for-obtaining-marketing-consent

SimSync:
What about adding a new pref section "GDPR" with (currently) 2 prefs:
- url to privacy policy
- url to terms and conditions 

Than I would move the privacy policy from contact to the new section and we can put all GDPR related settings that may come in the new section.

SimSync:
Just implemented a new preferences section for GPDR related settings.
Moved "Privacy Policy" setting to the new section and added pref for "Terms and conditions"
Shortcode for comment form adapted to new pref name.

CaMer0n:
Looks great! :+1:  My only issue is that some people may have no idea what GDPR is. Perhaps "Legal" or something similar would be a better pref section title?

SimSync:
I would never look for "Legal" if i search for GDPR settings ....
Maybe we should include a short introduction about GDPR including a link to wikipedia about this term. This way all languages are "supported".

CaMer0n:
@SimSync I realized there's just one issue with the notice on the contact page. The data isn't stored, it is simply emailed to the admin. I wonder if it is still required?

PR #3416, opened 2018-09-04
Fix ampersand handling in the legacy rating method

newbthenewbd:
There seems to be a problem with `e_QUERY` such that all the query ampersands (`&`) become `&amp;`s. It gets temporary fixes in many places, such as [the line 2946](https://github.com/e107inc/e107/blob/master/e107_admin/users.php#L2946) of `e107_admin/users.php`.

However, the legacy method contained in `rate.php`, still used by some of the otherwise working plugins, has never been patched, and possibly redirects users to URLs that don't work. This pull request aims to change that, hopefully for the better.


CaMer0n:
Thank you! :+1:

PR #3449, opened 2018-09-23
Fixes #3437 – e_form::inlineToken() performance

### LINKED ISSUE #3437: Pages / menus of the admin area has been too slow
Hi!
Since I updated to the latest github version (Sep 12, 2018), the pages / menus of the admin area has been too slow. I mean the loading time is almost 20-30 sec, while the rest of the admin area (like settings, or other plugins) is loading immediately. But just in the admin area... :\

Deltik:
This "inline token" is generated 30 times in my test, but it's the same `session_id()` being hashed. This is wasteful and can be mitigated in two ways:

* Reducing the time cost like so: `return password_hash(session_id(), PASSWORD_DEFAULT, ['cost' => 04]);`
* Storing the hash as an instance variable the first time it's generated

This pull request applies both mitigations.

### DIFF
--- a/e107_handlers/form_handler.php
+++ b/e107_handlers/form_handler.php
@@ -67,7 +67,7 @@ class e_form
 	protected $_tabindex_enabled = true;
 	protected $_cached_attributes = array();
 	protected $_field_warnings = array();
-
+    protected $_inline_token = null;
 
 	/**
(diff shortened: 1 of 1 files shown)


CaMer0n:
Thank you!!! :+1:

PR #3451, opened 2018-09-23
Fixes #3450 – SITELINKS_ALT parent button fix

### LINKED ISSUE #3450: sitelinks_alt: e_parse::replaceConstants() is not run on parent links
If a theme uses the `{SITELINKS_ALT}` shortcode, the generated text from `sitelinks_alt::sitelinks_alt_shortcode()` does not replace constants with `e_parse::replaceConstants()` on the `link_button` value, which results in broken images like this:

Theme with SITELINKS_ALT showing no parent button

Expected output:

Theme with SITELINKS_ALT showing expected output

The database table `e107_links` looks like this:

e107_links database table

In the example above, the resulting button image URL for the "General" link is as follows:

```
/e107_images/icons/{e_IMAGE}icons/icon2.png
```

Deltik:
Parent links in `sitelinks_alt::sitelinks_alt_shortcode()` now have their button image URI parsed by `e_parse::replaceConstants()`, bringing it into consistency with the children, which are already parsed the same way.

There is no corresponding test for this in [e107-test](https://github.com/e107inc/e107-test) because hard-coded dependencies in `sitelinks_alt::sitelinks_alt_shortcode()` are difficult to monkey-patch without crippling the performance of e107-test.

### DIFF
--- a/e107_core/shortcodes/single/sitelinks_alt.php
+++ b/e107_core/shortcodes/single/sitelinks_alt.php
@@ -68,7 +68,11 @@ class sitelinks_alt
 			}
 			else
 			{
-				$link_icon = $lk['link_button'] ? e_IMAGE_ABS.'icons/'.$lk['link_button'] : $icon;
+				$link_icon = $lk['link_button']
+					? (($lk['link_button'][0] == "{")
+						? $tp->replaceConstants($lk['link_button'],'abs')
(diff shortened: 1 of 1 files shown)


CaMer0n:
Nice! Thanks. :+1:

PR #3471, opened 2018-09-27
Matching scope for e_form::$_inline_token

Deltik:
Should be private to match `private function inlineToken()`


CaMer0n:
Thanks :+1:

PR #3482, opened 2018-09-30
FIX check if user has moderator permissions for the thread/forum

phibel:
Without admin permissions (member of user class 254) it was not possible
to delete a post in the forum. This fix add the threadId to the ajax query
and fetchs the forum-moderator for the thread which will then checked
against the user permissions/classes.

I added also an additional error message, if something goes wrong.


CaMer0n:
Excellent! Thank you @phibel !

Issue #3493, opened 2018-10-06
LAN variable nedded in page e107_plugins/page/e_search.


net1313:
Hard coded   
$search = array(
			'name'			=> **"Pages"**,

For multilingual sites it's better to replace "Pages" to LAN variable

PR #3495, opened 2018-10-07
Update English.php

yesszus:
Issue: #3493

### DIFF
--- a/e107_languages/English/English.php
+++ b/e107_languages/English/English.php
@@ -142,5 +142,4 @@ define("LAN_SUMMARY", "Summary");  // TODO   more files use summary replace
 define("LAN_REQUIRED_BLANK", "Required field(s) were left blank.");
 define("LAN_PLEASEWAIT", "Please Wait");
 define("LAN_CHOOSE_FILE", "Choose a file");
-define("LAN_PAGES", "Pages");
 ?>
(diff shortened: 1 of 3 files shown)


tgtje:
@yesszus  would it not be better to change the English define in the plugin itself? , instead of making it site wide ? Pages is also in use in social... and as word also in e_sitelink in /page.
Just to not 'confuse' the usage of the word 'Pages'.  so it becomes more of a 'function use' just for search only?

e107_languages/English/English.php
```diff
@@ -142,5 +142,5 @@
 define("LAN_REQUIRED_BLANK", "Required field(s) were left blank.");
 define("LAN_PLEASEWAIT", "Please Wait");
 define("LAN_CHOOSE_FILE", "Choose a file");
-
+define("LAN_PAGES", "Pages");
```
CaMer0n: Not here. Should be in e107_plugins/page/languages/English/English_global.php
```
define("LAN_PLUGIN_PAGE_NAME", "Pages");
```

yesszus:
@CaMer0n changed this.

tgtje:
👍  think this way is better ( eg. in overall 'understanding' of a Title versus a Result)
Nice work anyway

CaMer0n:
Thank you @yesszus @tgtje

PR #3499, opened 2018-10-13
Update tinymce4/langs/en_GB.js

yesszus:
Include all e107 phrases.


CaMer0n:
Won't this be overwritten as soon as TinyMce is upgraded?

yesszus:
@CaMer0n Only imported files in e107_plugins / tinymce4 / wysiwyg.php all e107 phrases and removed some non used outdated.

PR #3510, opened 2018-10-18
display correct version in message - Optimise PHP version check for G…

Jimmi08:
…ithub sync

related:  https://github.com/e107inc/e107/commit/94adbd06ccec4187d34a097742f5697fb577d53e


Moc:
Thanks, I forgot that one

PR #3515, opened 2018-10-24
Do not trim language editor translation whitespaces

newbthenewbd:
Currently, when there are whitespaces at the ends of translations, e.g. `define("LAN_ONLINE_1", "Guests: ");`, they get trimmed in the language editor. This causes two kinds of problems: one about the original translations not getting accurately presented when they actually do contain whitespaces at the trimmed location, and the other about the translations added by translators getting the whitespaces removed the second time they are saved.

As far as I'm aware, this issue might be the cause of [that](https://github.com/e107inc/e107/blob/3a74aea669f41abf11f1270e1439443af65b1454/e107_plugins/online/languages/English.php#L22) [discrepancy](https://github.com/e107translations/Polish/blob/dc5d73fef5a8bffe436fc046a6fea5573d8fb3c7/e107_plugins/online/languages/Polish.php#L16) in the Polish e107 language pack, and sure slowed down my ~still awfully slow~ efforts in order to get the pack up to date somewhat. :slightly_smiling_face:

This change should hopefully fix that.

### DIFF
--- a/e107_admin/lancheck.php
+++ b/e107_admin/lancheck.php
@@ -1922,7 +1922,7 @@ class lancheck
 			$retloc[$type][$locale[1]]= $locale[2];	
 		}
 				
-		if(preg_match_all('/^\s*?define\s*?\(\s*?(\'|\")([\w]+)(\'|\")\s*?,\s*?(\'|\")([\s\S]*?)\s*?(\'|\")\s*?\)\s*?;/im',$data,$matches))
+		if(preg_match_all('/^\s*?define\s*?\(\s*?(\'|\")([\w]+)(\'|\")\s*?,\s*?(\'|\")([\s\S]*?)(\'|\")\s*?\)\s*?;/im',$data,$matches))
 		{
 			$def = $matches[2];
 			$values = $matches[5];	


CaMer0n:
Thank you @newbthenewbd ! Typically the white-space should be added in the PHP script, and not in the LAN_XXX, which is why it trims it by default.  So, we need to correct it on the other side, not in the LAN file.

rica-carv:
> 
> 
> Thank you @newbthenewbd ! Typically the white-space should be added in the PHP script, and not in the LAN_XXX, which is why it trims it by default. So, we need to correct it on the other side, not in the LAN file.

@CaMer0n Ouch! You're talking adding whitespaces when there's concatenated LANS? In all files? That's a TITANIC job.... :ship:

SimSync:
@CaMer0n What do you think of allowing 1 space at the beginning and at the end?
Or to use &nbsp; in such cases?

CaMer0n:
@SimSync I don't like it. Also, somewhat related, is that we shouldn't be using concatenated LANS to build sentences that contain variables between them, but rather use variables (like [x]) inside a single LAN.

SimSync:
@CaMer0n You are right. At least for new lans this should be the way it should be done.

Issue #3525, opened 2018-10-30
Admins and users rights.


net1313:
does someone able to explain me how privileges and user classes works. How to assign the rights to admins and users?
I'm a Main Admin, I've created forum theme which is visible only to Admins and Mods.

New user has the rights of Admins and Mods (granted by me - Main Admin). 
The issue - I can't  create posts in this topic, I can't even see this topic.  
I repeat - I'm a Main Admin. 
The issue - new user, assigned to Admins and Mods can't create posts in this topic. He can't even see this topic.

Is this normal that Main Admin can't see forum topic created by Main Admin (himself), but assigned only to Admins and Mods, access to it and post there. So my questions are:

1. How to solve this issue without going through MIT or Harvard?
2. Why Main Admin has no rights to his web-site?
3. How to grant users rights attribution only by assigning them to apropriate user class? It was very easy to make in old versions.

Jimmi08:
Check your user hierarchy. Main admin has to be parent of Admins and Mods.  (It happened to me, it was flat from some reason). But full support for this question.  

Maybe related:
https://github.com/e107inc/e107/issues/3457

net1313:
Thanks Jimmy. Of course with my knowledge of e107 I can solve this problem. But still it's a pity to spend 20 minutes to break my head by asking questions why I have to look for solutions to make a forum theme visible to me as the main administator. Why another user who was assigned administrator rights by me can not see the theme and make post there.

Imagine this is a new user who has just downloaded e107 wanting to create his website?

Jimmi08:
>  Imagine this is a new user who has just downloaded e107 wanting to create his website?
Welcome to my world :)  I am glad that somebody else is saying this too.  I can only support you, I am not able to solve this.

net1313:
The best one. Main Admin wants to delete his own post within forum topic. Got alert  when clicking delete: "Can't delete. you must be forum moderator" . Great.

tgtje:
Be carefull to not 'over-extend'the issue.
just for the test i used a 2.1.8 version to eliminate if other github versions interfere.

Forum is still/works in structure as v1 (by classes).

As in v1 (same applies for v2) Main Admin is able and can create the structure of the forum but has NO direct (if necessary or by trust) access to 'certain' amount of threads/topics etc..
Visualisize it as a home owner (with appartments).
The 'home is build', and each appartment is rented. (Imagine sub lease is possible).
[...]

Issue #3457, opened 2018-09-25
Forums with  Admins and Mods class can not be seen anymore


Alex-e107nl:
When you use classes in the forum plugin and you've got the right class it's still not showing, used 
Admins and Mods to hide it from the public but it's also been hidden from me :-)

(Tgtje see e107.nl forum..)

Alex

Jimmi08:
try to check user hierarchy.  I needed to move Admin and Mods under Main Admin. Or something like that.

Moc:
@Alex-e107nl Is this still an issue with the latest files?

Alex-e107nl:
Hi Moc, just updated with git from today on e107.nl it is solved!! 
THX You are the best (You-all!!)

Alex

Moc:
Thanks for the feedback!

PR #3530, opened 2018-11-01
part fix for correct main admin permissions  #3525

Jimmi08:
maybe for #3457 too

### DIFF
--- a/e107_plugins/forum/forum_class.php
+++ b/e107_plugins/forum/forum_class.php
@@ -1612,7 +1612,7 @@ class e107forum
 		{
 			return $this->modArray;
 		}
-		if($uclass == e_UC_ADMIN || trim($uclass) == '')
+		if($uclass == e_UC_MAINADMIN || trim($uclass) == '')
 		{
 			$sql->select('user', 'user_id, user_name','user_admin = 1 ORDER BY user_name ASC');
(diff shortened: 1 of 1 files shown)


Jimmi08:
Closing, because this is working solution, but it just looks weird:

```
	function forumGetMods($uclass = e_UC_ADMIN, $force=false)
	{
		$sql = e107::getDb();
		if(count($this->modArray) && !$force)
		{
			return $this->modArray;
		}
      
		if($uclass == e_UC_MAINADMIN || trim($uclass) == '')
		{    
			$sql->select('user', 'user_id, user_name','user_perms = "0" ORDER BY user_name ASC', false, true);
			while($row = $sql->fetch())
			{        
				$this->modArray[$row['user_id']] = $row;
			}
		}
		elseif($uclass == e_UC_ADMIN || trim($uclass) == '')
		{  
			$sql->select('user', 'user_id, user_name','user_admin = 1 ORDER BY user_name ASC');
			while($row = $sql->fetch())
			{        
				$this->modArray[$row['user_id']] = $row;
			}
		}    
		else
		{    
			$this->modArray = e107::getUserClass()->get_users_in_class($uclass, 'user_name', true);
		}

		return $this->modArray;
	}
```

Issue #3533, opened 2018-11-01
e_formTest::testRenderElement() fails due to differing execution environments


Deltik:
Ignoring that `e_formTest::testRenderElement()` should be broken into multiple tests, there are two known ways for the test to fail:

- [`templates_001`](https://github.com/e107inc/e107-test/blob/1fc01e08f0dc16b2fcb2ee750fac59e246626e9f/tests/unit/e_formTest.php#L842) fails because the output depends on how the operating system-dependent implementation of PHP `readdir()` sorts files in the file system.
  ```
  Expected :'<select name='templates_001' id='templates-001' class='tbox select form-control' tabindex='13'><option value='bbcode'>Bbcode</option><option value='forum_icons'>Forum Icons</option><option value='forum_poll'>Forum Poll</option><option value='forum_post'>Forum Post</option><option value='forum_posted'>Forum Posted</option><option value='forum_preview'>Forum Preview</option><option value='forum'>Forum</option><option value='forum_viewforum'>Forum Viewforum</option><option value='forum_viewtopic'>Forum Viewtopic</option><option value='newforumposts_menu'>Newforumposts Menu</option></select>'
  Actual   :'<select name='templates_001' id='templates-001' class='tbox select form-control' tabindex='13'><option value='forum_poll'>Forum Poll</option><option value='newforumposts_menu'>Newforumposts Menu</option><option value='forum_preview'>Forum Preview</option><option value='forum_post'>Forum Post</option><option value='forum_viewforum'>Forum Viewforum</option><option value='forum_posted'>Forum Posted</option><option value='forum_icons'>Forum Icons</option><option value='forum_viewtopic'>Forum Viewtopic</option><option value='bbcode'>Bbcode</option><option value='forum'>Forum</option></select>'
  ```
- [`date_001`](https://github.com/e107inc/e107-test/blob/1fc01e08f0dc16b2fcb2ee750fac59e246626e9f/tests/unit/e_formTest.php#L854) fails because the provided datestamp of "`2018-08-23`" is interpreted differently depending on the executing machine's locale.
  ```
  Expected :'<input class='e-date input-xlarge form-control' type='text' size='40' id='e-datepicker-date-001' value='Thursday, 23 Aug, 2018' data-date-unix ='true' data-date-format='DD, dd M, yyyy' data-date-ampm='false' data-date-language='en' data-date-firstday='0'    /><input type='hidden' name='date_001' id='date-001' value='1535007600' />'
[...]

PR #3534, opened 2018-11-01
Sorting for "templates" type e_form::renderElement()

Deltik:
The output from `e107::getFile()->get_files()` is now sorted in `e_form::renderElement()` for consistent output.

Fixes: #3533


CaMer0n:
:+1:

Issue #3536, opened 2018-11-02
If a plugin is downloaded by e_marketplace and installed by the same script, the installation skips the plugin addons.


Deltik:
When `e107plugin::install()` is called, it does not rescan the `e107_plugins` folder for any plugins that may have been added since `e107plugin` was instantiated.  New plugins would not be added into the database.

Since plugin addons are found from the database, any addons from the new plugin will be skipped during installation.  <sup><sub>This may cause troubleshooting headaches like it did for me.</sub></sup>

Plugins need to be registered in the database before installation.

PR #3537, opened 2018-11-02
Ensure plugin files are registered in database before installing any plugin

Deltik:
`e107::getPlug()->clearCache()` rescans the plugin folder for plugins and adds them to the database.

Fixes: #3536

### DIFF
--- a/e107_handlers/plugin_class.php
+++ b/e107_handlers/plugin_class.php
@@ -4148,15 +4148,19 @@ class e107plugin
 
 		$text = '';
 
+		e107::getPlug()->clearCache();
+
 		// install plugin ...
 		$plug = e107plugin::getPluginRecord($id);
-		
+
+		// XXX: The code below does not actually check if the plugin is in the database table.
 		if(!is_array($plug))
 		{
 			$message = $id." is missing from the plugin db table";
 			$this->log($message);
 			return $message;
 		}
+		// XXX: The code above does not actually check if the plugin is in the database table.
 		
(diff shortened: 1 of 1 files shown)


rica-carv:
What about modules, like there was in old V1.x?
Modules are simpler "plugins" that don't need install.
They are simply put in e107 plugins folder and are immediatly available....
Current e107 version can still work with them....

Deltik:
@rica-carv: I don't think I've ever used modules before, but this change should not affect their behavior.

What this pull request addresses is a specific broken flow in one page load:

- `e_plugin` initializes, scans the `e107_plugins` folder for plugins, and loads the discovered plugins into the database.
- Anything, like `e_marketplace`, extracts a new plugin into `e107_plugins`.
- Anything tries to install the plugin that was just extracted.

This flow is not normally encountered because extracting a new plugin and installing that plugin are usually two separate page loads, so `e_plugin` will initialize anew each page load.

[A broken sanity check](https://github.com/e107inc/e107/pull/3537/commits/7b259aec6fdc2c0972cf3ab4169a1dffe50b5c3d#diff-b4715a8381fe64fa8bf99bb98dfb5715R4157) means this flow is not caught and the plugin installation will proceed, but not everything will be loaded into the database.  Specifically, plugin addons will not be loaded.

This pull request fixes a test that fails [here](https://github.com/e107inc/e107-test/blob/7c2b4f8f25040986cc2d3d91a8d27c79c9db6b91/tests/unit/pluginsTest.php#L359).

PR #3552, opened 2018-11-16
Added event triggers

nesjett:
Some triggers to control when user rates something or user likes/dislikes something


CaMer0n:
Thank you! :+1:  I will tweak the names slightly for consistency.

PR #3584, opened 2018-12-20
forum search - possibility to style search button

Jimmi08:


e107_plugins/forum/shortcodes/batch/forum_shortcodes.php
```diff
@@ -164,14 +164,17 @@ function sc_userlist()
 		  return $text;
 	}
 
-	function sc_search()
+	/* example: {SEARCH: buttonclass=btn btn-fill btn-danger} */
+	function sc_search($parm = '')
```
CaMer0n: Did you test this in PHP7.2 ? I suggest $parm = null

Moc:
Committed in https://github.com/e107inc/e107/commit/6fd1aef610a8541891d90c19a4de7b5358f14f58
Tested in PHP 7.3

Jimmi08:
@Moc why you don't add examples in your changes in shortcodes?  They are really helpful and save time.   Others do it too.  Achim example below.  I can fix my ones similar way.  With them you just copy code to template, no need of manual writting of shortcode.  And you can see what is possible at one look without studying the code.

 
image

Moc:
I'll add them

PR #3600, opened 2019-01-06
Fix #3338 - marked more than one forum as read

### LINKED ISSUE #3338: threadMarkAsRead() changes only last forum ID

Jimmi08:
Because $currentUser['user_plugin_forum_viewed'] is not refreshed after adding one thread,  it's still the same and only last thread is changed

### DIFF
--- a/e107_plugins/forum/forum_class.php
+++ b/e107_plugins/forum/forum_class.php
@@ -1586,6 +1586,7 @@ class e107forum
 		$tmp = array_unique($_tmp);
 		// issue #3338 fixed typo, that caused issue with not marking threads are read
 		$viewed = trim(implode(',', $tmp), ',');
+		$currentUser['user_plugin_forum_viewed'] =  $viewed;
 		return e107::getDb()->update('user_extended', "user_plugin_forum_viewed = '{$viewed}' WHERE user_extended_id = ".USERID);
 	}
 


CaMer0n:
Thanks. :+1:

PR #3636, opened 2019-01-23
fixes #3632: Lancheck failed on some utf-8 values

### LINKED ISSUE #3632: Language verify question
Verify my language pack in localhost (windows 10 and xampp).
This is bug?

All errors: Non UTF-8 Characters Found

lan

php

SimSync:
the regex to extract the language defines and values missed the "u" (unicode) flag. Due to that, some utf-8 values were malformed during processing.


yesszus:
Great!
Work this without errors!

PR #3640, opened 2019-01-23
fixes #3639: Added description to GDPR prefs

### LINKED ISSUE #3639: Remove GDPR links on signup page if empty
Remove the GDPR links to "Privacy Policy" and "Terms and conditions" from signup page, in case they are empty.
Add an information about where the GDPR links are used to the GDPR preferences page.

SimSync:
Added check in shortcode to disable the GDPR message in case the prefs are empty

### DIFF
--- a/e107_languages/English/admin/lan_prefs.php
+++ b/e107_languages/English/admin/lan_prefs.php
@@ -300,3 +300,11 @@ define("PRFLAN_277", "GDPR Settings");
 define("PRFLAN_278", "URL to the Privacy Policy");
 define("PRFLAN_279", "Make sure the url exists! It's best to use an absolute url. This setting will be used on all places that require a consent from the user (e.g. signup, contact form/menu, etc.).");
 define("PRFLAN_280", "URL to the website terms and conditions");
+define("PRFLAN_281", <<<STR
+The 2 links above are used on various page on this site (e.g. signup and contact form/menu).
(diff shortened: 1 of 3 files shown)


e107_languages/English/admin/lan_prefs.php
```diff
@@ -300,3 +300,11 @@
...
 define("PRFLAN_279", "Make sure the url exists! It's best to use an absolute url. This setting will be used on all places that require a consent from the user (e.g. signup, contact form/menu, etc.).");
 define("PRFLAN_280", "URL to the website terms and conditions");
+define("PRFLAN_281", <<<STR
+The 2 links above are used on various page on this site (e.g. signup and contact form/menu).
+Please create 2 pages (if not already done) that contain your "Privacy Policy" and the websites "Terms and conditions". 
+There are several websites that can generate those text for you.
+Copy the urls of this websites into the fields above (e.g. /page/privacy-policy or /page/terms-and-conditions). 
+Just make sure, the pages and urls exist and are working!
+STR
+);
```
CaMer0n: I think the `<<<STR` will mess up the lancheck and other regex. I suggest simply using `\n` between lines and backslashes on the `"` .

SimSync: Yes, you are right. I'll modify it later.

PR #3642, opened 2019-01-24
fixes #3603: multiarray_sort returned only the last item in case the array keys are numeric

### LINKED ISSUE #3603: imageselector and multiarray_sort vs php 7
I noticed  that I can't use imageselector shortcode in php 7

Reason is this line in imageselector.php

`$icons = multiarray_sort($icons, 'fname');`

before it $icons is normal array, after it there is only last item of array

Arrays looks like this:
```
Array
(
    [0] => Array
        (
            [path] => ../../e107_plugins/forumthanks/user_images/loading//
            [fname] => default.gif
        )

    [1] => Array
        (
            [path] => ../../e107_plugins/forumthanks/user_images/loading//
            [fname] => Loading0.gif
        )

    [2] => Array
        (
            [path] => ../../e107_plugins/forumthanks/user_images/loading//
            [fname] => Loading1.gif
        )
....
```

SimSync:


CaMer0n:
Thanks. Would be good to have a test for this. (core_functionsTest.php)

PR #3643, opened 2019-01-24
fixes #3561: modified getTemplateInfo call to load info from templates

### LINKED ISSUE #3561: $NEWS_INFO is not able to be overridden in theme news template
If I put this into news/news_template.php, it works.
```
$NEWS_INFO = array(
	'section-card' 	=> array('title' => LAN_DEFAULT, 	'description' => 'unused'),
	'section-card-blog' 	    => array('title' => 'Section Blog Cards', 		'description' => 'Section Blog Cards'),
	'section-blog-horizontal'  => array('title' => 'Section Blog Horizontal',     'description' => 'Section Blog Horizontal'), 
);
```

As soon as I do it only in theme template, it doesn't work. 
But template itself is changed ($NEWS_TEMPLATE['section-card']['item']) after changing this in core template.

SimSync:
Modified getTemplateInfo call to load info from templates stored in the theme folder (theme/templates/news/news_template.php), in case that template is defined.

### DIFF
--- a/e107_admin/newspost.php
+++ b/e107_admin/newspost.php
@@ -1160,7 +1160,7 @@ class news_admin_ui extends e_admin_ui
 
 			$newsTemplates = array();
 
-			if($newInfo = e107::getTemplateInfo('news', 'news'))  //TODO  'category'=>'Categories'? research 'Use non-standard template for news layout' and integrate here.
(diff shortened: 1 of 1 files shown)


CaMer0n:
Great! Thank you!

PR #3664, opened 2019-02-03
fixes #3663: 1.custompages are "deleted" for the current selected layout

### LINKED ISSUE #3663: Theme layout custompages get removed on save
1. Unable to change the visibility filter for the current active layout. 
2. When changing the layout and saving it, the visibility filter for the new selected layout get's reset to "default".

SimSync:
1. The current layout displays now the entered visibility filters after save
2. Made the visibility filters now always editable, even the current selected layout.


CaMer0n:
Thank you!

Issue #3550, opened 2018-11-10
Can't send a private message


Evgura:
Stopped sending private messages on the site. Why is it that I'm the main administrator can't send a message?

![default](https://user-images.githubusercontent.com/25720089/48297462-75761600-e4ca-11e8-9adf-5f467909e69d.png)

Moc:
Did this happen suddenly? 
What are your settings in the PM plugin admin area?

Jimmi08:
I noticed this on e107.sk and update is sometimes when 2.2.0 was out (but not latest github).  I just haven't time to look at it. Mainadmin has problem on forum too, there is already issue about this.

Evgura:
The problem was on three sites, I don't think it was an accident. Version 2.1.9 on two sites and 2.2.0 on another.
Overwritten the plugin files from here and it worked.
https://github.com/e107-ru/e107-ru

Moc:
I have a feeling there's something weird going on with the userclass permissions. The forum seems to have permissions as well.

CaMer0n:
Bit of a mystery. The last real changes to PM happened in January 2018.

Jimmi08:
Admin works. Only Frontend:
![image](https://user-images.githubusercontent.com/5429548/48305096-f2ce7480-e524-11e8-9a2f-3b137d47c44a.png)

Moc:
@Evgura What's your PHP version?

Evgura:
@Moc Version PHP 7.0

Moc:
I suspect there's a PHP 7.x incompatibility issue. Will look into it asap.

CaMer0n:
@Jimmi08 Instead of using `print_a()`, try using `var_dump()` .

Evgura:
Updated today with Github and again the problem with personal messages.
PHP version-7.1.24

"System Information
You are not allowed to send PMs to: Maxim"


![](https://user-images.githubusercontent.com/25720089/48297462-75761600-e4ca-11e8-9adf-5f467909e69d.png)

Moc:
@Evgura Yes, no fix has been committed yet.

Evgura:
PHP:  5.6 Private Message work normally.

Moc:
Trying to reproduce this on PHP 7.1, but no luck yet. Clean install of e107, and clean install of the PM plugin. What are your settings here?: 

[...]

PR #3665, opened 2019-02-03
fixes an issue with pdoBind not being initialized properly

SimSync:
- fixes an issue with pdoBind not being initialized properly
- fixes an issue with update(), returning sometimes an object instead of
the number of updated records.
- fixes an issue with delete(), returning sometimes an object instead of
- the number of deleted records


Moc:
#3550

PR #3666, opened 2019-02-05
Fix a bug that failed to generate correct user profile links in chatbox_menu

arunshekher:
This PR fixes a bug that failed to generate correct user profile links in chatbox_menu view when legacy user link format was used.


CaMer0n:
Thank you @arunshekher !

Issue #3656, opened 2019-02-01
Forum TinyMCE is missing media manager icon


Moc:
Just taking notes; may close issue later if it's not a bug. 

Forum - Create new topic
TinyMCE is missing media manager options (image / video):

![image](https://user-images.githubusercontent.com/577978/52133675-1562bc00-2642-11e9-9827-1d4d0f43d992.png)

Example from Welcome message:

![image](https://user-images.githubusercontent.com/577978/52133815-6b376400-2642-11e9-9655-a32f6e3309e7.png)

SimSync:
This is because `mediaCat=_common`.
@CaMer0n I'm unsure what this mediaCat is about. If mediaCat is empty the media manager icons are available... Is it save to set mediaCat to an empty string?

SimSync:
I've looked into this issue a second time and there are 2 problems:
mediaCat and tinymce4 template

1. `mediaCat=_common` hides the buttons, regardless which template is used (`member.xml` for registered members, `public.xml` for guests). Setting `mediaCat=forum` would allow the media manager icons, but (and here comes problem 2) ...
2. For the forum area, neither of the tinymce4 templates has the mediamanager buttons defined. Which means, the buttons are missing in the "quick reply" an also on the normal new/edit post page.

My proposal woulld be as follows:
1. Set `mediaCat=forum` for both calls to bbarea in 'view_shortcodes.php' and 'post_shortcodes.php'.
2. Add the missing mediamanager icons to the `member.xml` template to have this buttons available for registered members. I think `public.xml` should be unchanged.

@Moc @CaMer0n What do you think? Is this the correct way to define/use the mediaCat setting?

CaMer0n:
@Moc @SimSync Members on the frontend should not have access to the admin's Media-Manager.

Moc:
@CaMer0n So there is no way to upload images to use inside the forum post itself, other than using externally hosted images, correct?

CaMer0n:
@Moc That's correct (as it should be), unless you use forum attachments.

Moc:
Alright. Closing this issue then.

PR #3671, opened 2019-02-12
Issue #3656: Tinymce shows Mediamanager icons on quick-reply

SimSync:
Tinymce was showing Mediamanager icons on forum quick-reply
Made sure, that the correct tinymce template is used for quick-reply an new post page


CaMer0n:
Thanks !

PR #3697, opened 2019-02-26
fixes e107inc/vstore#99: Difference between MariaDB and MySQL table def

SimSync:
Difference between MariaDB and MySQL on command TABLE CREATE returns CURRENT_TIMESTAMP w/o brackets
Updated method db_verify::diffStructurePermissive() to correct the given sql table definition.


CaMer0n:
Thank you!

PR #3707, opened 2019-03-02
fixes #3657: Userclass is reset on all user accounts to the one public userclass "Newsletter"

### LINKED ISSUE #3657: Userclass is reset on all user accounts to the one public userclass "Newsletter"
I have 56 members on my e107 site.  all have the default 'newsletter' class, with 2 admins and some forum moderators.

I changed my PHP from 7.1 to 5.6 due to another part of my website not working on 7.1 and it has changed every member in my list to ony 2 classes , 'newsletter'  and the new class I created manually.  it removed all forum mods and also admin classes from myself.

SimSync:
In case a user with user_class e_UC_NEWUSER logged in and was ready to be "promoted" to a "normal" user, the e_UC_NEWUSER should be removed. Unfortunately, the db:update() function was used in the
wrong way.
The third argument of update() function is for debugging purposes and NOT used for the WHERE clause.
Therefore the query was run without WHERE, which resulted into applying the new classes to all users....


CaMer0n:
Excellent find!! Thank you!

Moc:
Thanks @SimSync!

CaMer0n:
@SimSync @Moc Found a similar bad query in the ip handler class. ;-)

PR #3708, opened 2019-03-02
Switched add/remove user_class functions in e_user_model to make use of the unified functions

SimSync:
from the user_class class (user_class::ucAdd(), user_class::ucRemove())
Fixed some typo and removed missing template vars

### DIFF
--- a/e107_handlers/user_model.php
+++ b/e107_handlers/user_model.php
@@ -1062,15 +1062,18 @@ class e_user_model extends e_admin_model
 			return false;
 		}
 
-		$curClasses = explode(",", $this->getData('user_class'));
-		$curClasses[] = $userClassId;
-		$curClasses = array_unique($curClasses);
-
-		$insert = implode(",", $curClasses);
+//		$curClasses = explode(",", $this->getData('user_class'));
+//		$curClasses[] = $userClassId;
+//		$curClasses = array_unique($curClasses);
+//
+//		$insert = implode(",", $curClasses);
 
 		//FIXME - @SecretR - I'm missing something here with setCore() etc.
 	//	$this->setCore('user_class',$insert );
 	//	$this->saveDebug(false);
+
+		// Switched to unified remove user class method
+		$insert = e107::getUserClass()->ucAdd($userClassId, $this->getData('user_class'), false);
 
 		if(!$uid = $this->getData('user_id'))
 		{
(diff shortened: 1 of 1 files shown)


CaMer0n:
Thank you! I have created an empty test for you. 
https://github.com/e107inc/e107-test/commit/6bba285eeebebf0e5677ffa2cb9bac846d9390e1

CaMer0n:
@SimSync

SimSync:
@CaMer0n Done 😏

e107_handlers/user_model.php
```diff
@@ -1285,10 +1297,13 @@ public function renderEmail($type, $userInfo)
...
-			$EMAIL_TEMPLATE['quickadduser']['body']		= $QUICKADDUSER_TEMPLATE['email_body'];
-			$EMAIL_TEMPLATE['notify']['body']			= $NOTIFY_TEMPLATE['email_body'];
-			
+
+			// Both template vars are not defined...
+			//$EMAIL_TEMPLATE['quickadduser']['body']		= $QUICKADDUSER_TEMPLATE['email_body'];
+			//$EMAIL_TEMPLATE['notify']['body']			= $NOTIFY_TEMPLATE['email_body'];
+			$EMAIL_TEMPLATE['quickadduser']['body']	= '';
+			$EMAIL_TEMPLATE['notify']['body']			= '';
+
```
CaMer0n: Thank you!. Could you explain the thinking behind setting `$EMAIL_TEMPLATE['quickadduser']['body'] ` etc. to empty values?

SimSync: This was just a quick fix for the missing variables.
I did a search on the whole codebase and i couldn't find a var called `$QUICKADDUSER_TEMPLATE` or `$NOTIFY_TEMPLATE`. So instead making both unknown vars global to be able to use them in the method, i decided to just remove them by commenting out and setting an empty string instead.

Do you know, where to find those vars or what content to use instead for the `$EMAIL_TEMPLATE` var?

CaMer0n: I think it's a BC fix. We should probably check if `$QUICKADDUSER_TEMPLATE['email_body']` is not empty, and assign it to `$EMAIL_TEMPLATE['quickadduser']['body']` if it has a value. Similar with 'notify'

SimSync: I will make that missing vars global and use vartrue() for the assignment.
But first I must sort out my line endings issue... Don't want to screw up another file...

SimSync: @CaMer0n Pushed a new commit with the proposed changes to the 2 vars.

PR #3734, opened 2019-03-19
fixes e_form::checkboxes() not displaying labels properly

SimSync:
fixes e_form::checkboxes() not displaying labels properly that contain a '=' character. Also overwrote any supplied options.
The line endings seem to be screwed up, no matter what i do... 
This is a screenshot of my changes to that file...
grafik

### DIFF
--- a/e107_handlers/form_handler.php
+++ b/e107_handlers/form_handler.php
@@ -2576,8 +2576,12 @@ class e_form
 				$c = vartrue($checked[$k]);
 			}
 
-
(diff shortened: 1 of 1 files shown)


SimSync:
@CaMer0n @Deltik was so kind to help with this issue. For some reason, PhpStorm always changes the lineendings. Only after commiting and pushing using the console, Deltik was able to push the changes without screwing up the lineendings. It looks like, that this special file has a general problem...

Issue #3466, opened 2018-09-27
login shortcodes - easier templating


Jimmi08:
{LOGIN_TABLE_SUBMIT}

parameter class is missing

Jimmi08:
https://github.com/e107inc/e107/pull/3467

Jimmi08:
{LOGIN_TABLE_SIGNUP_LINK}

class is missing, parameter class is missing

Jimmi08:
{LOGIN_TABLE_FPW_LINK}
class is missing,  parameter class is missing, e_BASE used

CaMer0n:
Merged. Thanks.

PR #3747, opened 2019-03-22
#3466 {LOGIN_TABLE_USERNAME}  - added class

Jimmi08:


CaMer0n:
@Jimmi08 Could you tell me more about these Pull requests? What are you planning to substitute for 'class' ? Is it related to Bootstrap 4?

Jimmi08:
@CaMer0n there is hardcoded class 'tbox form-control input-block-level'
input-block-level was causing problems. 
This PR is only about the possibility to use a custom class if it's needed. In my case only form-control.  Otherwise, the original classes are used.

Issue #3750, opened 2019-03-23
fpw shortcodes - easier templating


Jimmi08:
{FPW_USEREMAIL}

Jimmi08:
{FPW_SUBMIT}

CaMer0n:
Closing. But please reopen if necessary.

PR #3751, opened 2019-03-23
#3750  {FPW_USEREMAIL} - added params

Jimmi08:


fpw.php
```diff
@@ -52,11 +52,19 @@ function sc_fpw_username($parm=null) // used when email login is disabled
...
 		// return "<input class='tbox' type='text' name='email' size='40' value='' maxlength='100' />";	
-		return e107::getForm()->email('email', '', 200, array('placeholder' => 'Email', 'required' => 'required')); 
+	  $options = array(
+	  'required' => true,
+      'placeholder' => vartrue($parm['placeholder']) ? $parm['placeholder']  : LAN_EMAIL,
+      'class' =>  vartrue($parm['class']) ? $parm['class']  : '',
+		);
+
+		return e107::getForm()->email('email', '', 200,  $options); 
 	}
```
CaMer0n: @Jimmi08 vartrue() shouldn't be used in an "IF" statement like this. (even if this issue does occur a lot in the core. I'm trying to clean it up) 
Here's the right way to do it: 


      'placeholder' => vartrue($parm['placeholder'], LAN_EMAIL),  // could break in PHP 7 under some circumstances 

OR

    'placeholder' => !empty($parm['placeholder']) ? $parm['placeholder']  : LAN_EMAIL,

Jimmi08: Thank you, I am sure I copied it from somewhere...   but I am not able to update this now.

Issue #3753, opened 2019-03-23
Extended User Fields Current Fields


LaocheXe:
I think we should update the current fields that are automatically added in the new field section. Some of them aren't needed anymore - such as the messenger apps. aim has been discontinue, also with yahoo address - unless that is for yahoo email? MSN was switched to windows messenger, which was migrated into skype.  If you want I can do a pull?

Moc:
Pull request welcome 👍

PR #3761, opened 2019-03-25
Removed Outdated User Extended Fields

LaocheXe:
Removed AIM, MSN, and Yahoo Messenger User Extended Fields #3753


CaMer0n:
Thanks :+1:

Issue #3768, opened 2019-03-29
db_verify::getPrevious() is broken in PHP 7


Deltik:
This failure was detected in regression testing.  In PHP 5.6, e107-test has this error:

```diff
There was 1 failure:

---------
1) db_verifyTest: Get fix query
 Test  tests/unit/db_verifyTest.php:testGetFixQuery
Failed asserting that two strings are equal.
- Expected | + Actual
@@ @@
-'ALTER TABLE `e107_table` ADD `table_auth` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0''
+'ALTER TABLE `e107_table` ADD `table_auth` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' AFTER table_ip'
#1  /builds/e107/e107-test/tests/unit/db_verifyTest.php:313
#2  db_verifyTest->testGetFixQuery
```

PHP 7.1, PHP 7.2, and PHP 7.3 have an actual value matching the expected.

It looks like the expected value is wrong and the actual value is the intended behavior from PHP 5.6 and earlier.

This is due to a bad implementation of `db_verify::getPrevious()` that breaks in PHP 7.

PR #3769, opened 2019-03-29
Improved portability of db_verify::getPrevious()

Deltik:
Now works in PHP 7 as it used to work in PHP 5

Fixes: #3768


CaMer0n:
Thank you ! :+1:

PR #3776, opened 2019-04-04
Removed Google+

LaocheXe:
Removed Google+ since it is shutdown.


LaocheXe:
Google+ has shut down, making this useless social link.

net1313:
Google+ is dead, but if you want to keep social plugin update and allow users login/register through Google account (eg. gmail) just replace a file Google.php in Providers directory of Hybridauth by this file
(https://github.com/hybridauth/hybridauth/blob/e7fb724aea4822b48e6d484e58ac8e541b205086/hybridauth/Hybrid/Providers/Google.php)
was successfully tested and works.

LaocheXe:
This just removed the options to add Google+ Profile/group link to be displayed on the frontend - has nothing to do with social login

CaMer0n:
Thanks @LaocheXe ! :+1:

Issue #2822, opened 2017-10-18
FAQs classic layout is broken


Jimmi08:
![image](https://user-images.githubusercontent.com/5429548/31709580-b6836580-b3f2-11e7-92d1-0acc456e6dbd.png)

![image](https://user-images.githubusercontent.com/5429548/31709593-bec4c694-b3f2-11e7-98e2-262bab4b60e6.png)
There are no clossing tags for this: 
![image](https://user-images.githubusercontent.com/5429548/31709605-c7b9b818-b3f2-11e7-84ec-698f6a2bce67.png)

PR #3848, opened 2019-06-14
missing shortcode category ID #2822

Jimmi08:


CaMer0n:
Thanks @Jimmi08 . Unfortunately there's too many different changes in this PR, to merge. The accordion code should be in a separate template, not inside shortcodes.

PR #3874, opened 2019-07-03
Added social login provider callback data to event

nesjett:
Event user_xup_signup was not giving info that may come from the social provider, for example the Birth date or gender, I just added a way to access all provider callback info from within the signup event.


CaMer0n:
Great! Thank you @nesjett !!

PR #3900, opened 2019-07-22
('WPLANG', '');

admin217:
('WPLANG', 'pt_BR');


tgtje:
NO such define/code does exist in e107 as it is part of 'wp'

Issue #3934, opened 2019-08-23
FR - using shortcode MENUAREA instead MENU


Jimmi08:
Using {MENU=9} is very confusing for new users.  {MENUAREA=9} is better term.

this issue is for this change
https://github.com/e107inc/e107/commit/68e4fd7d03421827a1e9b7687c830e65faa696f7

Tested with boostrap4  theme

It doesn't work if {MENUAREA} is used in _layout.html file.  
It works if {MENUAREA} is in theme.html  

The reason is that parseheader() in footer_default.php needs a single shortcode.  

I will post a quick fix to be able to test custom H+F but maybe there is another way.

Jimmi08:
Fixed by:

https://github.com/e107inc/e107/commit/49aac0b88a9001c6b94f96fa33e877e72be4fe25

PR #3935, opened 2019-08-23
#3934 added single shortcode menuarea

Jimmi08:
it is just a copy of menu  shortcode and renaming


CaMer0n:
Thanks. I had one, I just forgot to commit it. It's there now.

Issue #3930, opened 2019-08-20
Forum quickreply lan bug


akogo1:
In this part of the view_shortcodes.php file should be lan LAN_FORUM_2007 instead LAN_FORUM_2006.

`<div class='center text-center form-group'>
							<input type='submit' data-token='" . e_TOKEN . "' data-forum-insert='" . $ajaxInsert . "' data-forum-post='" . $this->var['thread_forum_id'] . "' data-forum-thread='" . $this->var['thread_id'] . "' data-forum-action='quickreply' name='reply' value='" . LAN_FORUM_2006 . "' class='btn btn-success button' />
							<input type='hidden' name='thread_id' value='" . $this->var['thread_id'] . "' />
						</div>`

tgtje:
Textual change happens on button (post) should indeed be 2007 = Post a quick reply where 2006 = Post  a reply.
Only for line 1241//   pull request created

PR #3937, opened 2019-08-25
Update view_shortcodes.php

tgtje:
LAN number change due to particular textual use (button text for QUICK reply (post command) )
LAN 2006 is general for Post a reply..


Moc:
#3930

Issue #3945, opened 2019-09-02
Feature Request: User Delete Option


LaocheXe:
Admin Area -> Users -> Options:
Allow users to delete account: enable/disable

(enable will show the delete button)
(disable will hide the delete button)

LaocheXe:
I don't know if you guys want this or not, but pull request has been made #3946

Moc:
It will be looked into asap :) thank you!

Jimmi08:
How your solution solves related user records? Extended fields, forum posts, chatbox etc.

Moc:
This PR just adds a preference that controls whether the delete button is shown or not. It does not change any of the current functionality.

CaMer0n:
Merged. Thank you! Closing.

PR #3946, opened 2019-09-02
Delete Account Optional

LaocheXe:
Added option in admin area to allow users to delete their account or not. In reference to #3945

Basically adds an option in Admin Area > Users > Options > Allow members to delete account? (yesno)
Hides the Delete button on usersettings.php (using shortcode)


Moc:
This will be a new preference right? In that case it also needs to be added to the default installation XML file

LaocheXe:
@Moc what file do I need to edit? I checked out the default_install.xml and notice none of the user options are listed there.

Moc:
That's indeed the file. It's weird that some of them are missing. Not all though, for example `memberlist_access` is there. 

@CaMer0n What are your thoughts on this?

CaMer0n:
@Moc Do you have a list of the preferences which are missing?

LaocheXe:
@CaMer0n want me to add those  to the install.xml as well, or is this good to go?

LaocheXe:
I'm going to close this pull request and apply a newer one with updated install.xml

Issue #3986, opened 2019-10-28
Strange issue creating custom menu


Serios:
Using latest v.2.2.1 I can't create custom menu to use in menu manager with my theme. Menu is actually created trough Custom pages/Menus but menu_name field is not saved to the DB even if I try to modify existing one the menu_name field gets blank and thus is not listed in Menu manager under custom menus. Menu is created with name, title, text and icon.

sudakk:
I can 't repeat this problem. The menu is created and renamed without problems. And it's visible in the Menu Manager.

Serios:
@sudakk on new install or on updated system?

sudakk:
@Serios on updated.

Jimmi08:
@Serios  2.2.1 from e107.org?  Check database validity and export one of the problematic records and put it here. If you use Chrome, use a different browser. 
My guess is that you have already menu name with the same name or it saved your username as menu title if you are using Chrome. 
2.2.1 doesn't have this problem, I would notice this, your data could be a problem. 
If it is not problem, you can export your cpage table to xml (Database/tools) and put it here, I could try to import it to my 2.2.1 version.

Serios:
@Jimmi08 @sudakk I managed to track the problem in relation with the php version of the server. On php ver 7.3 the problem exist. If php version is set to the minimum required 5.6 the problem no longer exist and menu_name is properly submitted to the DB. Confirmed it by trying both on my local server and on production server as well.

Jimmi08:
@sudakk try 7.2,  e107 officially support only this version,  I use only this one or 5.6.   I will need to install xampp with 7.3.   Yes, today I can :)

Serios:
@Jimmi08 It's working properly on 7.2 

You can setup your xampp to run different versions of php for different projects. If you are interested, you can check this [guide](https://stackoverflow.com/a/49586592)

Jimmi08:
Thanks, too complicated.  I know better and easier way :)  for somebody who is not server/administrator guy :)

Jimmi08:
Confirmed only for Chrome 
![image](https://user-images.githubusercontent.com/5429548/67679820-3adee500-f98a-11e9-9055-5671a76da7af.png)


Firefox:

![image](https://user-images.githubusercontent.com/5429548/67679770-24d12480-f98a-11e9-8fd8-f4a9fd75867e.png)

It's old not solved problem with prefilled fields in Chrome:
https://github.com/e107inc/e107/issues/3196
[...]

PR #3989, opened 2019-10-28
Update cpage.php

sudakk:
Added check to prevent saving page\menu with empty page_title & menu_name fields when edit.
Possible fix to [#3986](https://github.com/e107inc/e107/issues/3986)


CaMer0n:
Thanks @sudakk !

PR #3997, opened 2019-11-01
Some corrections

yesszus:
Add language codes.


Moc:
@yesszus thanks! I missed those

Issue #4040, opened 2019-12-05
News SEF URL's logic need great rewrite when using categories


Serios:
In an ideal world I think news url's when one is using categories, should look something like this:
website.com/news - Main news page, lists news items from all categories.
website.com/news/sport - Listing only news items from category "Sport"
website.com/news/sport/premier-league-predictions OR  website.com/news/premier-league-predictions OR website.com/news/1/premier-league-predictions (depending on URL settings) - Extended view of the news item.
Every other variant of the url's should return 404 page

However we currently have:
website.com/news - good, lists all news from all categories.
website.com/news/category/sport - Where this category thing came from? If we ignored for a bit this url actually works as expected, listing all news items from category "Sport". But website.com/news/sport works too and lists same things - duplicated content. Also the "category" part is hardcoded somewhere and one can't change it.

If we go further the thing goes weirder. If under URL's we set it like news/news-category/news-title  and view the extended news item, the URL goes website.com/news/sport/premier-league-predictions, but shouldn't it be website.com/news/category/sport/premier-league-predictions ?!? (if we try website.com/news/category/sport/premier-league-predictions it returns 404 page). If we follow the same schema it should. So, once we have website.com/news/category/sport and once we omit category keyword and go website.com/news/sport/premier-league-predictions. I think the "category" part should be omitted in both cases to come closer to "the ideal world" scenario.

Also if the news item have some related news their url's are generated without category, completely ignoring the setting of URL's, so we have website.com/news/sport/premier-league-teams in the category list and website.com/news/premier-league-teams listed in extended view as related news,  which are the same news but with different urls - duplicated content.

Moc:
Just did some testing and all news URL configurations seem to work as configured. 

>  Where this category thing came from?

This is the way it was coded. Currently is not possible to have an URL configuration without the 'category'. 

These are the currently possible URL configurations:
![image](https://user-images.githubusercontent.com/577978/146920585-5e5aac64-e432-4ff4-b086-a2e212453d98.png)

I'll leave this issue open as feature request, to see if we can add an URL configuration that does not have the 'category' in the URL. @CaMer0n What do you think?

> Also if the news item have some related news their url's are generated without category, completely ignoring the setting of URL's,

You've fixed this one yourself already :) https://github.com/e107inc/e107/pull/4044

PR #4044, opened 2019-12-13
Resolving an issue with related SEF url's not generating properly

Serios:
Resolving an issue with related news SEF url's not generating properly when news/news-category/news-title is configured in URL's conected to #4040


CaMer0n:
@Serios Thank you for this, and sorry for the lengthy delay.

PR #4047, opened 2019-12-16
Fixing an issue that with contact form sending unnecessary additional…

Serios:
… data

There is an option that dev can add additional fields to contact form, that are sent back. But GDPR field is send back too, also if using NoCaptcha ReCaptcha plugin, response token from Google is send back too.


CaMer0n:
Thank you @Serios !

PR #4071, opened 2020-01-17
Extend cpagefield image type

veskoto:
1. Check if `$value` is **null** or **empty**
2. `{CPAGEFIELD: name=XXXX&type=full}` return full path to image. This ignore thumb params.
3. `{CPAGEFIELD: name=XXXX&type=popup}` return a wrapped image with link to full image path.

### DIFF
--- a/e107_handlers/e_customfields_class.php
+++ b/e107_handlers/e_customfields_class.php
@@ -138,7 +138,7 @@
 
 			$fieldType = $this->_config[$key]['type'];
 
-			if($value === null)
+			if($value === null || empty($value))
 			{
 				return null;
 			}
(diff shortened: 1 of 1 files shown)


Jimmi08:
@veskoto 
Are you aware that with replaceConstants()  you are exploring your hashed sitekey?
And that correct way for this is `$tp->thumbUrl($value,$parm)`  (with parameters w=0&h=0, not sure about correct syntax now to get full image, not resized)

And your popup should be solved in the template by calling image field 2x - one with raw parameter and one without it. The rest is HTML markup that can be different for another theme.

CaMer0n:
Thank you @veskoto . I'll look into this.. the PR is not exactly how I think it should happen, but inside thumbUrl().

Issue #3492, opened 2018-10-06
Maybe it's time to change hybridauth to the new version?


net1313:
3.0 instead of outdated?
More features and more providers. Fewer connectivity issues and more modern approach to authentication through social media.

CaMer0n:
Yes! @net1313 Are you okay to test all the e107 social logins for  FB, Twitter etc.. and report back here?

net1313:
Yes! with great pleasure.

CaMer0n:
@net1313 Excellent! :+1:  I'll commit the new hybrid version to git so you can test it.

CaMer0n:
@net1313 v3 appears to be VERY different to v2 of HybridAuth. It could require a major rewrite of how e107 utilizes it (which I have no time to do). So, this is not a simple upgrade unfortunately.  If you wish to try it yourself and send a pull-request once it is working, it would be very welcome! I don't want to risk breaking everyone's social logins by simply committing it. 

The best I could do for now is to upgrade to the latest v2, since it is primarily a simple overwrite of the files/folders.

net1313:
Sorry, i didn't mention this before. I've already tested new hybridauth. i'm not totally sure (it was one month ago) but it seems that only one thing to change is one function in /handlers/e107_class.php (and then to add more options in social plugin)

CaMer0n:
Okay, that's good news. Would you be okay to upgrade to fork e107 and upgrade hybridAuth on your fork it to the latest version? Then others here could also test it out. And once everything is confirmed working, a pull-request could be made.

net1313:
just give me one week. I'll test it locally.

net1313:
Here is what I tested during the last week. 
Version 3.0.0. RC - apart from declaring namespaces, there are many changes to be made in many of e107 files. Does anybody want?

Version 2.12.0 I have not finished testing yet.

Version 2.11.0 succeffully checked. It works. It has a lot of new providers. I've tested with Facebook, Twitter, Github, Google, Yandex, Vkontake ...

The **"additional providers"** folder goes with version 2.11 of Hybridauth. 
But it is not part of the "Providers" folder and you have to copy the file you need from "adittional providers" into "Hybrid/Providers" folder.
I propose to copy all providers files there.

So the 2.11.0 is OK and could be integrated in without changing e107_class and other files of the system except only one file in Social plugin.
_But at this stage integration would not be done automatically._

In e107_plugins/social/admin_config file in: 
```
$this->social_logins = array (
.....
[...]

PR #4099, opened 2020-02-14
New 3rd-party PHP dependency manager for e107 core

Deltik:
### Motivation and Context
3rd-party dependencies are put into the e107 core repository but periodically need updating, which is a bit of a hassle without a dependency manager.

The e107 core also does not have a standard structure for where 3rd-party libraries are stored.  Examples:

- [`./e107_handlers/pclzip.lib.php`](https://github.com/e107inc/e107/blob/24fe5c80aca4f7c875327ef98a327b4c3d764f11/e107_handlers/pclzip.lib.php)
- [`./e107_handlers/jsshrink/`](https://github.com/e107inc/e107/tree/a2fef504a2c639ab9d7eef3584bd7abea1ea9df7/e107_handlers/jsshrink)

These external libraries are mixed in with core handlers/interfaces, which can make it confusing for plugin authors to know which public interfaces are stable and part of e107.  One notable case is [the removal of the calendar library](https://github.com/e107inc/e107/commit/0a8ec03ab4bfeed4a654a946d6cda65a13ab4a0b), which required all plugins that used it to be modified to [use the equivalent stable e107 core interface](https://web.archive.org/web/20200201102140/https://www.e107sk.com/e107_plugins/forum/forum_viewtopic.php?26.last) (thanks @Jimmi08).

### Description
This pull request introduces [Composer](https://getcomposer.org/) as the external dependency manager for the e107 core.  External dependencies now go into a dedicated folder: `./e107_handler/vendor/`.

The public interfaces in `./e107_handlers/vendor/` are not considered stable.

Plugins **should not** call code inside `./e107_handlers/vendor/` directly.
They **should** only use interfaces (handlers) provided by the e107 framework.  In e107 v2, the handler factory is the `e107` class.

Because this change fundamentally changes how third-party libraries are included in e107, there is a transition plan.

e107 has historically bundled the full source code of external dependencies in the core repository.
Some code, particularly syncing from the GitHub remote, expects dependencies to be included in the core source.

This behavior will be maintained until all existing code depending on the behavior is updated to support resolving dependencies with Composer.

Transition plan:

|e107 Version|Dependency Location|Managed with Composer?|Dependencies Copied in Core Repository?|Behavior Change|
|---|---|---|---|---|
|`<2.3`|`./e107_handlers/`|No|Yes|Legacy behavior|
|`^2.3`|`./e107_handlers/vendor/`|Yes|Yes|Dependencies begin moving to be managed by Composer.  Dependencies target the lowest version of PHP supported by e107 (`config.platform.php` option in `./composer.json`).|
|`^3`|`./e107_handlers/vendor/`|Yes|No|All dependency code is deleted from the core repository's `./e107_handlers/vendor/` folder.  The e107 installer runs `composer install` at the beginning of the install process.  The e107 self-updater runs `composer install` after deploying the desired e107 version.  Only e107 releases may have dependencies bundled in the release package for offline/Intranet/firewalled installations.

To demonstrate these new changes, the Hybridauth (formerly known as HybridAuth) library has been updated from version 2.9.6 to version 3.1.1 in this pull request.  Fixes: #3492

### How Has This Been Tested?
Tests have not been written for the updated Hybridauth because I haven't figured out how Hybridauth works yet…

### Types of Changes

- [x] Documentation (a change to man pages or other documentation)

### Checklist


CaMer0n:
> The e107 core also does not have a standard structure for where 3rd-party libraries are stored. 

Actually it does. subfolders of e107_handlers (phpmailer, phpthumb etc) ARE the vendor folders. 
./e107_handlers/pclzip.lib.php is not a typical example as it dates back to pre v1.x. and could actually be phased out altogether in favor of ZipArchive which is bundled with PHP.

LaocheXe:
@Deltik -  While updating the code, Social Plugin did not save the Keys/ID code, (I used Steam Login on a site, so idk if the secrets, and scopes  are saved or not).

Deltik:
@LaocheXe: This pull request changes the `social` plugin to generate the list of social login providers dynamically.  Unfortunately, there's no API in Hybridauth 3 to know what supplemental fields are present.  Steam is currently treated as plain OpenID, which doesn't accept a secret key.  I hope to have a satisfactory fix for this in a future update to this pull request.

LaocheXe:
@Deltik alright, I removed the key/id that I need for the old method, try to log in and get an Error   Code 5 - SQL Insert Error. In System Log I see XUP_SIGNUP Failure, under that I see ADMINUI_04: Admin-UI DB Error: user
[ERROR] => SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'user_email' cannot be null

The testing site I have set up uses  the same method for logging in as the current main site (which uses the current e107 social plugin), I have it set just to Login, and it never ask for email address because Steam doesn't give out that type of information.

Deltik:
@LaocheXe: The SQL insert error has been fixed in daa31bef5658dd5f21605f529bb00f93cea72a90.

----

f03e60dbd743793afd01476fc4732e5dbc1cc8a0 is my ridiculous solution to figuring out the odd fields in some social login providers (namely `Facebook`, `Foursquare`, `Google`, `Odnoklassniki`, `StackExchange`, `Steam`, `Telegram`, `Twitter`, and `Vkontakte`), but it's not enough for all situations.

For example, `StackExchange` requires the `site` argument, but the way the frontend currently works, only one `site` can be used site-wide.  This means that everyone has to log in via one specific Stack Exchange site; users cannot select their own Stack Exchange site.

Another example is that `Twitter` has two authentication methods depending on the `authorize` argument.  I don't quite understand what the differences are yet, but this is another bit of uncharted complexity.

----

Some social login providers have weird nuances that can't be expressed in the three existing input columns, "Key/ID", "Secret", and "Scope". `./e107_plugins/social/admin_config.php` needs to be redesigned to accommodate the "special" fields.

LaocheXe:
@Deltik Two, maybe three issues now. 1st: new method of steam login uses a different login name (example - new: Steam_76561198006790310 - old: SteambB8047) so I can't login. 

2nd: [query] => `user_xup`='Steam_76561198006790310'   <--- new method, old method: Steam_https://steamcommunity.com/openid/id/76561198006790310

3rd. I have it set up to login only, Which allowed new users to just use Steam to login, so  even with the first two changes, it would login me in as a new account.

Deltik:
@LaocheXe: Okay, lemme unpack the issues you reported:

> 1st: new method of steam login uses a different login name (example - new: Steam_76561198006790310 - old: SteambB8047) so I can't login.

This is a red herring.  The value of `e107_user.user_loginname` doesn't matter.  The lookup is done on `e107_user.user_xup`.  A new account is being created because of the issue below.

> 2nd: [query] => `user_xup`='Steam_76561198006790310' <--- new method, old method: Steam_https://steamcommunity.com/openid/id/76561198006790310

The old output from HybridAuth 2.9.6 is wrong; it is a bug.

**Expected output:** `Steam_76561198006790310`
**Actual (wrong) output:** `Steam_https://steamcommunity.com/openid/id/76561198006790310`

HybridAuth v2.9.6 was coded to find and remove the string `"http://steamcommunity.com/openid/id/"`, but since Steam switched to HTTPS, that string no longer matched.  Hybridauth v3.0.0-rc.8 fixed this in commit https://github.com/hybridauth/hybridauth/commit/4d6e38af71b3bd4cfb1ca7d30ed135b65e9ce521, but that doesn't fix the existing records that have the wrong output.

To correct this in e107, a database migration is required.  **I shall work on fixing this next.**

> 3rd. I have it set up to login only, Which allowed new users to just use Steam to login, so even with the first two changes, it would login me in as a new account.

I don't understand what you mean by this.  If a user tries to sign in via social login without an existing account, they see the error message "_Your login details don't match any registered user. Check if you have the CAPS-LOCK key activated, as logins on this site are case sensitive._" as intended.

LaocheXe:
@Deltik with your version I  see that message, however, with current e107 - it allows the user to login (and add's their account to the database). That's is what I would like it to do, it's a way to combat spam bots.

Deltik:
@LaocheXe: I am still not following.  The error message didn't show before due to a bug that was fixed in c260152b577eee8d2aff85fd32f76ea6a03763f9.  If you use `?route=system/xup/test` on the `master` branch today, the "Test login only with Steam" button would fail the login silently if the user does not exist.

The "Test signup/login with Steam" button will actually register and log in the new user.

What "login only" setting are you talking about?

LaocheXe:
@Deltik  Preferences > User Registration/Login > User  registration system - "Login Only".

https://www.501stlegion-a3.com - using master branch of e107.
https://testing.501stlegion-a3.com - using your branch of  e107. 

"3rd. I have it set up to login only, Which allowed new users to just use Steam to login, so even with the first two changes, it would login me in as a new account."

What I mean is I failed to login, and it threw up an error that it couldn't find "Steam_xxx" but  in the database it was listed as "steam_https://xxx" - it should have created that account in the database  since it couldn't find it?

Both sites use the same setting as listed above, on the main site, new users just login,  account is created.

Testing site, new user login's but fails, because their account is not on the database.

Deltik:
@LaocheXe: Your staging site is using the login-only endpoint (`system/xup/login`), but your production site is using the signup-and-login endpoint (`system/xup/signup`).  To restore the behavior you have in production, use the signup-and-login endpoint.

The bug with `user_xup` being stored as `Steam_https://steamcommunity.com/openid/id/76561198006790310` instead of the correct `Steam_76561198006790310` has been fixed in 639943e7ed18929b17c1990c68ec5294ccb8c2ec.  The incorrect records will be corrected by a manually triggered database update at `/e107_admin/e107_update.php`.

PR #4114, opened 2020-03-28
Automated Release Builds

Deltik:
### Motivation and Context
e107 release builds are currently done manually in an unreproducible closed-source process.

To reduce the effort of building releases and to allow anyone to generate consistent builds, automated release builds are needed.

#### Release Builds
This pull request configures this GitHub repository to run a release build on every push and pull request.  Builds can be **tagged** or **untagged**.

**Tagged** releases are built from a specific Git tag, which is pushed to the GitHub repository at the same time as the revision.  The Git tag is converted into a PHP version, which will be assigned to that e107 release.  For example, the tag `v2.3.0` will build e107 2.3.0.  Tagged releases are expected to be published on [the releases page](https://github.com/e107inc/e107/releases).

**Untagged** releases are practically "nightly" builds.  They are created on every commit as snapshots leading up to the next tagged release.  The version number is the latest tag version or the version in `./e107_admin/ver.php`, whichever is higher, plus `dev` to indicate a prerelease, plus a hyphen (`-`), plus the number of commits since the latest tag version, plus (`-g`), plus the first 9 characters of the Git commit hash.  For example, e107 2.3.0dev-881-g5eb116142 is the `v2.3.0` prerelease built from commit `5eb116142`, which was probably 881 commits after the last tag, whatever that was.

All release builds remove certain files that are not used in production (such as the `./e107_tests/` folder) and generate a checksum map (`./e107_admin/core_image.php`) so that the e107 administrator can use File Inspector to check the integrity of their e107 installation.

#### Core Image

The core image format has been rewritten.  Previously, `$core_image` and `$deprecated_image` were stored as PHP arrays and used by File Inspector to verify the e107 installation.

Now, a phar gzip-compressed JSON object is created containing the following information:

- Default relative path – The path to each file of an uncustomized e107 installation
- Version – The version corresponding to the checksum
- Checksum – The checksum of the path and the version

All core e107 files in the release are checksummed and put into the JSON object using the release version in the version field.

The build script then looks back into all past stable e107 release tags to generate a list of core e107 files that have been removed or renamed by the current release.  These old files are also checksummed and inserted into the JSON object.  This information allows the e107 administrator to see what files they should delete after upgrading their e107 installation:

image

For the core image rewrite, this pull request also includes:
- a base JSON implementation (`e_file_inspector_json`) that stores a pretty-printed JSON string in `$core_image`, but it is not used because the compressed child implementation (`e_file_inspector_json_phar`) has a space savings of over ⅔, and
- an SQLite phar implementation (`e_file_inspector_sqlphar`) that stores the same data in SQLite format, but it is not used because the data retrieval performance was unacceptably poor.

#### `e_file_inspector`

A new abstract class `e_file_inspector` has been introduced to read integrity database files.  The current API is read-only, but it can be expanded later to:
- allow plugin authors to generate integrity images for their own plugins, and
- heal damaged files by downloading them from a mirror.

In the future, a new implementation can be added with low effort to download a detailed integrity database from an official e107 mirror to do integrity checks from an authoritative online source.

#### File Inspector

File Inspector has also been rewritten to support the new `e_file_inspector` interface.  The output behavior of File Inspector differs in some ways:

- Files are now shown with folders instead of in an overlay view.
- JavaScript-expanded help has been fixed.
- In the tree view, the root directory is expanded by default, all sub-folders are contracted.
- If a file checksum matches a previous version of e107, that version is shown.
- Filters that were not clearly defined have been removed.
- The option to perform a scan without an integrity check has been removed.

### How Has This Been Tested?
Only `e_file_inspector` has a few basic unit tests.

The release builds have been verified manually by installing e107 from those builds and then running File Inspector to ensure the new installation is intact.

### Types of Changes

### Checklist


Moc:
@Deltik Can we add the instruction to upload in binary mode within the warning message? I forsee that more users will experience this issue, and having the direct instruction there on how to solve it might prevent (some) new issue reports.

Deltik:
@Moc: Done in https://github.com/e107inc/e107/commit/7291ebc8b1ab29f5abe628b9065b7bcc474a4150

I'm not sure how to localize the message, though.

Issue #4055, opened 2019-12-30
Questions to signup template


Jimmi08:
1. why is used defined and not isset() for variables? (empty is better choise in fact)
`if(!defined($SIGNUP_BEGIN))`

2. at the end of template is correct template array used:
```
$SIGNUP_TEMPLATE['start'] = $SIGNUP_BEGIN;
$SIGNUP_TEMPLATE['end'] = $SIGNUP_END;
$SIGNUP_TEMPLATE['body'] = $SIGNUP_BODY;
$SIGNUP_TEMPLATE['extended-user-fields'] = $SIGNUP_EXTENDED_USER_FIELDS;
$SIGNUP_TEMPLATE['coppa']= '';
```

but inside signup.php are variables used:
`$text = $tp->parseTemplate($SIGNUP_BEGIN.$SIGNUP_BODY.$SIGNUP_END, TRUE, $signup_shortcodes);
`
Couldn't be they changed to new array?

Why do I ask?  Thank require_once there is the way how to extend template with another plugin. But it fails with signup page because those variables are hardcoded at the beginning of the file:

```
$SIGNUP_BEGIN = null;
$SIGNUP_BODY = null;
$SIGNUP_END  = null;
$COPPA_TEMPLATE = null;
$COPPA_FAIL = null;
```

Thanks

CaMer0n:
@Jimmi08 On new themes you can remove `$SIGNUP_BEGIN`, `$SIGNUP_BODY` and `$SIGNUP_END` from the template and place your html code directly into the array, as you would with other templates. 

The intended behavior is that old themes/templates that use only `$SIGNUP_BEGIN`, `$SIGNUP_BODY` and `$SIGNUP_END` continue to function, while new templates should use the array format. 

Ideally, `$SIGNUP_BEGIN`, `$SIGNUP_BODY` and `$SIGNUP_END` should be removed from the core TEMPLATE to avoid confusion.

Jimmi08:
@CaMer0n problem is in signup.php, not in templates

1. last sync e107, bootstrap3 theme,  signup looks ok

2. used core signup template removed $SIGNUP_BODY and content put into $SIGNUP_TEMPLATE['body']  
Form is not displayed.
The reason is:  

`$text = $tp->parseTemplate($SIGNUP_BEGIN.$SIGNUP_BODY.$SIGNUP_END, TRUE, $signup_shortcodes);`

After changing it to:
[...]

PR #4149, opened 2020-05-02
#4055 signup template correct implementation of GDPR shortcode

Jimmi08:
otherwise, there is a visible border with an empty field


CaMer0n:
Thank you @Jimmi08 !

Issue #4130, opened 2020-04-17
still log out from admin area


Jimmi08:
## Question
I have main admin rights to somebody else site.  It worked normally after installation.

Now, there was some question about the theme and I needed to log in again. 

I am able to successfully log in to admin area (I can see dashboard), but if I click anywhere, I am logged out (Menu Manager, Plugins list, Feature box)

The main admin can work normally.   Except for standard plugins they used Stat visits and Newsfeeds.  

I tried different browser, I cleared browser cache and cookies. But I am using the different network than first time (different IP).  

What can be a problem?

Thanks

Jimmi08:
It was definitely by the network (IP) - I switched back to the original network I was logged in the first time and now it is working.

Moc:
I don't know if this is related but I have a similar issue on most (if not all) of my installations where I get random logouts, especially after initially entering the admin area and then clicking my first link to another part of the admin area.

CaMer0n:
@Jimmi08 Were you using a different account (with '0' perms), or the same account as the owner of the site?

Jimmi08:
> @Jimmi08 Were you using a different account (with '0' perms), or the same account as the owner of the site?

Different accounts (no ID 1) but they added main admin rights.  I can't simulate it now, because I am on the same IP (network) like when I log in the first time.

CaMer0n:
@Jimmi08 Okay thanks. User tracking was set to Cookies or Sessions?

Deltik:
This issue is a duplicate of #1836, and the bug is now confirmed.

Issue #1836, opened 2016-09-12
Constant logging out


Khade76:
Whenever changing any page it will log out the user no matter who they are. I have downloaded 1.04 to see if this happened with my current config, but it works perfectly no login issues at all.

Moc:
Is this also the case in the admin area? 

If not, can you try accessing the preferences >  Security & Protection section? Please tell us the `User Tracking method` that is currently set and try switching it to the other option. 

Can you also provide your PHP version please? You can find it in the admin area in About > PHP Info

LaocheXe:
I have this issue on my smart phone, idk if it's Google Chrome browser or something else.

Moc:
@LaocheXe Same question for you as in my previous reply.

LaocheXe:
@Moc sorry for the delay, the session is being tracked via cookies, and last time I used Sessions was a few years back (old e107) when a guy had issues with cookies, and again this happens on my smart phone. 

PHP version is 5.6.27

Deltik:
This bug was caused by 220d258915313c9b895379e6be9db750a01ba3ed (specifically [this line](https://github.com/e107inc/e107/commit/220d258915313c9b895379e6be9db750a01ba3ed#diff-a5745596eb9bb46dabf26d7e53f651f4R434)).

If you are on revision c03945cd6377999e32ddbaf0a0ec89f3a9b52bc8, follow along in this stack trace:

1. e107 is shut down, with all singletons destroyed (including `e_core_session`):
   ```
   e107_class.php:5194, e107->destruct()
   footer.php:400, require_once()
   admin.php:875, {main}()
   ```
2. The problematic line creates a new `e_core_singleton`, but the validation preferences do not get set because of the guard at the top of `e_session->setDefaultSystemConfig()`.  This essentially causes `e_session::SECURITY_LEVEL_HIGH` to be used, which may be higher than what is actually configured.  As a result, this validation now fails:
   ```
   session_handler.php:653, e_core_session->validate()
   session_handler.php:466, e_core_session->init()
   session_handler.php:865, e_core_session->__construct()
   e107_class.php:1011, e107::getObject()
   e107_class.php:1416, e107::getSession()
   e107_class.php:4721, e107::canCache()
   class2.php:2895, e_http_header->send()
   footer.php:413, require_once()
   admin.php:875, {main}()
   ```
3. The session is destroyed:
   ```
   session_handler.php:835, e_core_session->destroy()
[...]

PR #4176, opened 2020-05-24
Overhauled session handler

Deltik:
The e107 session handler has been rewritten.  The code was written for e107 v3 ([namespace style](https://github.com/e107inc/e107/pull/3529)) and backported for use in e107 v2 code.

Applying this patch to an existing e107 site will log out all users due to the completely new in-database way sessions are stored (by default).  If an administrator wants to avoid logouts, they must be using the "session" user tracking mode and set the `session_handler` pref to either `\e107\SessionHandlers\FilesSessionHandler` or `\e107\SessionHandlers\NonblockingFilesSessionHandler` after applying this patch.

Added session handlers:
* `\e107\SessionHandlers\DatabaseSessionHandler` is the new preferred
  session handler, which stores session data in the e_db singleton. A
  new database table called "session" has been created to track the
  session name, session expiry, user who owns the session, and the
  session data.
* `\e107\SessionHandlers\FilesSessionHandler` passes through the native
  PHP "files" session save handler. This is the only locking session
  handler implemented so far.
* `\e107\SessionHandlers\NonblockingFilesSessionHandler` is similar to the
  native PHP "files" session save handler, but session file locks are
  disabled so that concurrent processes on the same session can run at
  the same time.
* A database migration in `./e107_admin/update_routines.php` replaces the
  `user_tracking` pref with `session_handler`.

Also new:
* `\e107\Factories\FactoryInterface` provides an interface for making
  objects in e107 v3.
* `\e107\Factories\BaseFactory` is a normal factory that creates new
  instances on every invocation of `FactoryInterface::make()`
* `\e107\Factories\SessionHandlerFactory` creates new instances of
  `\e107\SessionHandlers\BaseSessionHandler` based on a new pref,
  `session_handler`.

Removed from e107 v2:
* The global-scope helper method `session_set()`
* The pref `user_tracking` has been removed. It used to toggle between
  "session" mode and "cookies" mode, but the "cookies" mode has been
  removed entirely, making `user_tracking` unnecessary. Different
  implementations of "session" mode are now decided by the
  `session_handler` pref.
* "autologin" (AKA "Remember me") has been removed because it only
  worked in `user_tracking` = "cookies" mode, and "cookies" mode has
  also been removed. Sessions currently last as long as the
  system-configured lifetime. More work can be done here to make session
  lifetimes user-tunable.

Fixes: #1836
Fixes: #4130

### Motivation and Context
There are multiple motivations for overhauling the session handler:
* The old session handler prevents anyone from loading multiple pages at the same time from the same session (web browser) because it locks the `$_SESSION` variable until the session is closed.  PHP processes pile up and consume memory, possibly exhausting the available RAM on the server.  This can happen if a user or even search engine spider rapidly opens up multiple pages.  A slow page (for any reason) also prevents other pages from loading until that slow page is finished.
* The old session handler used the native PHP "files" session save handler, which has these additional downsides:
  * Consumes one inode per session. If never cleaned, the admin's hosting account could exhaust its file count quota.
  * Resource-intensive to scan through session files to determine which ones to prune
  * Another application could garbage-collect session files earlier than configured, leading to unexpected logouts for users.
  * It is a common practice to store the session files in `/tmp`, which is typically wiped every reboot. This causes unexpected logouts for users.
  * Session files are often stored with permissive permissions, which leaves potentially sensitive data to be accessed by other sites/applications on the same shared hosting server.
  * Harder to scale horizontally; the session files must be put in shared storage for the sessions to be usable by multiple web nodes in the cluster.
* With the new non-blocking session handler options, multiple pages can load concurrently, which results in a faster user experience.  e107 sessions largely consist of read operations, which hopefully makes session-related race conditions not much of a problem.
* Even if race conditions are a problem in the non-blocking session handler, the site administrator can choose to use a blocking session handler to prevent any drawbacks from the non-blocking options.
* The new database-based session handler stores session data in the e107 database.  This implementation has practically none of the downsides of the file-based session handlers.
* The new database-based session handler allows these new features to be added to e107 in the future:
  * A dashboard for a user to:
    * see their existing sessions and
    * terminate any of their existing sessions
  * A dashboard for an administrator to manage all user sessions
* This pull request fixes a security issue. ([Private link to discussion](https://gitter.im/e107inc/tests?at=5e49c415d56ddb68a4ac0576))

### How Has This Been Tested?
The newly added classes are tested with [the automated unit test suite](https://github.com/e107inc/e107/tree/master/e107_tests).

### Types of Changes

- [x] Performance enhancement (non-breaking change which improves efficiency)

### Checklist

- [ ] My code adheres to the e107 [code style standard](https://github.com/e107inc/e107/wiki/e107-Coding-Standard).


CaMer0n:
Closing this in favor of incremental changes that conform to e107 standards. (see https://github.com/e107inc/e107/commit/7439d599f84d122df06dae61ea09ce3e0415d8af ) 
Some (still necessary) changes may be added manually or in separate pull-request.

Issue #4319, opened 2020-12-29
LM_REMEMBERME - outdated code for checkbox


Jimmi08:
This is there now:

```
			function sc_lm_rememberme($parm='')
			{
				$pref = e107::getPref();
				if($parm == "hidden"){
					return "<input type='hidden' name='autologin' id='autologin' value='1' />";
				}
				if($pref['user_tracking'] != "session")
				{
					return "<label for='autologin'><input type='checkbox' name='autologin' id='autologin' value='1' checked='checked' />".($parm ? $parm : "".LAN_LOGINMENU_6."</label>");
				}
				return '';
			}
```

CaMer0n:
If you can add classes that make it BS3 and BS4 compatible, that would be ideal. :+1:

Jimmi08:
I just asked for using form class with checkbox()  It works correctly for bt3 and now for bt4 (I mean snippets)

CaMer0n:
If you can send a pull-request for these, that would be appreciated. The important thing is that the HTML markup remains almost the same. (so as not to break existing theme styling)

Jimmi08:
closing in favor of https://github.com/e107inc/e107/issues/4331

PR #4330, opened 2020-12-31
Update remember me checkbox for signin plugin

Jimmi08:
Using e107::getForm()->checkbox() to be able use snippets for HTML markup

similar as  #4319, but not created separate issue


CaMer0n:
I'm thinking to remove the 'label' altogether from the shortcode.

PR #4340, opened 2021-01-10
Adding shortcode for displaying link to  login page

Jimmi08:
It is needed if you don't use the login dropdown version.


CaMer0n:
Thanks you!

PR #4360, opened 2021-01-28
adding class for styling pm nav item

Jimmi08:
just detail, similar like dropdown-avatar


CaMer0n:
Great! Thank  you @Jimmi08

PR #4379, opened 2021-02-12
Added Portuguese_admin language file to theme bootstrap3

rica-carv:
Added Portuguese_admin language file to theme bootstrap3


CaMer0n:
Thank you @rica-carv ! This is great, but it needs to go into the languagepack.

Issue #4372, opened 2021-02-07
Separate pref in admin for Phone, Email, and Address


fizi:
I think would be great a separate pref in admin for Phone, Email, and Address, and google map (Contact page). Now they are in the one textbox and only one shortcode for it. I can not style the details and I think this option is poor.

CaMer0n:
@fizi This has been completed.  But pls let me know of any template changes you think would be useful. 

fyi @Jimmi08 @Moc

CaMer0n:
![image](https://user-images.githubusercontent.com/2825679/107987454-e0a0d380-6f82-11eb-9624-4305c5a31106.png)

Jimmi08:
Please, add this as a preference to admin display options to be able to hide those fields.  There are themes that used theme preferences for something like this.
 
At least get back the description field for the contact page, not replace it. There is no way after emptying this field to get it back.

This is just not the correct approach.  I want on contact page Working hours, I want different zoom or google maps by coordinates, not by address,  What now?  Will I have contact information on 2 places now? So give me an option to hide this. 
 
So now each theme should support those fields?  Because users can fill them now and they will ask why they are not displayed.

And if you added an email field, why it is not type email?

CaMer0n:
@Jimmi08 It is up to themes to conform to e107's architecture, not the other way around. You would have done better to suggest this enhancement to the core ( like @fizi did) but sooner, rather than added non-theme-related data fields to your theme preferences. Theme preferences should be about layout/styling, not about storing user data/content for core pages like 'contact.php'. 

>  I want on contact page Working hours, I want different zoom or google maps by coordinates, not by address, 

Exactly, these are feature requests that should be directed to the e107 core here on github, not to theme developers to put into theme preferences.

Jimmi08:
I asked for this with an Industrial theme. It was discussed on gitter. I had to create a Contact Page Extended plugin to have those fields and then I found that there was already a legacy plugin for this.  @rica-carv will maybe remember this discussion.

I don't mind your changes. I want the back description field for the contact page.  It is missing now.

CaMer0n:
Yes, but a discussion != Feature-request created. 

Now, the field that was removed was create to hold company name, address, phone etc. together. 
This has now been replaced by separate fields. 

So, my question is: why do you need this generic contact field back when it is replaced by the new fields?

Jimmi08:
Because you limited functionality of that field to just some listed fields!.   There can be a text field too f.e. adding some text information...  
[...]

PR #4406, opened 2021-02-19
#4372  missing strings for frontend

Jimmi08:
### Description
missing strings for frontend 

#4372 Separate pref in admin for Phone, Email, and Address

### DIFF
--- a/e107_languages/English/lan_contact.php
+++ b/e107_languages/English/lan_contact.php
@@ -42,4 +42,8 @@ define("LANCONTACT_22", "Privacy policy");
 define("LANCONTACT_23", "You can find our privacy policy here: [x]");
 define("LANCONTACT_24", "GDPR Agreement");
 
+define("LANCONTACT_25", "Phone");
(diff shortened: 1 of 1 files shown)


CaMer0n:
@Jimmi08 They're not 'missing' because they are not hard-coded in the template. You just want to have them added to the core for your theme template?

Jimmi08:
You added new fields but you are using only icons as labels.  Just now I have the theme where I need text label. Their labels should be in the contact LAN file if they are core fields now.

e107_languages/English/lan_contact.php
```diff
@@ -42,4 +42,8 @@
 define("LANCONTACT_23", "You can find our privacy policy here: [x]");
 define("LANCONTACT_24", "GDPR Agreement");
 
+define("LANCONTACT_25", "Phone");
+define("LANCONTACT_26", "Adress");
```
CaMer0n, line 46: Should be "Address"

Issue #3912, opened 2019-07-30
theme visibility filter problem with $


Jimmi08:
On first look everything looked correct
![image](https://user-images.githubusercontent.com/5429548/62115782-4a7e5d00-b2b9-11e9-937a-1e37006d4513.png)
But I couldn't get it to work.

if I do this
` var_dump($kpage); print_a($kpage); `
result is:
![image](https://user-images.githubusercontent.com/5429548/62116443-76e6a900-b2ba-11e9-8546-dc5c43a893eb.png)

so I checked the database and there is value 
 
![image](https://user-images.githubusercontent.com/5429548/62116615-c4fbac80-b2ba-11e9-9aa5-061675891acf.png)

so of course this test fails

```
          var_dump($lastChar); print_a($lastChar);  
					if($lastChar === '$') // script name match.
					{  
```
![image](https://user-images.githubusercontent.com/5429548/62116744-fd02ef80-b2ba-11e9-9f3d-9eea121aae7c.png)

If I check Database Tools/Prefs Editor, it displays
![image](https://user-images.githubusercontent.com/5429548/62116868-38052300-b2bb-11e9-9b98-d070b1091141.png)

Maybe it's environment issue (xampp), no idea, but that check should count with both version of $.

Jimmi08:
Adding this line helped:  `$kpage = html_entity_decode($kpage, ENT_NOQUOTES, 'UTF-8');`

Moc:
Confirmed

CaMer0n:
@Jimmi08 Please re-test and open again if necessary. Thank you.

CaMer0n:
As of v2.3.1 one can now use ```:(e_ROUTE)``` for theme layout switching in the Theme Manager.
eg. ```:news/list/item```

![image](https://user-images.githubusercontent.com/2825679/107885766-eded8d80-6eb0-11eb-8a13-2fab42160911.png)

In the browser, add `?[debug=paths!]` to the SEF URL to find the current value of  `e_ROUTE` listed in the footer of the page. (must be logged in as admin)   

![image](https://user-images.githubusercontent.com/2825679/107886020-45402d80-6eb2-11eb-9ef2-b2f293245225.png)

 It also accepts partial matches. 
eg. You could enter `:vstore` for all SEF urls of the vstore plugin, or `:news/list` to match all `news/list` routes such as: 

- news/list/all
- news/list/short
- news/list/category
- news/list/day
- news/list/month
- news/list/tag
[...]

PR #4417, opened 2021-02-22
#3912 e_ROUTE for legacy version gsitemap plugin

Jimmi08:
#3912 e_ROUTE for legacy version gsitemap plugin


CaMer0n:
Thank you!!

PR #4421, opened 2021-02-23
#2898 fix comments offsets for bootstrap4

Jimmi08:
#2898 fix comments offsets for bootstrap4

it should be tested, I did it only for metis theme.


CaMer0n:
Thank you @Jimmi08 !

Issue #4402, opened 2021-02-18
problem with book template and pages listing on book page.


Jimmi08:
## Question
What are you confused or uncertain about?
book/chapters templating started to behave the weird way

What are you trying to accomplish?

I have a chapter with a new template (new template key).  The chapter page is correct and everything works.
```
$CHAPTER_TEMPLATE['chapterpage']['listPages']['caption']				= "{CHAPTER_NAME}";
$CHAPTER_TEMPLATE['chapterpage']['listPages']['start'] 	 	= '<div  >';
$CHAPTER_TEMPLATE['chapterpage']['listPages']['item'] 					= '<h3 id="{CPAGEANCHOR}" >{CPAGETITLE}</h3>
<div class="content">{CPAGEBODY}</div>';
$CHAPTER_TEMPLATE['chapterpage']['listPages']['end'] 					= "</div>"; 
```

But this template is used on the book page too for shortcodes {PAGES}.   
It shouldn't work this way (by looking at the code) - 
You set the default template key for book/chapter in preferences - in my case "default"

On the book page that template is used:

```
$CHAPTER_TEMPLATE['default']['listChapters']['caption']				= "{BOOK_NAME}";	
$CHAPTER_TEMPLATE['default']['listChapters']['start']				= "<ul class='page-chapters-list'>";
$CHAPTER_TEMPLATE['default']['listChapters']['item']				= "<li><h4><a href='{CHAPTER_URL}'>{CHAPTER_NAME}</a></h4>{PAGES}";
$CHAPTER_TEMPLATE['default']['listChapters']['end']					= "</ul>";
```
Still correct 

And for {PAGES} should be used "default" too.  We are on the book with template "default". 

So how it should work? If I am correct, it is a bug, if not, you can have just one template for chapters, nothing more.

CaMer0n:
@Jimmi08 Thank you.  I want to be sure we're talking about the same thing.
If I understand well, you're referring to the 'listing' page, which utilizes 3 templates. 

ie.  http://localhost/e107v2/page/ with `List pages if no page selected` enabled in the prefs. 

Here's my test case. I created an 'other' key and added descriptions (eg. `(default/listChapters)`)  so we can see where they are used: 

``` 
$CHAPTER_TEMPLATE['default']['listPages']['caption']				= "{CHAPTER_NAME} ";
[...]

PR #4422, opened 2021-02-24
#4402 fix wrong template key,

Jimmi08:
Probably just mistypo


CaMer0n:
@Jimmi08 Yes it was, thanks.

PR #4425, opened 2021-02-26
#3912 e_ROUTE for non pretty news urls

Jimmi08:
Not sure where `news/list/short` is used


CaMer0n:
Thank you for this @Jimmi08 .

Issue #275, opened 2013-05-10
Accessibility improvements


Moc:
Posted by Axeia
http://e107.org/e107_plugins/forum/forum_viewtopic.php?245498

> - Links are ignoring default hotkeys. In Firefox (and afaik, Internet Explorer, Chrome, Safari Opera and so on) the middle mouse button when pressed on a link opens the link in a new tab. This isn't the case in the admin panel.

> - Lack of labels basically everywhere. e107_admin/prefs.php#nav-core-prefs-email for example lends itself to the use of labels just perfectly. You have the title on the left and the control on the right, simply use the title as a label for the field, set the for attribute. This helps the browser associate the text with the field improving accessibility.

> - Some 'clever' bit of javascript converting <select> elements to other stuff that doesn't allow you to type part of what you have entered to have it automatically match the rest. Take for example e107_admin/prefs.php#nav-core-prefs-email again. If I select the "Emailing method" field I expect it to have "smtp" selected after I type "sm", this isn't the case. There should be other styleable solutions like for example this one: http://adam.co/lab/jquery/customselect/ which do allow it.

> - /e107_admin/docs.php is undeniably a list of some sorts. So it should be using html defining it's a list 
  ( [W3.org > Structures > Lists](http://www.w3.org/TR/html401/struct/lists.html) ). 

> - The default admin theme doesn't make much use of heading elements (h1-6). The "title" of each box on for example the admin frontpage should be a heading in my opinion. On the same page there is again a list with options (the control panel) which isn't in a HTML list element

MikeyGMT:
This is a really good point, not everyone uses the web visually, there are screen reader browsers and other which help people with impairments to navigate, but also for future proofing, for speech navigation - like "OK Google".  
Just did a test with http://wave.webaim.org/ on contact.php on bootstrap3 latest git, standard contact template - Cyborg Style. 
I don't know if this is the best and most accurate tool. 
It marking the page down because of these points:-

* Lack of H1
** Hence we need a heading e.g. "Contact Us" 
* Labels and data fields orphaned
** There is a problem here, the template contains the Labels and the Shortcode contains the fields
** Can we move the Label to within the shortcode or are there any other suggestions?
** Ooo how about if we could change the way templates work? Could they be more like the Menu Manager, then the user can move the order of the fields on the fly in the application, but keep the accessibility. field/label integrity. 
* click to visit link for e107 Banner 
** Simple change to the default text to "Learn more about e107 by clicking here"

[...]

PR #4436, opened 2021-03-03
#275 fix for Unlabeled form control with title for WAVE

Jimmi08:
contact form fields  if no labels are used


CaMer0n:
Thank you @Jimmi08 . I'm very happy that you're checking this.

Issue #4455, opened 2021-03-18
Multilanguage theme preferences 


Serios:
I'm a bit lost or maybe incidentally find a bug. Does multi language support includes theme preferences? Do I need to install some additional plugin?
I found a similar question by @Jimmi08 in [issues #2591](https://github.com/e107inc/e107/issues/2591) from couple of years ago. 

The code is actually working if defined with 'multilan'=>true in theme_config.php the desired multi language field get marked as such in theme pref tab, after that it can be used in theme_shortcodes returned as array, however the problem is that every time the language value is updated trough theme pref tab for the given language, the values for the other languages are omitted. 

After bit investigating I found that in core DB table only the value for the last updated language is stored.
In theme_test_theme row the stored pref are:

array( 'footer_text'=> array( 'English'=>'This is footer text', ) )
but should be 
array( 'footer_text'=> array( 'English'=>'This is footer text', 'Dutch'=>'This is just a test for nl', ) )

If I switch the language  and edit theme pref for the Dutch, the stored pref become 
array( 'footer_text'=> array( 'Dutch'=>'This is just a test for nl', ) )
omitting the previously stored value for English

Jimmi08:
@Serios After this answer by CaMeRon
> Theme preferences should be about layout/styling, not about storing user data/content for core pages like 'contact.php'. 
I stopped using theme prefs to store any text strings.

And read my comment here:
https://github.com/e107inc/e107/issues/3803#issuecomment-496993657
So my suggestion (what I do): create a simple plugin with a table for needed strings and then use a language table solution. 
And by the way... with the plugin you have the option to add user access to those strings (edit them), with theme prefs you need to give them access to the full theme manager, it is sometimes not a good idea.

Serios:
@Jimmi08 I agree with you, but in fact it was about a couple of texts, that I want to be able to edit from time to time, thus storing them in pref for the theme seemed as a good idea. I will take your advice and go with the plugin...

Jimmi08:
@e107help this can be closed as solved

working example:
```

	public function config()
	{
		return array(

			// Home page > Features section subtitle. A theme pref rather than a
			// Featurebox item because it is section chrome, not repeatable content.
			'features_subtitle' => array(
				'title'      => defset('LAN_THEME_PREF_FEATURES_SUBTITLE', 'Features: subtitle'),
				'type'       => 'text',
				'multilan'   => true,
				'writeParms' => array(
					'size'      => 'xxlarge',
[...]

PR #4456, opened 2021-03-19
Fixing an issue with multilanguage values in Theme preferences

Serios:
When theme pref field have multilanguage enabled, values for the field are overwritten by last edited language. More detailed description about this behavior is given in #4455


CaMer0n:
@Serios Thank you! :+1:

PR #4457, opened 2021-03-20
Fixing an issue with gallery prettyPhoto plugin

Serios:
.size() function was deprecated as of jQuery 1.8. And removed in jQuery 3.0. More information [here](https://api.jquery.com/size/)

Using it, is causing prettyPhoto not to work, with newer versions of  jQuery. It must be replaced with .length property


CaMer0n:
@Serios Thank you! :+1:

PR #4554, opened 2021-09-04
PHP 8.1 compatibility

Deltik:
### Motivation and Context
PHP 8.1 about to be released, but e107 is not compatible with various deprecations and behavior changes.

### Description
This pull request fixes all PHP 8.1 errors caught by the existing tests.  Specifically:

#### [Deprecations](https://wiki.php.net/rfc/deprecations_php_8_1)

* `strftime()` has been replaced with a polyfill based on `DateTime`.
* Explicit type casts/assertions added where required by PHP 8.1
* `filter_var(…, FILTER_SANITIZE_STRING)` replaced with `strip_tags()`
  or HTML entity encoding of quotation marks, depending on a guess of
  what the intended "sanitization" was
* `http_build_query()` usage type mismatches fixed
* Removed usages of the `FILE_TEXT` constant
* To avoid breaking PHP 5.6 compatibility (function return types),
  `e_session_db` no longer implements `SessionHandlerInterface`.
  Instead, the alternative non-OOP invocation of
  `session_set_save_handler()` is used instead to apply the session
  handler.
* The shim for `strptime()` still calls the native function if available
  but now suppresses the deprecation warning.

#### [Behavior Changes](https://github.com/php/php-src/blob/4025cf2875f895e9f7193cebb1c8efa4290d052e/UPGRADING)

* `e_db_pdo` explicitly asks for `PDO::ATTR_STRINGIFY_FETCHES` to
  maintain consistent behavior with past versions of PHP.
* `e_db_mysql` explicitly sets `mysqli_report(MYSQLI_REPORT_OFF)` to
  maintain consistent behavior with past versions of PHP.

#### Other

* Removed pointless random number generator seed from `banner` plugin
* Workaround for `COUNT(*)` SQL query in
  `validatorClass::dbValidateArray()` without a proper API for avoiding
  SQL injection

### How Has This Been Tested?
The changes make the existing tests pass on all PHP major versions since version 5.6.

### Types of Changes

### Checklist

### DIFF
--- a/comment.php
+++ b/comment.php
@@ -107,7 +107,7 @@ if(e_AJAX_REQUEST) // TODO improve security
 	{
 		$pid 				= intval(varset($_POST['pid'], 0)); // ID of the specific comment being edited (nested comments - replies)
 		$row 				= array();
-		$authName           = filter_var($_POST['author_name'],FILTER_SANITIZE_STRING);
+		$authName           = e107::getParser()->filter($_POST['author_name'], 'str');
 		$clean_authorname 	= vartrue($authName,USERNAME);
 		$clean_comment 		= e107::getParser()->toText($_POST['comment']);
 		$clean_subject 		= e107::getParser()->filter($_POST['subject'],'str');
--- a/class2.php
+++ b/class2.php
@@ -1233,7 +1233,7 @@ function check_class($var, $userclass = null, $uid = 0)
 
 	$class_array = !is_array($userclass) ? explode(',', $userclass) : $userclass;
 
-	$varList = !is_array($var) ? explode(',', $var) : $var;
+	$varList = !is_array($var) ? explode(',', (string) $var) : $var;
 	$latchedAccess = false;
 
 	foreach ($varList as $v)
--- a/e107_admin/admin_log.php
+++ b/e107_admin/admin_log.php
@@ -371,7 +371,7 @@ class admin_log_ui extends e_admin_ui
 			if(($action == "backdel") && isset($_POST['backdeltype']))
 			{
 			//	$old_date = intval($qs[1]);
-				$old_string = strftime("%d %B %Y", $old_date);
+				$old_string = e_date::strftime("%d %B %Y", $old_date);
 				$qry = "dblog_datestamp < ".$old_date; // Same field for both logs
 
 				switch($_POST['backdeltype'])
--- a/e107_admin/banlist_export.php
+++ b/e107_admin/banlist_export.php
@@ -65,7 +65,7 @@ if (!empty($_POST['ban_types']))
 	}
 }
 
-$filename = 'banlist_'.strftime("%Y%m%d_%H%M%S").'.csv';
+$filename = 'banlist_'.e_date::strftime("%Y%m%d_%H%M%S").'.csv';
 
 if ($error_string = do_export($filename, $type_list, $format_array, $use_separator, $use_quote))
 {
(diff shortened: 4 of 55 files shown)


comment.php
```diff
@@ -107,7 +107,7 @@
 	{
 		$pid 				= intval(varset($_POST['pid'], 0)); // ID of the specific comment being edited (nested comments - replies)
 		$row 				= array();
-		$authName           = filter_var($_POST['author_name'],FILTER_SANITIZE_STRING);
+		$authName           = e107::getParser()->filter($_POST['author_name'], 'str');
```
CaMer0n, line 110: @Deltik Maybe I missed something, doesn't this do exactly the same as the code you replaced?

```
public function filter($text, $type = 'str', $validate = false)

...



if($validate === false)
				{
					$filterTypes = array(
						'int'   => FILTER_SANITIZE_NUMBER_INT,
						'str'   => FILTER_SANITIZE_STRING, // no html.
						'email' => FILTER_SANITIZE_EMAIL,
						'url'   => FILTER_SANITIZE_URL,
						'enc'   => FILTER_SANITIZE_ENCODED
					);
				}
				```

**Comment by CaMer0n** [comment]
Thank you!! Looks good, please see comment about filter().

File: comment.php (hunk 1 above)
Comment (Deltik) [reply comment.php:110]: `e_parse::filter()` has been reworked not to use `filter_var(…, FILTER_SANITIZE_STRING)`.  ([Source](https://github.com/e107inc/e107/pull/4554/files#diff-dc57abd51013ddd4e6bad33d34ac2dae0d2e5302900c00ed69dd4dee3556d939R4855))

According to [the documentation of the deprecation of `FILTER_SANITIZE_STRING`](https://wiki.php.net/rfc/deprecations_php_8_1#filter_sanitize_string), the usage of the filter is unclear.  I looked at e107's usages of that filter and found that some of them expected [`strip_tags()`](https://www.php.net/manual/en/function.strip-tags.php) behavior while others thought that the filter would make the provided string safe to quote in an SQL query.  As there is no way for a filter to concatenate arbitrary strings safely for SQL queries, I decided to make `e_parse::filter(…, 'str', false)` just mean `strip_tags()`.

I replaced the confounded usages of `filter_var(…, FILTER_SANITIZE_STRING)` with guesses for what the usages' intentions were:
* Preventing SQL injection – [Example](https://github.com/e107inc/e107/pull/4554/files#diff-061a207c81a13f011e0c692099bfffd7bad34c9650f804f9968b022d9691b5cdR1389-R1402)
* Remove HTML tags – [Example](https://github.com/e107inc/e107/pull/4554/files#diff-068ed6f9c1bcc16dcb2cec1aa9ebccec2bcd68400ae947c9b53465a885a2d3c6R202)
* Kinda hide quotation marks for use in SQL – [Example](https://github.com/e107inc/e107/pull/4554/files#diff-dc57abd51013ddd4e6bad33d34ac2dae0d2e5302900c00ed69dd4dee3556d939R511)
* No apparent reason at all – [Example](https://github.com/e107inc/e107/pull/4554/files#diff-19e3c84c1b3ce87245588ae6ee7791fe3b94c98a8f955ea3e89f9f5865144ca0L268)

**Comment by CaMer0n** [comment]
@Deltik This commit introduced a bug into the form handler. 

git.exe bisect good
20882920a0b68937570264949512acc0c4841dbd is the first bad commit
commit 20882920a0b68937570264949512acc0c4841dbd
Author: Nick Liu 
Date:   Sat Sep 4 15:06:19 2021 +0200

Specifically it renders the following invalid HTML in the batch options and elsewhere. 

I previously committed a fix for the pop-up confirmation on the admin area cache delete button, which had a similar issue. ( https://github.com/e107inc/e107/commit/632f33526f41bc09dcb725b2fc2076a64d297855)

It appears the parsing of  the `$options['other'] ` value has been corrupted (see quick-fix commit above) 


```
<select name="etrigger_batch" id="etrigger-batch" class="tbox form-control input-large select batch e-autosubmit reset" data-original-title="" title="">
<option value="">With selected...</option>
<option value="copy" class="ui-batch-option class" style="&quot;padding-left:" 15px&quot;="">Copy</option>
<option value="delete" class="ui-batch-option class" style="&quot;padding-left:" 15px&quot;="">Delete</option>
<option value="export" class="ui-batch-option class" style="&quot;padding-left:" 15px&quot;="">Export</option>	

```

Issue #4560, opened 2021-09-10
`e_library_managerTest::testDetectionVersionConsistency()` depends on the Internet


Deltik:
## Bug Description
The test [`e_library_managerTest::testDetectionVersionConsistency()`](https://github.com/e107inc/e107/blob/8fc922c126e36bc2a6e6b4c320d1b567090c85de/e107_tests/tests/unit/e_library_managerTest.php#L69-L92), located in [`e107_tests/tests/unit/e_library_managerTest.php`](https://github.com/e107inc/e107/blob/8fc922c126e36bc2a6e6b4c320d1b567090c85de/e107_tests/tests/unit/e_library_managerTest.php), downloads content from the Internet.

Unit tests should not depend on the Internet, as those resources may disappear or change and cause inconsistent testing behavior.

## How to Reproduce

1. Disconnect the test machine from the Internet.
2. Run all unit tests.
3. Observe the test failure:
   ```
   There was 1 failure:

   ---------
   1) e_library_managerTest: Detection version consistency
    Test  tests/unit/e_library_managerTest.php:testDetectionVersionConsistency
   No looked-up version in core_library:config() -- cdn.jquery
   Failed asserting that false is true.
   #1  ~/Documents/Development/PhpstormProjects/e107/e107_tests/tests/unit/e_library_managerTest.php:84

   FAILURES!
   Tests: 707, Assertions: 4801, Failures: 1.
   ```

## Expected Behavior
No Internet connection for any unit tests

## Additional Context
This bug was introduced by https://github.com/e107inc/e107/commit/2f780fca69fda54843f5f73ae835d15f2875b4d0.

Past fixes for tests depending on the Internet:
* https://github.com/e107inc/e107/commit/556aea02ea0d5d848ebbd2761adeeb89c6a9c4c2
* https://github.com/e107inc/e107/commit/9e6bb14641bcc3bd62320e53004e51803536baef
* https://github.com/e107inc/e107/commit/55bd50aac0aae8054a9a3a581a9531c5cd27ba93

PR #4561, opened 2021-09-10
Removed Internet dependency for test `e_library_managerTest::testDetectionVersionConsistency()`

Deltik:
Fixes: https://github.com/e107inc/e107/issues/4560

### Motivation and Context
`e_library_managerTest::testDetectionVersionConsistency()` is pointlessly downloading resources from the Internet, so it was really only testing that the CDN was accessible.  This is not what a unit test should be doing.

### Description
Changed the behavior of `e_library_managerTest::testDetectionVersionConsistency()` so that it checks if the CDN library version is the same as the local library version without connecting to the Internet

### How Has This Been Tested?
By fixing the existing test so that it actually tests something useful

### Types of Changes

### Checklist


CaMer0n:
> Changed the behavior of e_library_managerTest::testDetectionVersionConsistency() so that it checks if the CDN library version is the same as the local library version without connecting to the Internet

The point of the test was to check that the hard-coded version number located in the library class matched the version number detected and written in the CDN file itself. Hard-coding the version like this improved page-load performance dramatically. 

It was never about testing local versions of the library against CDN versions of the same library. Perhaps a separate test should be added for this.

Deltik:
> The point of the test was to check that the hard-coded version number located in the library class matched the version number detected and written in the CDN file itself. Hard-coding the version like this improved page-load performance dramatically.

Can we mock the Internet part by loading the local files?  It should be out of our scope to check that the CDN is serving the right content, as that behavior is not within our control.

> It was never about testing local versions of the library against CDN versions of the same library. Perhaps a separate test should be added for this.

The name of the method `testDetectionVersionConsistency()` suggested to me that the intention was to compare the CDN version with the local version.  Maybe this test should become what the name looks like and then a new test be written to check that the CDN URL includes the library's version number.

PR #4721, opened 2022-03-28
Fix for Recalculation forum replies in Tools

### LINKED ISSUE #4570: Recalculation forum replies - wrong counting
I think that this problem is related to the changed logic (structure ) in the forum for v2.  

1. you have a forum - table forum
2.  you add a new topic -  thread record is created and first post (like a copy of it)
3. you add reply - the second post is created 

But you count replies as all posts, not one down: 
`	$replies = $sql->count('forum_post', '(*)', 'WHERE post_forum='.$forumId);`

Look:
Forum: 
image

Thread (it was correct before using recalculation)
image

Topic:
There are 2 posts, but only one is a reply. 

$replies should be = post's count minus one.  Correct?

Jimmi08:
Fix #4570


CaMer0n:
Thank you @Jimmi08 !

PR #4725, opened 2022-03-29
#4724 forum - possibility to add forum ID as column

### LINKED ISSUE #4724: forum - possibility to add forum ID as column
## Motivation
The forum ID column is not visible by default, but it should be part of the available columns. (now you need to edit record to read it from URL)

## Proposed Solution
adding `'forced'=> true, `

Jimmi08:
Closes #4724


CaMer0n:
This commit doesn't provide "possibility to add forum ID as column" it forces the ID to be visible at all times. Also the PR contains other unrelated changes. `'type'=>'number'` would be more appropriate. Please resubmit.

Jimmi08:
Thanks, I didn't know about this option. 

Sorry, it somehow combined other PRs.

PR #4727, opened 2022-03-29
#4706 forum post with quote fix

### LINKED ISSUE #4706: quote functionality in forum is not working
## Bug Description
when you use quote in a forum, it should be visibly different. 

This is not working:
`$text = $tp->toText("[quote={$quoteName}]\n".$post."\n[/quote]\n",true);`

Not sure why, but maybe because of this detection for bbcode use?

`$bbsearch = array('[/img]', '[/h]', '[/b]', '[/link]', '[/right]', '[/center]', '[/flash]', '[/code]', '[/table]');`

## How to Reproduce
1.  Try to quote any post in forum with 2.3.1 with tinymce editor

Jimmi08:
I am aware that this is not the best solution, but it is the solution that works in any version of bootstrap. And it fixes the quote issue only in the forum where it is needed to be fixed.  

I will put information related to styling to that issue.  

FIX #4706


Jimmi08:
> I fixed the main issue. (BOOTSTRAP not defined and deprecated LAN file being loaded) . PR no longer needed.

Excellent. Your way is always welcome. Thanks

PR #4728, opened 2022-03-29
#4722 forum - issue with forum topic pagination

### LINKED ISSUE #4722: forum - issue with forum topic pagination
## Motivation

pagination on forum topic is done with ?p=1 ?p=2 etc 

Problem is that you can use any parameter and the page is displayed (no 404 error, no error message, anything) 

It shouldn't work this way

## Proposed Solution
No idea for now, just marked.

**Note:** 
if you use by mistake for example:  ?p=-3 
you got fatal error Fatal error: Uncaught TypeError: count(): Argument #1 ($var) must be of type Countable|array,
There should be some checks for this

Jimmi08:
FIX #4722

### DIFF
--- a/e107_plugins/forum/forum_viewtopic.php
+++ b/e107_plugins/forum/forum_viewtopic.php
@@ -638,6 +638,11 @@ class e107ForumThread
 
 		$totalPosts = $this->threadInfo['thread_total_replies'] + 1; // add +1 for the original post. ie. not a reply.
 		$this->pages = ceil(($totalPosts)  / $this->perPage);
+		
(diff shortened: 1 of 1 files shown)


CaMer0n:
@Jimmi08 Codeclimate came back with an error. https://codeclimate.com/github/e107inc/e107/pull/4728  
Could you please check `$thread` has a value?

Jimmi08:
@CaMer0n Closing, my mistake, it should be   $this not $thread,
I don't know how to edit this one. 
I will create a new one after some checks (why it worked, it should't )

PR #4730, opened 2022-03-30
#4659 forum main admin as silent moderator

### LINKED ISSUE #4659: Add main admin access to forum without adding them as moderator
## Motivation
Main admin should have access to full site, but forum editing is limited by forum moderate class.
For example - class 248 Forum moderators
You can set this class to main admin too, but then is the main admin listed as a forum moderator, what is not correct. 

Access to the forum (edit, moderate) should be for moderators, forum admin, and main admin.  

For main admin just this is needed:

`define('MODERATOR', (USER && in_array(USERID, $moderatorUserIds) || getperms('0') ));`   
in forum_viewforum.php
in forum_viewtopic.php
in forum_post.php
 
`if(!in_array(USERID, $moderatorUserIds) && !getperms('0'))`
in forum_class.php

I don't know how to add check if user is forum plugin admin P?) so... at least this should work for main admin. 

Thanks

Jimmi08:
- without displaying in the moderator list (if they are not forum moderators directly)

CLOSES #4659

### DIFF
--- a/e107_plugins/forum/forum_class.php
+++ b/e107_plugins/forum/forum_class.php
@@ -535,7 +535,7 @@ class e107forum
 		}
 
 		// Check if user has moderator permissions for this thread
-		if(!in_array(USERID, $moderatorUserIds))
+		if(!in_array(USERID, $moderatorUserIds) && !getperms('0'))
 		{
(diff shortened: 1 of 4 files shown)


CaMer0n:
:+1:

PR #4731, opened 2022-03-31
correct user last visit information #4665

### LINKED ISSUE #4665: user last visit shortcodes (2) look  wrong
## Bug Description
I have imported this field according this
		$target['user_lastvisit']   = $source['user_lastvisit'];
and it is not displayed correctly (but it is correct in database)

I think that this:

```
	function sc_user_lastvisit($parm='')
	{
		return $this->var['user_currentvisit'] ? e107::getDate()->convert_date($this->var['user_currentvisit'], "long") : "<i>".LAN_USER_33."</i>";
	}
	
	
	function sc_user_lastvisit_lapse($parm='')
	{	
		return $this->var['user_currentvisit'] ? e107::getDate()->computeLapse($this->var['user_currentvisit']) : '';
	}
```

Should be at least:
```
	function sc_user_lastvisit($parm='')
	{
		return $this->var['user_currentvisit'] ? e107::getDate()->convert_date($this->var['user_currentvisit'], "long") : e107::getDate()->convert_date($this->var['user_lastvisit'], "long");
	}
	
	
	function sc_user_lastvisit_lapse($parm='')
	{	
		return $this->var['user_currentvisit'] ? e107::getDate()->computeLapse($this->var['user_currentvisit']) : e107::getDate()->computeLapse($this->var['user_lastvisit']);
	}
```

Not sure when user_currentvisit is filled. I suppose it is mean that the user is online? if yes, then a message "online" would be better.  

The problem with the solution above is that if that field is not filled it shows the time 1.1.1970 (in according format)

I would suppose that convert_date() or computeLapse() returns nothing if there is no date sent.

value of $this->var['user_lastvisit']  is `string(1) "0" `

Thanks

Jimmi08:
CLOSES #4665

### DIFF
--- a/e107_handlers/news_class.php
+++ b/e107_handlers/news_class.php
@@ -671,7 +671,7 @@ class e_news_item extends e_front_model
 		$id = intval($id);
 		$nobody_regexp = "'(^|,)(".str_replace(",", "|", e_UC_NOBODY).")(,|$)'";
 
-	  	$query = "SELECT n.*, u.user_id, u.user_name, u.user_customtitle, u.user_image, nc.category_name, nc.category_sef, nc.category_icon FROM #news AS n
+	  	$query = "SELECT n.*, u.user_id, u.user_name, u.user_customtitle, u.user_image, nc.category_id, nc.category_name, nc.category_sef, nc.category_icon FROM #news AS n
 		LEFT JOIN #user AS u ON n.news_author = u.user_id
 		LEFT JOIN #news_category AS nc ON n.news_category = nc.category_id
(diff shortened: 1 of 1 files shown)


e107_handlers/date_handler.php
```diff
@@ -937,8 +938,7 @@ public function dateFormats($datestamp = null)
 						
 						"%Y-%m-%d",
 						"%d-%m-%Y",
-						"%m/%d/%Y",
-						"%d.%m.%Y"
+						"%m/%d/%Y"
```
CaMer0n, line 941: @Jimmi08 Please explain the removal. Thanks.

Jimmi08: It wasn't intention...  I didn't notice this. Please, either change it or close and I do it again. Thanks for noticing.

PR #4736, opened 2022-03-31
#4245 stay on correct page after editing paginated topic

### LINKED ISSUE #4245: Forum: Editing message bring back to first message of the thread
## Bug Description
Recently I made a lot of test in the forum to validate the wysiwyg editor TinyMCE
I have a dedicated thread in which I post tests messages since a long time (the first message of the thread is from 2007 !!) , so it have a lot of pages.
I was testing the capability to edit old messages, let say from page 4. When I was done and click on "Update reply" , the forum bring me back on the first page to the first message which is quite annoying !

## How to Reproduce
Take a long thread in the forum, several pages.
Take a message at the end of the thread and select "Edit" 
Do what you want
Click "Update reply" 
You are back to the first page

## Expected Behavior
I was expecting to see the modified message

Jimmi08:
CLOSES #4245

**Description:**
When you edit a post on the next page (with the paginated topic), after saving you are redirected to the first page and you don't see that edited message then. 

**Changes**
- added p parameter to edit URL in post options and edit image shortcodes
- used this parameter when redirecting after  reply is updated
- it is not needed when updating the thread because you are always on the first page
- added check if it is the first page, then the parameter is not used (to have the same behavior as with thread). 

**Tested**
- PHP 7.4. + PHP 8
- core bootstrap5 theme and custom theme with custom forum templates

**Important Note:**:
Updating/Fixing postoptions are/will be separated issues (more than one). It needs some changes but they are not related to this issue. So now just changed links, nothing more.

**Enhancement Note**:  $_GET['p'] could be replaced by saving the actual page in post data (like thread works - that information should be available) but it is too complicated now for me and there is a bigger risk that something will not work. Free do it this way if you have time.

### DIFF
--- a/e107_handlers/news_class.php
+++ b/e107_handlers/news_class.php
@@ -671,7 +671,7 @@ class e_news_item extends e_front_model
 		$id = intval($id);
 		$nobody_regexp = "'(^|,)(".str_replace(",", "|", e_UC_NOBODY).")(,|$)'";
 
-	  	$query = "SELECT n.*, u.user_id, u.user_name, u.user_customtitle, u.user_image, nc.category_name, nc.category_sef, nc.category_icon FROM #news AS n
+	  	$query = "SELECT n.*, u.user_id, u.user_name, u.user_customtitle, u.user_image, nc.category_id, nc.category_name, nc.category_sef, nc.category_icon FROM #news AS n
 		LEFT JOIN #user AS u ON n.news_author = u.user_id
 		LEFT JOIN #news_category AS nc ON n.news_category = nc.category_id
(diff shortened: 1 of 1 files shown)


CaMer0n:
@Jimmi08 Could you please test this with PHP 8.1 and confirm it works without errors? Also, please pay attention to code formatting and indents.  Thank you.

Jimmi08:
> @Jimmi08 Could you please test this with PHP 8.1 and confirm it works without errors? Also, please pay attention to code formatting and indents. Thank you.

I can't do it, my latest version is 8.0.  This PR is more about what was needed to change to get it works. 
It needs to wait then. I am not able to do it better way.   Sorry.

Issue #4705, opened 2022-03-16
{EMAIL_ITEM=} fails in bootstrap5


Jimmi08:
## Bug Description

file:  shortcodes/email_item.sc

shortcode works but nothing is displayed because this test

```
if(deftrue('BOOTSTRAP'))
{
	$img = "<span class='glyphicon glyphicon-envelope' aria-hidden='true'></span>";
}

```

Output: 
`<a href="***" title="Email"><span class="glyphicon glyphicon-envelope" aria-hidden="true"></span></a>`

It should be replaced by glyph or something like that.

Thanks

Jimmi08:
Or this is another solution:


```
	function sc_emailitem()
	{
        
        if(is_null($this->postInfo['thread_start'])) {
            return "<a class='e-tip btn btn-default btn-info' href='" . e_HTTP . "email.php?plugin:forum." . $this->postInfo['thread_id'] . "'>".e107::getParser()->toGlyph('fas-envelope')."</a>";
        }
        
        if(!empty($this->postInfo['thread_start']))
		{
            return "<a class='e-tip btn btn-default btn-info' href='" . e_HTTP . "email.php?plugin:forum." . $this->postInfo['thread_id'] . "'>".e107::getParser()->toGlyph('fas-envelope')."</a>";
            //return e107::getParser()->parseTemplate("{EMAIL_ITEM=" . LAN_FORUM_2044 . "^plugin:forum.{$this->postInfo['post_thread']}}");
		}
	}
```
that first test is there because this should be displayed outside the thread itself f.e.
`{EMAILITEM} {TRACK} {BUTTONSX} `

![image](https://user-images.githubusercontent.com/5429548/158563749-1e5957c5-57bd-41d8-a378-bd3270ee7664.png)

PR #4738, opened 2022-03-31
#4692 email.php on forum

### LINKED ISSUE #4692: email.php on forum is not working
## Bug Description
 
check e107.nl
image

When you click on the option Email, nothing happens.

I noticed this thanks to Google search console with this error:
```
Redirect error
for URL 
```
https://www.e107sk.com/email.php?plugin:forum.38

Because I wasn't sure if this forum exists (you don't see IDs anywhere) - I checked e107.nl 
 
## Expected Behavior
There is an issue 
`https://github.com/e107inc/e107/issues/4453`
so there could be some changes...   

It is not the same issue, this one is about 404 results for google searching. 

I don't have any email client installed (like Outlook, Thunderbird or something), but I should get at least some info box about not access or not working or something... 

and email.php should be by default added into robots.txt..

Thanks

Jimmi08:
CLOSES #4692 

**Description**
if the user doesn't have the right to send an email (only members have this by default), the email option was still displayed for them. But it did nothing (looked like not working) because redirection to the homepage in email.php itself. 

**Changes**:
- fixed email.php redirection 
- added checks for email rights - to post options and email icon too

**Notes** 
- there is another issue related directly to email icon functionality, so for now, only a check was added.  A better solution is to display a disabled button like the post reply option does. But this is another issue. 
 
- there is another issue related to post options (or will be). so for now, just check was added. 

There should be rules that `{POSTOPTIONS}` is used for posts but {`EMAILITEM}` and `{PRINTITEM}` should be used primarily for thread (or works on both places}

### DIFF
--- a/e107_admin/image.php
+++ b/e107_admin/image.php
@@ -130,7 +130,7 @@ class media_admin extends e_admin_dispatcher
 
 
 	protected $adminMenu = array(
-		'main/list'			=> array('caption'=> LAN_IMA_M_01, 'perm' => 'A'),
+		'main/grid'			=> array('caption'=> LAN_IMA_M_01, 'perm' => 'A'),
 	//	'main/create' 		=> array('caption'=> "Add New Media", 'perm' => 'A'), // Should be handled in Media-Import.
 		'main/import' 		=> array('caption'=> LAN_IMA_M_02, 'perm' => 'A|A1'),
 			'divider/01'        => array('divider'=>true),
@@ -160,8 +160,8 @@ class media_admin extends e_admin_dispatcher
 
 
 	protected $adminMenuAliases = array(
-		'main/edit'	=> 'main/list',
-		'main/grid' => 'main/list'
+		'main/edit'	=> 'main/grid',
+		'main/list' => 'main/grid'
 	);
 
 	protected $menuTitle = LAN_MEDIAMANAGER;
@@ -206,7 +206,7 @@ class media_cat_ui extends e_admin_ui
 
 	function init()
 	{
-	
+
 		$this->restricted = array(
 			'_common'  => '_common',
 			'_icon'    => '_icon',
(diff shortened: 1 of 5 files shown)


CaMer0n:
@Jimmi08 Please check your code formatting. Which editor are you using? Does it work with **.editorconfig** ?

Jimmi08:
> @Jimmi08 Please check your code formatting. Which editor are you using? Does it work with **.editorconfig** ?

PSPAD but I can use Visual Studio Code  (you can't past the code directly to e107 tinymce editor, it adds spans to the code itself, so I always forgot to check it) 

But I just tried 2 formatter there, no change with the code above. 
I need to know what formatter to use (PHP CS FIXER or PHP FORMATTER or other) and their settings (f.e. .php_cs)  Somebody else has to use it too, it is free and good. 

By the way, I couldn't find a link to e107 rules for formatting. I know it exists. 

I have other issues fixed already but they are more complicated and this will be always the problem. Maybe you just could use the idea and write it yourself in needed way. There is no hurry. I am sure it would be save your time. 
Sorry this mess again.

CaMer0n:
@Jimmi08 Please go here, it will tell you which editors support **.editorconfig** already, or have a plugin for download. 
https://editorconfig.org/#pre-installed
After that is set up, it typically should load https://raw.githubusercontent.com/e107inc/e107/master/.editorconfig automatically.

Jimmi08:
@CaMer0n thanks, I don't see any difference but I did it manually. 

Maybe is better to show you the full code (issue #4705}

image


{EMAILITEM} is probably legacy shortcode, the same functionality is in postoptions. 
What is needed is to get it outside posts, the button for the thread.  Solution above works for both scenarios.

You don't have access:  (disabled buttons, red cursor like with reply button, no URL for bots  )
image

You have access (normal button)
image

Issue #4872, opened 2022-09-11
Add wrapper support for custom menus 


Jimmi08:
## Motivation
Minimalize number of needed template keys for custom menus

## Proposed Solution

Adding support for wrappers in renderMenu()  method: 
`$page_shortcodes->setVars($page)->wrapper('menu/'.$page['menu_template'] );`
to menu handler.

Thanks

Jimmi08:
This is example of use:
Without wrapper I would need two templates with button and without button. 

![image](https://user-images.githubusercontent.com/5429548/189536606-02f47cb8-58ff-4530-9c99-f21d37d7f9bd.png)

PR #4873, opened 2022-09-11
Add support for wrapper in custom menus

Jimmi08:
Solves #4872

Minimalize number of needed template keys for custom menus

I am not able to test if it is not breaking something in global, but it works in theme I am working on.


CaMer0n:
It "should" be okay. Thanks :+1:

PR #4882, opened 2022-10-03
Fix #4847 - mistypo in route

### LINKED ISSUE #4847: Missing News breadcrumb for route: news/list/items
## Bug Description
not sure what is wrong

Breadcrumbs are used in layout
```
			<div class="col-lg-12 col-md-12 col-12 mt-4">
				<div class="breadcrumbs-content left">
					{---BREADCRUMB---}
				</div>
			</div>
```

If you set in prefs "List" display, everything is ok
image

If you change it to "default"
image

Tried with both - empty and filled caption for default template
`$NEWS_TEMPLATE['default']['caption'] = null; // add a value to user tablerender()`
or
`$NEWS_TEMPLATE['default']['caption'] = '{NEWS_CATEGORY}'; `

 Is that route correct for default listing?  
Shouldn't be it news/all or something like that?

image

Jimmi08:


CaMer0n:
Thanks :+1:

PR #4896, opened 2022-10-27
Fix for #4895 - wrong message chatbox plugin

### LINKED ISSUE #4895: chatbox plugin - wrong message with float protection enabled
## Bug Description
Message is always 
You may only post once every 1 second
but float time is not 1 s.

## How to Reproduce
 The reason - wrong constant

`$emessage = $tp->lanVars(CHATBOX_L19, FLOODPROTECT ?: 'n/a');`

it should be

`$emessage = $tp->lanVars(CHATBOX_L19, FLOODTIMEOUT ?: 'n/a');`

Jimmi08:
CLOSES #4895


CaMer0n:
Thanks @Jimmi08

Issue #4925, opened 2022-12-18
not able to generate sef url from title with national characters


Jimmi08:
## Bug Description
This is string:  "Ľudia okolo vás - na každom z nás záleží"

In version 2,3,2 result is on live version:
`udia_okolo_vas_na_kazdom_z_nas_zalezi`
(first letter is missing)
on localhost version
`?_udia_okolo_vas_na_kazdom_z_nas_zalezi`
No error in console.  
No idea why there is difference.

in version 2.3.3 (latest git)  nothing is generated and console says:
![image](https://user-images.githubusercontent.com/5429548/208315221-64a3782e-1372-49f2-b25f-4855b86475d6.png)
I don't have clean installation (without plugins) so this could be caused with something else. 
but without national characters, sef generation is working

Thanks

CaMer0n:
@Jimmi08 Can this be closed or are there other characters that cause issues?

Jimmi08:
Closing, I will reopen it or create new issue, not able to test it now.

PR #4926, opened 2022-12-18
missing national character from toAscii()

Jimmi08:
part fix for national https://github.com/e107inc/e107/issues/4925


CaMer0n:
Thanks @Jimmi08 !

Issue #4998, opened 2023-05-10
PM plugin fatal error in PHP 8.2 


Jimmi08:
## Bug Description
 When you send PM without attachment

`Fatal error: Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in ***plugins\pm\pm_class.php:283 Stack trace: #0 ***plugins\pm\pm.php(672): private_message->add(Array) #1 ***plugins\pm\pm.php(869): pm_extended->post_pm() #2 {main} thrown in***plugins\pm\pm_class.php on line 283`

This line:
`$this->pm_send_notify($vars['to_info']['user_id'], $vars, $pmid, count($a_list));`
  
`$vars['uploaded'] is not set`

Vodhin:
It seems that the foreach($array) function in versions of PHP after 7.5 fail if the array is empty, missing, or null. Try setting your PHP version to 7.5 (I have not had any luck running anything e107 on PHP versions newer than 7.5 at all)

Jimmi08:
@Vodhin Thanks for advice but it is not point of creating issues :)

Deltik:
Fixed by https://github.com/e107inc/e107/pull/5006

PR #5006, opened 2023-06-07
PM class - fix PHP 8 #4998

Jimmi08:
fix for #4998 Fatal error: Uncaught TypeError: count()


Deltik:
The test failures are due to an unrelated Internet change: Debian 9 was discontinued and the package repositories have moved to an archive.

I'll have to fix them later.

PR #5027, opened 2023-07-01
Fix #5025: Type error when sending a PM without the attachment field

### LINKED ISSUE #5025: PM plugin - fatal error if user is not allowed to add attachment
## Bug Description
`$this->pmPrefs['attach_class']` has value 254

A normal user doesn't see this field thanks this check:
`if (isset($this->pmPrefs['attach_class']) && check_class($this->pmPrefs['attach_class']))`

BUT in PHP 8.2. it fails on 
`foreach (array_keys($_FILES['file_userfile']['size']) as $fid) `
because value is NULL and not array

`Fatal error: Uncaught TypeError: array_keys(): Argument #1 ($array) must be of type array, null given in ***plugins\pm\pm.php on line 618`

Thanks

Deltik:
### Motivation and Context
There is a bug where the `pm` plugin would return fatally error in PHP 8.2 if the user doesn't have permission to add an attachment when sending a private message. This error is due to the fact that the code expects `$_FILES['file_userfile']` to be an array, but this array will not be present if the user doesn't have permission to add an attachment.

Fixes #5025

### Description
This change adds a check before trying to iterate through `$_FILES['file_userfile']`, ensuring that it is an array before processing the attachment. If this is not an array, then no actions will be taken on `$_FILES['file_userfile']`, which avoids the `TypeError`.

### How Has This Been Tested?
The fix was manually tested in a local development environment with PHP 8.2 by sending PMs with and without the attachment permission.

### Types of Changes

### Checklist

### DIFF
--- a/e107_plugins/pm/pm.php
+++ b/e107_plugins/pm/pm.php
@@ -615,14 +615,18 @@
 
 				$maxsize = intval($this->pmPrefs['attach_size']) * 1024;
 
-				foreach(array_keys($_FILES['file_userfile']['size']) as $fid)
+				if(is_array($_FILES['file_userfile']))
 				{
-					if($maxsize > 0 && $_FILES['file_userfile']['size'][$fid] > $maxsize)
+					$file_userfile = $_FILES['file_userfile'];
+					foreach(array_keys($file_userfile['size']) as $fid)
 					{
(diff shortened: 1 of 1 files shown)


CaMer0n:
Thank you @Deltik !

PR #5423, opened 2025-01-31
Forum postoptions bug

rica-carv:
### Motivation and Context
Prevent duplicated delete buttons when in moderator mode

### Description
Added moderator check

### How Has This Been Tested?
E107 latest github

### Types of Changes

### Checklist
- [ ] I have added [tests](https://github.com/e107inc/e107/tree/master/e107_tests) to cover my changes.
- [ ] All new and existing tests pass.


Jimmi08:
@rica-carv  
Where is related issue to this PR?  I mean - there is template for any bug.  Any PR should have issue or discussion at first.  PR without explanation shoudln't be accepted.  

Yes, you are fixing problem for you, but others can have other opinion.  PR should be created only if there is approved approach to solve issue like this. 

Your PR is related to this: 

image


If I got it right, you are hidding delete button if user is moderator.  This is not correct for me.

User should always had option to delete post if:
- it is its own post, 
- if it is the last in the thread 
- if post is not the initial post of the topic)

Moderator should see delete button, if:
- it is not their own post (there is user button already)  - 
- rules: only show delete button when post is not the initial post of the topic and not user  

Do you see what I mean?  It is opposite approach then yours, this should be discussed and only after final solution PR should be created.  

To complicate things:
-  what if rules for deleting are different for user and moderators (or will be in future)?  And moderator wants to delete post as user (with user rules only)?  Then both buttons are needed. 

So, yes, this is something what should be solved, but not by direct PR, by discussing and making rules.  For example, I don't know why only last post can be deleted by moderator.  

This is example of next approach (changed label "mod" - to let know you are deleting post as moderator):
image

Just my 2 cents.

rica-carv:
> Moderator should see delete button, if:
> 
>     * it is not their own post (there is user button already)  -

Moderator already can see the user buttons, so why had duplicated buttons that do the same thing? To add confusion?

I know that the forum plugin needs a major rework, but it's scheduled to do, but....

Jimmi08:
> > Moderator should see delete button, if:
> > ```
> > * it is not their own post (there is user button already)  -
> > ```
> 
> Moderator already can see the user buttons, so why had duplicated buttons that do the same thing? To add confusion?
> 
> I know that the forum plugin needs a major rework, but it's scheduled to do, but....

You prefer delete button in moderator's section, I prefer delete button in user section and hide it for moderator .  The same thing but different approach.

Deltik (approved):

PR #5445, opened 2025-03-17
Proper generation of Gallery breadcrumb

Serios:
Fixing a mistyped condition, causing improper gallery plugin breadcrumb generation


CaMer0n:
@Serios Good catch. Thanks.

Issue #444, opened 2013-09-20
menu manager will not load


billi2a:
menu manager only shows menu list not locations. the loader icon keeps spinning.

PR #5595, opened 2026-04-20
Admin: restore sidebar menu (icon + title) on legacy pages and fix Docs transitions

Kanonimpresor:
## Problem

Several core admin pages render only via `tablerender()` and never populate
the left admin sidebar (`{ADMIN_MENU}` shortcode). When the user collapses
the sidebar, those pages lose any visual anchor — the panel becomes empty
and unidentifiable, and there is no icon to click to re-expand it.

The Docs page (`docs.php`) additionally lost its v2.3.3 hash-based
content transition because the global `admin.jquery.js` handler is
scoped to `#admin-prefs .plugin-navigation a`, which `docs.php` does not
expose.

`phpinfo.php` overflowed the admin content area because `phpinfo()` ships
hardcoded `width="600"` / `cellpadding` / `cellspacing` / `align`
attributes on its tables, and the dark Modern Light theme rendered the
table contents in low-contrast dark text on dark background.

### Sidebar icons (legacy admin pages)
Add an auto-discovered `<basename>_adminmenu()` function (same pattern as
the existing `frontpage_adminmenu()` in `frontpage.php`) so the
`{ADMIN_MENU}` shortcode populates the left sidebar with a recognizable
icon and title in both expanded and collapsed states.

| File | Function | Icon |
|------|----------|------|
| `e107_admin/ugflag.php`        | `ugflag_adminmenu()`        | `e-maintain-24` |
| `e107_admin/updateadmin.php`   | `updateadmin_adminmenu()`   | `e-adminpass-24` |
| `e107_admin/administrator.php` | `administrator_adminmenu()` | `e-admins-24` |
| `e107_admin/cache.php`         | `cache_adminmenu()`         | `e-cache-24` |
| `e107_admin/emoticon.php`      | `emoticon_adminmenu()`      | `e-emoticons-24` |

Auto-discovery happens here:
`e107_core/shortcodes/batch/admin_shortcodes.php#L786` (`$adminmenu_func = $tmp[0].'_adminmenu';`).

### Docs page (`e107_admin/docs.php`)
- Add `'icon' => 'fa-question-circle'` to each `$adminMenu` entry so the
  collapsed sidebar shows a recognizable icon for every doc.
- Add `protected $adminMenuIcon = 'e-docs-24'` so the panel-heading icon
  stays visible (and clickable) when the sidebar is collapsed.
- Add a small page-scoped `e107::js('footer-inline', ...)` hook that
  listens to clicks on `#admin-ui-nav-menu a[href^="#doc-"]` and animates
  show/hide of the matching `.docs-item`, replicating the v2.3.3
  transition without changing the global `admin.jquery.js` handler.

### `phpinfo.php` + Modern Light theme
- Wrap `phpinfo()` output in `.phpinfo-wrapper`, strip the hardcoded
  `width` / `cellpadding` / `cellspacing` / `align` attributes from its
  tables, and embed them in `.table-responsive` so the page no longer
  overflows the admin content area.
- Add `.phpinfo-wrapper` overrides in `modern-light.css` (white
  background, dark text on `.forumheader2/3` and `.fcaption`) so the
  table contents are readable under the Modern Light admin theme.

## Testing

Manually verified on a local WAMP install (PHP 8.x, MySQL/MariaDB) using
the bootstrap3 admin theme with both Modern Light and Modern Dark CSS:
- All 6 admin pages now show the sidebar with icon + title and remain
  identifiable when the sidebar is collapsed.
- Clicking each doc entry in the sidebar animates the corresponding
  `.docs-item` (matches v2.3.3 behaviour).
- `phpinfo.php` no longer overflows; tables wrap and text is readable.
- No global JS / template changes — every fix is scoped to its own file.


e107_admin/updateadmin.php
```diff
@@ -124,3 +124,21 @@
...
+
+/**
+ * Build the admin sidebar menu (auto-invoked by the {ADMIN_MENU} shortcode).
+ * Provides the icon + caption shown in the left admin panel, consistent with
+ * frontpage.php / docs.php and visible also when the sidebar is collapsed.
+ */
+function updateadmin_adminmenu()
+{
+	$var = array();
+	$var['main']['text'] = UDALAN_8;
```
Deltik: [`UDALAN_8`](https://github.com/e107inc/e107/blob/4f0f32ffe12728389a61c71ecd8b6774d967bfe8/e107_languages/English/admin/lan_updateadmin.php#L18) is "Password updated for", which doesn't make sense in context:

Image

Please choose a more fitting localization string for use in this sidebar.

Deltik (changes requested):

e107_admin/updateadmin.php
Kanonimpresor: Thanks for the review @Deltik – switched to UDALAN_7 ("Change password") which matches the page's own submit button label and avoids reusing the success-message string. Pushed as d33b3f8.

e107_admin/phpinfo.php
```diff
@@ -26,16 +26,37 @@
...
+// Remove hard-coded width/cellpadding/cellspacing attributes that cause horizontal overflow.
+$phpinfo = preg_replace('/\s(width|cellpadding|cellspacing|align)="[^"]*"/i', '', $phpinfo);
 //$phpinfo = str_replace('<table ','<table class="table table-striped adminlist" ',$phpinfo);
 $phpinfo = str_replace('name=','id=',$phpinfo);
 $phpinfo = str_replace('class="e"','class="forumheader2 text-left"',$phpinfo);
 $phpinfo = str_replace('class="v"','class="forumheader3 text-left"',$phpinfo);
 $phpinfo = str_replace('class="v"','class="forumheader3 text-left"',$phpinfo);
 $phpinfo = str_replace('class="h"','class="fcaption"',$phpinfo);
-$phpinfo = preg_replace('/<table[^>]*>/i', '<table class="table table-striped adminlist"><colgroup><col style="width:30%" /><col style="width:auto" /></colgroup>', $phpinfo);
+$phpinfo = preg_replace('/<table[^>]*>/i', '<table class="table table-striped table-bordered adminlist phpinfo-table"><colgroup><col style="width:30%" /><col style="width:auto" /></colgroup>', $phpinfo);
```
Deltik: This forced rewrite creates some visual awkwardness where tables are 30% wide:

Image

Compare to before, which required scrolling:

Image

The root cause is line 42 injecting a 2-column `<colgroup>` into *every* phpinfo sub-table:

```php
$phpinfo = preg_replace('/<table[^>]*>/i', '<table class="table table-striped table-bordered adminlist phpinfo-table"><colgroup><col style="width:30%" /><col style="width:auto" /></colgroup>', $phpinfo);
```

… but some of phpinfo's sub-tables are inherently 1-column (the "PHP Quality Assurance Team", "PHP Documentation", "PHP License" blocks all look like this):

```html
<table>
<tr class="h"><th>PHP Quality Assurance Team</th></tr>
<tr><td class="e">Ilia Alshanetsky, Joerg Behrens, …</td></tr>
</table>
```

Forcing those rows into a 30%/auto column grid pins the single `<td>` to column 1 and leaves the rest empty, hence the 30%-wide boxes.

While we're here, I'd actually like to propose going further and dropping the attribute-stripping and class-renaming pass entirely.  HTML presentational attributes (`width`, `cellpadding`, `cellspacing`, `align`, `bgcolor`, `valign`, etc.) map to the lowest-specificity CSS, so any author rule overrides them, so no regex needed.  Something like:

```php
ob_start();
phpinfo();
$phpinfo = ob_get_clean();

// Keep only <body>…</body>, drop phpinfo's own <style> so it doesn't bleed into the admin shell.
if (preg_match('#<body[^>]*>(.*?)</body>#is', $phpinfo, $m))
{
    $phpinfo = $m[1];
}
$phpinfo = preg_replace('#<style\b[^>]*>.*?</style>#is', '', $phpinfo);

$phpinfo = '<div class="phpinfo-wrapper"><div class="table-responsive">' . $phpinfo . '</div></div>';
```

… paired with CSS that styles phpinfo's native `td.e` / `td.v` / `tr.h` classes directly (scoped under `.phpinfo-wrapper`), without `table-layout: fixed` and without a forced `<colgroup>`:

```css
.phpinfo-wrapper table         { width: 100%; border-collapse: collapse; margin-bottom: 1rem; }
.phpinfo-wrapper td,
.phpinfo-wrapper th            { padding: 6px 10px; border: 1px solid #ddd;
                                 text-align: left; vertical-align: top;
                                 word-break: break-word; overflow-wrap: anywhere; }
.phpinfo-wrapper tr.h th       { background: #eee; text-align: center; font-weight: bold; }
.phpinfo-wrapper td.e          { background: #f5f5f5; font-weight: bold; }
.phpinfo-wrapper td.v          { background: #fff; font-family: monospace; }
```

Because each phpinfo sub-table keeps its natural column count, auto-layout hands the single-cell rows the full width for free.  `width:100%` beats `width="600"`; `border-collapse` + `padding` beat `cellspacing`/`cellpadding`; `text-align:left` beats `align="center"`.  `.table-responsive` still provides horizontal scroll if a value ever pushes wider than the viewport.

One small follow-on: the security-risk highlighter below ([lines 111–120](https://github.com/e107inc/e107/blob/d33b3f80ec1731d8363a291e15519c9fcbf01f81/e107_admin/phpinfo.php#L111-L120)) is currently matching the renamed classes, so it would need to match the originals instead:

```php
$srch = '<tr><td class="e">'.$risk.'</td><td class="v">';
$repl = '<tr><td class="e">'.$risk.'</td><td title="'.e107::getParser()->toAttribute($diz).'" class="v alert alert-danger">';
```

The net effect is fewer moving parts, no brittle attribute-regexes, full-width rows restored, and it should stay correct across future phpinfo HTML changes.

Deltik (changes requested):

e107_admin/docs.php
```diff
@@ -27,6 +27,39 @@
 
 ');
 
+// Sidebar (admin nativo) -> show/hide de los .docs-item con animación, sin tocar el JS global.
+// Los items se generan en init() con uri="#doc-N"; aquí enganchamos los clicks de esos enlaces
```
Deltik: This line and the line above restate what the code is doing.  They don't add value and should be removed, especially if the code changes in the future and the comments are forgotten.

e107_admin/docs.php
```diff
@@ -27,6 +27,39 @@
...
+// dentro del sidebar admin (#admin-ui-nav-menu) y replicamos el efecto que en otras páginas
+// hace el handler "#admin-prefs .plugin-navigation a" de admin.jquery.js.
```
Deltik: This line and the line above do capture why this change is needed, but it's written in a language that not all the maintainers (like myself) know.  Because the commit message would capture the _why_, I suggest removing the comment altogether.

e107_admin/docs.php
```diff
@@ -49,6 +82,11 @@ class docs_admin extends e_admin_dispatcher
 
 	protected $menuTitle = LAN_DOCS;
 
+	// Icono del título del menú lateral (panel-heading). Visible también cuando
+	// el sidebar está colapsado, sirviendo de "ancla" para volver a expandirlo.
+	// Usa el sprite nativo e-docs-24 (mismo patrón que admin_log.php, users.php, etc.)
```
Deltik: As in the other comments about the comments, this what-why-reference pattern is already captured in the commit message, so this comment should be removed as well.

Deltik (changes requested):

e107_admin/phpinfo.php
Kanonimpresor: Thanks @Deltik for the detailed review — pushed d57c3c5f8 which rewrites the phpinfo handling along the lines you proposed.

Changes vs. previous version

Capture phpinfo() with a single ob_get_clean() (the previous code also had an orphaned ob_end_clean() further down which was closing the wrong buffer and dumping the page above the admin shell — removed).
Extract the <body> and strip the inline <style> so PHP's own palette no longer wins over the admin theme.
Drop the attribute-stripping, the class="e" → forumheader2/3 renaming and the forced <colgroup> injection. The output now keeps phpinfo's native markup (tr.h, td.e, td.v).
Wrap the whole thing in a single <div class="phpinfo-wrapper"> and style the native classes scoped under that wrapper. As you noted, HTML presentational attributes have the lowest specificity so plain selectors are enough — no !important needed.
The security highlighter is back to targeting the original td.e / td.v selectors (it was previously rewritten to chase the renamed classes).
Two small deviations from your sketch, with rationale

The CSS is injected inline via e107::css('inline', ...) instead of living in modern-light.css. The admin area can be rendered with several Bootstrap 3 variants (modern-light, modern-dark, corporate, kadmin, bootstrap-dark…) and the previous CSS-in-theme-file approach only styled one of them. Putting it on the page that needs it keeps it portable. The .phpinfo-wrapper block that had been added to modern-light.css is reverted in this commit.
No interior .table-responsive wrapper. Bootstrap 3's .table-responsive is designed to wrap a single <table>, but phpinfo's body is a sequence of <h1>/<h2>/<table> blocks — wrapping all of it broke the BS3 grid (the admin sidebar dropped below the content). Instead the outer .phpinfo-wrapper itself uses overflow-x: auto, which gives the same horizontal-scroll safety net without disturbing the grid.

Also added one extra rule that came up while testing on this WAMP install:


.phpinfo-wrapper tr > td.e:not(:only-child) { white-space: nowrap; }
Without table-layout: fixed (which you explicitly recommended against), a very long value in td.v — long PATH, cookie, opcache directive list, etc. — was squeezing the label column down to one character per line. The :not(:only-child) guard makes sure the 1-column sub-tables (PHP QA Team, License, Documentation) still wrap naturally and stay inside the admin column width.

Happy to iterate further if you'd prefer a different trade-off on either of those two points.

e107_admin/docs.php
Kanonimpresor: Done in 3d95c1a39 — removed all three Spanish comment blocks you flagged (the one above e107::js('footer-inline', …), the one inside the hash-activation block, and the what-why-reference above $adminMenuIcon). Agreed on the policy: the why lives in the commit message, the code speaks for itself.

e107_admin/phpinfo.php
```diff
@@ -23,19 +23,61 @@
...
+
+// Wrap once. .phpinfo-wrapper scopes all CSS to this page and provides a safety overflow-x.
+// The CSS below styles phpinfo's native td.e / td.v / tr.h classes (width:100% + word-break
+// on cells, so wide values wrap instead of forcing the admin column wider).
+$phpinfo = '<div class="phpinfo-wrapper">'.$phpinfo.'</div>';
+
+// Inject the phpinfo CSS inline so it works under any admin theme variant (modern-light,
+// modern-dark, corporate, kadmin…) without coupling phpinfo.php to a specific stylesheet.
+// Styles target phpinfo()'s native td.e / td.v / tr.h classes scoped under .phpinfo-wrapper.
+e107::css('inline', '
```
Deltik, line 48: https://github.com/e107inc/e107/pull/5595/changes/d57c3c5f8c5b8b0251ca74ff0be02e778828d47f does ensure styling works on different skins, but the light-colored table styling is now dominating the dark skins like "Modern Dark":

Image

This `e107::css('inline', …)` is mixing some concerns:

1. Structural things that don't depend on the theme
2. Palette, AKA theme-specific colors

To fix this, I highly recommend pulling the palette into the theme stylesheets that already own the color decisions for the rest of the admin area.

Deltik (changes requested):

e107_admin/phpinfo.php
Kanonimpresor: Hi @Deltik,

Done in a52321df1 — split phpinfo styling into structure vs. palette as you suggested.

phpinfo.php — the inline e107::css('inline', …) block is now strictly structural: overflow-x, box-sizing, table layout / collapse / width, padding, border width/style (no color), word-break, the td.e:not(:only-child) nowrap guard, monospace font on td.v. Every background-color, color and border-color is gone.

e107_themes/bootstrap3/css/*.css — added a .phpinfo-wrapper palette block at the end of each admin stylesheet shipped with the bootstrap3 theme, owned by the same files that already own the rest of the admin palette:

Skin	Palette
modern-light.css, kadmin.css, corporate.css	white tables, #f5f5f5 label cells, #eee header rows, #ddd borders, dark text, #f2dede/#a94442 alert
modern-dark.css, modern-dark-2.css, bootstrap-dark.min.css	#2f2f2f/#373737/#2a2a2a/#1f1f1f cells, #c6c6c6/#e6e6e6 text, #444 borders, #5bc0de links, #5a1a1a/#f2b5b5 alert
The CDN-hosted skins (Flatly, Sandstone, Superhero) can't be patched locally, but being plain Bootstrap 3 they fall back to white tables and stay legible.

Verified locally on Modern Light and Modern Dark.

PR #5676, opened 2026-05-26
lan files to use array style

tgtje:
### Why

last remainging ph files in array style

### What Changed

define to array

### How It Was Tested

fresh git installs with and without locale packs (_full array and define mode_)

### Backwards Compatibility

No BC impact noticed.

### Checklist

- [] One issue per PR: the diff is scoped to this change only

### DIFF
--- a/e107_plugins/comment_menu/languages/English.php
+++ b/e107_plugins/comment_menu/languages/English.php
@@ -9,19 +9,19 @@
  * $URL$
  * $Id$
 */
+return[
 
-define("CM_L1", "No comments yet.");
-define("CM_L2", "");
-define("CM_L3", "Caption");
-define("CM_L4", "Number of comments to display?");
-define("CM_L5", "Number of characters to display?");
-define("CM_L6", "Postfix for too long comments?");
-define("CM_L7", "Show original news title in menu?");
-define("CM_L8", "New Comments Menu Configuration");
-//define("CM_L9", "Update Menu Settings");
-//define("CM_L10", "Comments menu configuration saved");
-define("CM_L11", "on");
-define("CM_L12", "Re:");
-define("CM_L13", "Posted by");
-
-
+'CM_L1'=> "No comments yet.");
+'CM_L2'=> "");
+'CM_L3'=> "Caption");
+'CM_L4'=> "Number of comments to display?");
+'CM_L5'=> "Number of characters to display?");
+'CM_L6'=> "Postfix for too long comments?");
+'CM_L7'=> "Show original news title in menu?");
+'CM_L8'=> "New Comments Menu Configuration");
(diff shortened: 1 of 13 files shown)


tgtje:
closing, added  before error message arrived will create new pr with fixes

tgtje:
Thanks, as i can not get a git client to work, i have to do it live online through github itself.
For the standard > information unknown  (should be public).

rica-carv:
So this means that locale lan files also need to be converted????

kreossino:
I created a small script and added it to the directory. It formatted all the files according to the style specified by e107help. (  _4-space indent on each entry, spaces around =>._  )
If you want to try it, just place it in the language directory and it will scan and modify all subdirectories automatically.
[fix_e107_lan.php](https://github.com/user-attachments/files/28285482/fix_e107_lan.php)

`php fix_e107_lan.php`

Note: this script fixes the formatting for all files in the directory and its subdirectories. However, in e107_languages/admin/lan_admin.php at line 155, it encounters an empty line and incorrectly inserts double quotes "". This specific modification temporarily breaks the rendering of ../e107_admin/admin.php. I have manually corrected this line, so now all files are perfectly aligned according to the guidelines, and everything works correctly.
this the line

`   
 'ADLAN_ERR_3' => "There are one or more files in your public upload directories that are not in your allowed upload filetypes list. These may have been placed here by an attacker and if so should be removed [b]immediately[/b]. You should [b]not[/b] open these files as this may execute any malicious code the file might contain; e.g. do not open them with your browser.
If you recognise these files as being legitimate, it is likely that the filetype is no longer in the allowed filetypes list for some reason, and you will need to re-add it (see admin => uploads). You should not allow the upload of .html, .txt, etc as an attacker may upload a file of this type which includes malicious javascript. You should also, of course, not allow the upload of .php files or any other type of executable script.

Below is the list of files that could potentially be malicious:",

`

tgtje:
@kreossino thank you , script is usefull

rica-carv:
> If you want to try it, just place it in the language directory and it will scan and modify all subdirectories automatically.
> [fix_e107_lan.php](https://github.com/user-attachments/files/28285482/fix_e107_lan.php)

Sorry, @kreossino, but the script doesn't work with standard v2.3 Portuguese lan files....
I put it in the directory and it simply changes a single line inside a couple of files, completelly messing the file....

I ended up writing a new conversion script, if it helps anybody: [conv_e107_lan.php](https://github.com/user-attachments/files/28422373/conv_e107_lan.php)
For me this one works now....

tgtje:
@rica-carv nice, just tested>  a complete conversion  (incl define to array)  where @kreossino 's intended as style fix.
Both are keepers. 
Thanks

Issue #5210, opened 2024-03-12
[Bug]: [Accessibility]: bootstrap 5 theme - the collapseable menus for user information and private messaging system doesn't have any alternative text for screen readers


sanslash332:
### What e107 version are you using?

Latest Github version (just updated)

### Bug description

If your site are using the bootstrap 5 example theme, and your enter to it with a not signed session, you  find two sign-in related links below (or next to, i cant tell you the visual part) the navigation area, one for sign-up and sign-in.
These links are OK, and have their corresponding alternative text for screen readers.

But, after you sign-in, these links are replaced with two expandable menus. One for the private messaging system, and other for control your profile with links to sign-out, view the profile page, etc.
These expandable menus doesn't have any understandable alternative text to understand their function when you use a screen reader. They contains a simple "#" label. You can understand their function just after press it and view the expanded menu.

So is possible add a aria-label with the corresponding text that represent their function?
I've tried to find the template where these menus are built, but on the theme I just found the " {SIGNIN}" shortcode that aren't created on the theme itself, and cant found it on the built-in templates or shortcodes to check how it works and try to modify by myself.

### How to reproduce

1. grab your screen reader of preference. NVDA, JAWS or narrator on windows, voiceover on apple systems.
2. Open any web browser
3. enter to any e107 website using the default bootstrap 5 example theme. For example www.tiflojuegos.com
4. Find the sign-in and sing-up links. check that they are OK
5. enter to the sign-in option, and input your credentials to sign-in.
6. After the page loads, go to the same place where the sign-in and sign-up links were, and note that now, are two links labelled as "#"
7. Press enter on one of it, and surprise, you found that one of them can be a menu of your private messaging system, or your personal menu.

### Expected behavior

That these menus have a small name (not a complex description) with their function.
Like messages, pm, private messages or something similar for the first menu, and profile menu, profile, or simple the name of your signed user for the second.

### What browser(s) are you seeing the problem on?

Firefox, Chrome / Brave, Microsoft Edge

### PHP Version

7.4.33

sanslash332:
O, thanks @Alex-e107nl for the tip.

I don't expected that the signing template was part of a plugin 😢
But well, lets try some changes, and if they fit wells, prepare a small PR to fix this in general. Maybe can be a small change and benefit to all blind or vision impaired people that enter to  any e107 site :-)

Thanks.

Moc:
PR is very much appreciated!

PR #5716, opened 2026-06-05
Refs #5210: add accessible name + ARIA state to PM nav toggle

Jimmi08:
### Why

Refs #5210 (partial). The private-messages toggle in the signin menu
(`sc_pm_nav`) exposes only an icon glyph and a number to assistive tech, so a
screen reader announces nothing meaningful about the control — at best it reads
the unread count "1" with no context. This gives the toggle a proper accessible
name and dropdown state.

### What Changed

- Added `aria-label` to the `.pm-nav` toggle, built from translatable LAN
  strings. When there are unread messages the label includes the count
  (e.g. "Private messages, 3 unread"); otherwise just the base label.
- Added `role="button"` and `aria-expanded="false"` to the toggle (it is an
  `href="#"` link acting as a button; Bootstrap flips `aria-expanded` at
  runtime, the initial state belongs in the markup).
- Removed a duplicate `data-bs-toggle="dropdown"` attribute on the toggle.
- New LAN constants for the a11y text (`LAN_PLUGIN_PM_NAV`,
  `LAN_PLUGIN_PM_UNREAD`), with a `defined()` fallback so existing skins keep
  working before translations are added.
- Label value is escaped with `$tp->toAttribute()` and the count cast via
  `intval()`.

### How It Was Tested

Loaded the signin menu while logged in, both with and without unread messages.
Confirmed the toggle now has an accessible name in the browser accessibility
tree, that the count is included when present, and that `aria-expanded` flips
on open/close. Verified the menu renders identically for sighted users. Not run
against an automated a11y suite.

### Backwards Compatibility

No visual or behavioral BC impact. Changes are additive ARIA attributes plus a
duplicate-attribute removal; the `.pm-nav` class, dropdown structure, URLs and
existing CSS/JS hooks are untouched. The rendered HTML gains attributes but
nothing existing skins or scripts rely on is removed or renamed.

Not addressed here (follow-up): the three `dropdown-item` links are still
wrapped in a single `<li>`, so the list is announced as one item instead of
three and keyboard item-navigation is affected. Fixing that is a structural
markup change (splitting into one `<li>` per item) and is better scoped to its
own PR with separate keyboard-navigation testing.

### AI Model (Optional)

Claude Opus 4.8

### Checklist

- [ ] New or changed behavior has test coverage (or explain why not) — N/A, ARIA-only markup output with no template-output test harness

### DIFF
--- a/e107_plugins/pm/e_shortcode.php
+++ b/e107_plugins/pm/e_shortcode.php
@@ -37,31 +37,31 @@ class pm_shortcodes extends e_shortcode
 	function sc_pm_nav($parm='')
 	{
 		$tp = e107::getParser();
-
 		if(!isset($this->prefs['pm_class']) || !check_class($this->prefs['pm_class']))
 		{
 			return null;
 		}
-
 		$mbox = $this->pm->pm_getInfo('inbox');
 
 		if(!empty($mbox['inbox']['new']))
 		{
 			$count = "<span class='label label-warning'>".$mbox['inbox']['new']."</span>";
-			$icon = $tp->toGlyph('fa-envelope');
+			$icon  = $tp->toGlyph('fa-envelope');
+			$aria  = LAN_PLUGIN_PM_NAV.', '.intval($mbox['inbox']['new']).' ' . LAN_PLUGIN_PM_UNREAD;
 		}
 		else
 		{
-			$icon = $tp->toGlyph('fa-envelope-o');
+			$icon  = $tp->toGlyph('fa-envelope-o');
 			$count = '';
+			$aria  = LAN_PLUGIN_PM_NAV;
 		}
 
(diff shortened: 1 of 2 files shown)


Jimmi08:
@e107help It is some mistake on my side, I will check it. That code shouldn't be there.

e107_plugins/pm/e_shortcode.php
```diff
@@ -34,42 +34,41 @@ function __construct()
 
 
 
-	function sc_pm_nav($parm='')
+function sc_pm_nav($parm='')
```
Deltik: This method lost its indentation, which should be restored.

Deltik (changes requested):

Deltik (approved):

PR #5739, opened 2026-06-08
Restore auto date generation for Welcome Message

Serios:
In some code rewrite, date generation when creating Welcome message was omitted, this is restoring it. I think is not a bad idea for date of creation to exist and if user decide to show it or not


Deltik (approved):
@e107help, could you write a history report about the motivation and context for this fix so that future readers of this pull request can follow along?  Also, please create a backport to branch `release/v2.3.x`.

PR #5776, opened 2026-06-15
fix for conversion error in core + 1 LAN change

tgtje:
### Why

Fix conversion error whre commented LAN's are activated

### What Changed

restyled and repaired ( 1 lan in poll class changed into word) as no longer exists in that matter (forum old) the (LAN_4) itself is reused for other (eg in lan top.php)

### How It Was Tested

on local machine heavy testing

### Backwards Compatibility
No BC impact. (only late  v 0.6 versions might be affected)

### Checklist

- [ ] One issue per PR: the diff is scoped to this change only
- [ ] No unrelated reformatting, renames, or import reordering


tgtje:
For your convenience (untouched)  : the MAIN English language file >English.php is not equal in styling (uses "..." ).
Do not know if intended or typo.

Alex-e107nl:
Hi @tgtje do you mean this: 

```
// const CORE_LC = "en";
// const CORE_LC2 = "gb";

return [
	"CORE_LC" => "en",
	"CORE_LC2" => "gb",
```
```
const CORE_LC = "en";
const CORE_LC2 = "gb";	
```
Gives a php error double declaration...

tgtje:
No i think, (guessing what you mean) @Alex-e107nl  _styling_ as explained for lan files should have
 **'LAN_TEST'** =  >   "descript",  << single ' front '
Now for main English php they are 
**"LAN_TEST"** =  >   "descript",   < double " front "

e107_plugins/poll/poll_class.php
```diff
@@ -580,7 +580,7 @@ function render_poll($pollArray = "", $type = "menu", $POLLMODE = "", $returnMet
 		}
 		elseif ($type == 'forum')
 		{
-			$caption = LAN_4;
+			$caption = 'Post Poll'; //LAN_4 former forum
```
Deltik, line 583: This should be a new LAN string, shouldn't it?

Deltik (changes requested):

e107_plugins/poll/poll_class.php
tgtje: @Deltik LAN_4 is in lan_top php with description Replies.

The descrition Post Poll is in forum English front commented out used in v 06.16.
Since regarded as a possible 'thing' for BC , one can assume there are no longer v 06 versions really active.
So just in case for BC i deleted/altered to plain text description. ( can not see it appear anywhere in use, not in forum nor in poll itself, not sure iff leftover ).

origin trigger was :  Use of undefined constant LAN_4 - assumed 'LAN_4' (this will throw an Error in a future version of PHP), Line 583 of xxxxxx\107_plugins\poll\poll_class.php

Having two same LAN's with descrition differences might confuse anyway.

Deltik: @tgtje, I meant that shouldn't the "Post Poll" string turn into a brand new constant, like `POLLAN_POST_POLL => 'Post Poll'` in the `e107_plugins/poll/languages/English_admin_poll.php` file so that this text may be translatable?

tgtje: @Deltik the decision is yours, yes a new LAN is possible if its uncomfortable. ( if needed at all = ?)


Problem : trigger reacted as undefined, likely because LAN_4 does not exist in relation to/from poll anymore,  OR can no longer find LAN_4      (present in lan_top php (called 2 times from forum files >> which call the lan top).
 
Tested : using POLLAN_POST_POLL + description OR unchanged eg txt name, the trigger is gone for undefined, but still not able to find the description (**as text**) anywhere (which by code would be somewhere in a tableblock  in forum-header-) , also not in database ( v 2.3.7 used).

Diving deeper for cause : modification/recoding of forum  and LAN_4 got left behind in this case (overseen?) thosedays (2013/14) OR Poll had changes then too.

(and unrelated for now helas.. a few duplicate LAN descriptions (eg LAN's are still present in forum).

Deltik (approved):

PR #6134, opened 2026-09-02
Accept jpeg and gif attachments as images in the SMF importer

darkdi:
`smf_import::processAttachments` decides between an image attachment and a plain file, and its condition tests `jpg` twice, so the third slot never covered anything. An SMF board that has `.gif` or `.jpeg` attachments imports them as generic files rather than images.

I did not guess the missing name. Everywhere else e107 spells the image set out as jpg, jpeg, png and gif, in `media_class.php`, `e_thumbnail_class.php`, `e_parse_class.php` and the form handler regex, so I used that same set here and folded the chain into `in_array` to keep it readable. Newer parts of the tree also carry `webp`, which I left out because an SMF board old enough to import predates it.

### DIFF
--- a/e107_plugins/import/providers/smf_import_class.php
+++ b/e107_plugins/import/providers/smf_import_class.php
@@ -325,7 +325,7 @@ class smf_import extends base_import_class
 		}
 
 
-		if($source['fileext'] == 'png' || $source['fileext'] == 'jpg' || $source['fileext'] == 'jpg')
+		if(in_array($source['fileext'], array('png', 'jpg', 'jpeg', 'gif')))
 		{
 			$type = 'img';
 		}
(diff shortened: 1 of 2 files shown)


Deltik:
@e107help: As this is an uncontroversial fix, I would like you to adopt this pull request (GitHub says, "Maintainers are allowed to edit this pull request."), add WebP support, add the test, and backport to the `release/v2.3.x` branch as a twin pull request.

darkdi:
webp is in. I had left it out because the boards being imported are usually old enough to predate it, but SMF 2.1 takes webp attachments and there is no cost to covering them.

The test and the v2.3.x twin I leave to @e107help as you assigned, and the branch is open for maintainer edits.

Deltik (approved):

PR #6261, opened 2026-09-03
Text to LAN's History

tgtje:
### Why

Hardcoded texts to LAN's for translation

### What Changed

Text to LAN's History

### How It Was Tested
Visual, debugging (local for errors)

### Backwards Compatibility

For MASTER History version only (not present in earlier (2.3.. branch versions)

### Checklist


tgtje:
_Thanks for the detailed answer._ 
It raises some intriging questions for me.. tab and spaces (wonder where it can be seen, using (as those are familiar for me) PSpad, Notepad++ (both Unix utf-8 no bom, maybe a small setting to help me in there, have to look, and find out).
(edit : found it, and a possibility)

Ok i take a new goal, will add the docblock into it.
LAN's _full decription words_ (had thought about it, but they might get lengthy hence... (my idea unsure if allowed).

About line152 till 204 > these line_types are like info/error/success/ warning . ( the top appearing coloured message bars after action)
Yep know them, error and warnings? i was somewhere in time told/requested to NOT touch these.

As for FULL translatable things, there are placeholders screens too with description, but those would be for a later state
(my idea), . 


And yes i was before also asked to try  > want [x] placeholders read back through $tp->lanVars().
That is where the problem lies. What ever i did, more tries than one, i just can't grasp, as it seemed to change on a per position/file ) method?. I gave up. No real fixed solution.
The same with Github and git clients...  Those are the real two things i am lacking in here (like my brain refuses to store it).
I use Github online.

Any way, wil have a new go with a fresh root/fork ( already in a few days 300 commits behind.....and doublecheck preferred).
Always trying to create a new pr at newest point, closing this.

PR #6288, opened 2026-09-07
Text to LAN's History + file creation

tgtje:
### Why

History hardcoded texts to LAN's for translations

### What Changed

History hardcoded texts to LAN's . LAN's in own admin lan_history.php Mian file added lang.path

### How It Was Tested

Visual (local) on 2 languages (EN,NL) diff php versions; 'cosmetics checked', multiple scenarios tested. 
Database> iff actions will take place in table using 2 langs.

### Backwards Compatibility
No: Master only

### Checklist

### DIFF
--- a/e107_admin/history.php
+++ b/e107_admin/history.php
@@ -9,10 +9,9 @@ if (!getperms('7'))
 	e107::redirect('admin');
 	exit;
 }
+
 e107::includeLan(e_LANGUAGEDIR.e_LANGUAGE."/admin/lan_history.php");
-
 e107::css('inline', " td.history-data pre { max-width: 800px; } }");
-
 
 class history_adminArea extends e_admin_dispatcher
 {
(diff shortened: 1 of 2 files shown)


tgtje:
@e107help ea. Ok, could fix the cosmetics, i can commit a new PR , is that what you like to see (likely will add a new one in the 2 present ?) That is not realy clear to me in the answer ( i don't mind)
e107::coreLan('history', true); tried this like commented line, with coreLan etc : it will  (for me at least) not work (tested it) , and hence needed the straight call for file.
As for the new file suggested.. i took over the idea, yes, but had that in mind earlier too (why make lan_admin more lengthy and increase possible errors), now all belonging is in 1 page (basic idea) except the LAN_HISTORY for 'link' in topbar menu

tgtje:
Let's keep in miind that Geltik has to call it, i close this one again, and restart a new one.
I don't mind, sorry for the time you have to spend on it..

PR #6294, opened 2026-09-08
Text to LAN's + admin file

tgtje:
### Why

History hardcoded texts to LAN's for translations

### What Changed

History hardcoded texts to LAN's . LAN's in own admin lan_history.php Mian file added lang.path

### How It Was Tested

Visual (local) on 2 languages (EN,NL) diff php versions; 'cosmetics checked', multiple scenarios tested.
Database> iff actions will take place in table using 2 langs.

### Backwards Compatibility

No: Master only

### Checklist

### DIFF
--- a/e107_admin/history.php
+++ b/e107_admin/history.php
@@ -10,7 +10,7 @@ if (!getperms('7'))
 	exit;
 }
 
-// e107::lan('history',true);
+e107::includeLan(e_LANGUAGEDIR.e_LANGUAGE."/admin/lan_history.php");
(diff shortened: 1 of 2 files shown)


tgtje:
Thanks. Quote: using coreLan and it still did nothing, say so and I will dig again.
Is it necessary > maybe
In previous runs using coreLan it did nothing,  Not tested in between the last 2 tries.
Maybe...just maybe.. is the position of that line in file (on top after require or somewhere deeper....those jump in a lot of files) of influence (not tested). As noticed..sometimes makes one wonder.

Deltik (approved):

PR #6310, opened 2026-09-10
Text to LAN for BANNER plugin

tgtje:
### Why

For translation hardcoded texts to LAN

### What Changed
Hardcoded text to LAN's

### How It Was Tested

Tested visual and by testcases (local) on v 2.3.12 php (8.3 14) and latest master (php 8.5.16)

### Backwards Compatibility

No BC impact

### Checklist


tgtje:
Just a question : as this is BC compatible : the PR is in array mode.
Of course i test in standard latest version (define English) too, but a PR for those can 'not'be made. How would one handle this. Or are changes for those made in backend? ...

Deltik:
You can create backports to older supported e107 release tracks by setting the base to that release branch.  Branch `release/v2.3.x` is where I make e107 v2.3 releases.

Ideally, I'd like to stop supporting e107 v2.3, but e107 v2.4 still needs a lot more work before I can drop e107 v2.3.

I'm having @e107help give this one a look, and it may choose to advise you on backports and the different localization styles.

tgtje:
Thanks for the info, will commit again with changes adjusted andmentioned.Closing this..

